PR #9, opened 2015-09-08
Refactor copycat-jdbc to work with kakfa and copycat data API

Ishiihara:
@ewencp Can you review it when you have time? Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
 
     <properties>
         <confluent.version>2.0-SNAPSHOT</confluent.version>
-        <copycat.version>2.0-SNAPSHOT</copycat.version>
+        <kafka.version>0.9.0.0-SNAPSHOT</kafka.version>
         <junit.version>4.11</junit.version>
         <easymock.version>3.0</easymock.version>
         <powermock.version>1.6.2</powermock.version>
@@ -70,9 +70,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>io.confluent</groupId>
+            <groupId>org.apache.kafka</groupId>
             <artifactId>copycat-api</artifactId>
-            <version>${copycat.version}</version>
+            <version>${kafka.version}</version>
         </dependency>
         <dependency>
             <groupId>io.confluent</groupId>
--- a/src/main/java/io/confluent/copycat/jdbc/BulkTableQuerier.java
+++ b/src/main/java/io/confluent/copycat/jdbc/BulkTableQuerier.java
@@ -1,27 +1,27 @@
 /**
  * Copyright 2015 Confluent Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
  *
  * http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
  **/
 
 package io.confluent.copycat.jdbc;
 
+import org.apache.kafka.copycat.data.Struct;
+import org.apache.kafka.copycat.source.SourceRecord;
+
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-
-import io.confluent.copycat.data.GenericRecord;
-import io.confluent.copycat.source.SourceRecord;
+import java.util.Collections;
+import java.util.Map;
 
 /**
  * BulkTableQuerier always returns the entire table.
@@ -44,9 +44,10 @@ public class BulkTableQuerier extends TableQuerier {
 
   @Override
   public SourceRecord extractRecord() throws SQLException {
-    GenericRecord record = DataConverter.convertRecord(schema, resultSet);
+    Struct record = DataConverter.convertRecord(schema, resultSet);
     // TODO: key from primary key? partition?
-    return new SourceRecord(name, null, name, null, null, record);
+    Map<String, String> partition =
+        Collections.singletonMap(JdbcSourceConnectorConstants.TABLE_NAME_KEY, name);
+    return new SourceRecord(partition, null, name, null, null, record);
   }
-
 }
--- a/src/main/java/io/confluent/copycat/jdbc/DataConverter.java
+++ b/src/main/java/io/confluent/copycat/jdbc/DataConverter.java
@@ -16,6 +16,9 @@
 
 package io.confluent.copycat.jdbc;
 
+import org.apache.kafka.copycat.data.Schema;
+import org.apache.kafka.copycat.data.SchemaBuilder;
+import org.apache.kafka.copycat.data.Struct;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -30,10 +33,6 @@ import java.sql.SQLXML;
 import java.sql.Timestamp;
 import java.sql.Types;
 
-import io.confluent.copycat.data.GenericRecord;
-import io.confluent.copycat.data.GenericRecordBuilder;
-import io.confluent.copycat.data.Schema;
-import io.confluent.copycat.data.SchemaBuilder;
 
 /**
  * DataConverter handles translating table schemas to Copycat schemas and row data to Copycat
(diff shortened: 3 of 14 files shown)


pom.xml
```diff
@@ -70,9 +70,14 @@
...
         <dependency>
-            <groupId>io.confluent</groupId>
+            <groupId>org.apache.kafka</groupId>
             <artifactId>copycat-api</artifactId>
-            <version>${copycat.version}</version>
+            <version>${kafka.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>copycat-data</artifactId>
```
ewencp: This artifact doesn't exist anymore. I wrapped this into the API jar.

src/main/java/io/confluent/copycat/jdbc/BulkTableQuerier.java
```diff
@@ -44,9 +44,10 @@ protected ResultSet executeQuery() throws SQLException {
 
   @Override
   public SourceRecord extractRecord() throws SQLException {
-    GenericRecord record = DataConverter.convertRecord(schema, resultSet);
+    Struct record = DataConverter.convertRecord(schema, resultSet);
     // TODO: key from primary key? partition?
-    return new SourceRecord(name, null, name, null, null, record);
+    Map<String, String> partition = new HashMap<>();
```
ewencp: Not critical, but for simple maps like this it can be nice to use `Collections.singletonMap` to keep it concise.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -67,48 +68,45 @@ public static GenericRecord convertRecord(Schema schema, ResultSet resultSet)
...
     }
-    return builder.build();
+    return struct;
   }
 
 
   private static void addFieldSchema(ResultSetMetaData metadata, int col,
-                                     SchemaBuilder.FieldAssembler<Schema> fields)
+                                     SchemaBuilder builder)
+                                     // SchemaBuilder.FieldAssembler<Schema> fields)
```
ewencp: Same cleanup here.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -45,21 +44,23 @@
   public static Schema convertSchema(String tableName, ResultSetMetaData metadata)
       throws SQLException {
     // TODO: Detect changes to metadata, which will require schema updates
-    SchemaBuilder.RecordBuilder<Schema> builder = SchemaBuilder.record(tableName);
-    SchemaBuilder.FieldAssembler<Schema> fields = builder.fields();
+    // SchemaBuilder.RecordBuilder<Schema> builder = SchemaBuilder.record(tableName);
```
ewencp: Commented out old code? Can we clean this up?

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -67,48 +68,45 @@ public static GenericRecord convertRecord(Schema schema, ResultSet resultSet)
...
       throws SQLException {
     // Label is what the query requested the column name be using an "AS" clause, name is the
     // original
     String label = metadata.getColumnLabel(col);
     String name = metadata.getColumnName(col);
     String fieldName = label != null && !label.isEmpty() ? label : name;
-    // Create the field for each type case to handle
-    SchemaBuilder.FieldBuilder<Schema> builder = fields.name(fieldName);
+
     // FIXME builder.aliases() or builder.doc()?
```
ewencp: Is there any info in the column metadata that could be used to fill out any of the other schema fields here? Or should we just remove this comment? Things like `doc` probably can't be filled in, but what about default value?

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -67,48 +68,45 @@ public static GenericRecord convertRecord(Schema schema, ResultSet resultSet)
...
 
     int sqlType = metadata.getColumnType(col);
-
+    boolean optional = false;
     // We can check for nullable types at once here, but because of the DSL for schema builder
```
ewencp: I think some of these comments are outdated and might get confusing. We should clean these out/update them.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -118,30 +116,30 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
       case Types.TINYINT:
```
ewencp: Can we match the type exactly for `TINYINT` and `SMALLINT`? Shouldn't those map to `int8` and `int16`?

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -150,10 +148,10 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
       // for single precision
       case Types.FLOAT:
```
ewencp: Same question here. Copycat has `float32` and `float64`.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -225,10 +222,10 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
       case Types.TIMESTAMP: {
         // Current approach uses UNIX time in milliseconds. May lose nanosecond precision
         // available in source type
-        if (nullableBuilder != null) {
-          nullableBuilder.longType().endUnion().noDefault();
+        if (optional) {
```
ewencp: We'll want to change this to a logical type once we have those in Copycat's API. For the time being, we might want to annotate specialized types like these at least with a name.

src/test/java/io/confluent/copycat/jdbc/JdbcSourceTaskTestBase.java
```diff
@@ -26,18 +28,26 @@
...
-import io.confluent.copycat.source.SourceTaskContext;
-import io.confluent.copycat.storage.OffsetStorageReader;
 
 public class JdbcSourceTaskTestBase {
 
   protected static String SINGLE_TABLE_NAME = "test";
+  protected static Map<String, Object> SINGLE_TABLE_PARTITION = new HashMap<>();
+
+  static {
+    SINGLE_TABLE_PARTITION.put("table", SINGLE_TABLE_NAME);
```
ewencp: The constant "table" is used all over the place. Refactor it into a constant in one of the Jdbc* classes?

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
Ishiihara: The JDBC float type is the same as double type. I think here we should combine them.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
Ishiihara: Done.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
Ishiihara: Can you elaborate what do you mean by annotation?

Ishiihara:
@ewencp Addressed most review comments except the one with logical type. Can you elaborate what do you mean by annotation here? Thanks!

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
ewencp: Good point. This trips me up since i usually think of Float and Double as being different precision, JDBC's Real just makes this all confusing.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -177,21 +180,20 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
...
+          builder.field(fieldName, Schema.OPTIONAL_STRING_SCHEMA);
         } else {
-          builder.type().stringType().noDefault();
+          builder.field(fieldName, Schema.STRING_SCHEMA);
         }
         break;
       }
 
       // Binary == fixed bytes
       case Types.BINARY: {
```
ewencp: Not too important, but since there's no difference between `binary` and `fixed` now, you can merge this case with the ones below it.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
ewencp: Ah, I just mean giving the schema a name. So something like `builder.name("io.confluent.copycat.jdbc.Timestamp")`. But we can also just hold off until we have the logical type in Copycat, which will have a name set to something like `org.apache.kafka.copycat.Timestamp` anyway.

src/main/java/io/confluent/copycat/jdbc/util/StringUtils.java
```diff
@@ -0,0 +1,45 @@
...
+public class StringUtils {
+
+  /**
+   * Generate a String by appending all the @{elements}, converted to Strings, delimited by
+   * @{delim}.
+   * @param elements list of elements to concatenate
+   * @param delim delimiter to place between each element
+   * @return the concatenated string with delimiters
+   */
+  public static <T> String join(Iterable<T> elements, String delim) {
```
ewencp: There's a version of this in o.a.k.common.utils.Utils that we might be able to use directly. Need to check if that's considered stable, public API or not.

ewencp:
@Ishiihara LGTM. Left a couple of comments for things that could be tweaked, but feel free to commit.

src/main/java/io/confluent/copycat/jdbc/util/StringUtils.java
ewencp: Conclusion from discussion offline: keep this version because the one in o.a.k.common.utils is not considered public API currently.

ewencp:
@Ishiihara Can we wrap this up and merge? I started working on other copycat-jdbc issues on top of this patch and would like to be able to submit PR for them soon.

I think there are only 1 or 2 comments I left, so this should be quick. One final thing we also need to do is bump the Kafka version to 0.9.0.0-SNAPSHOT since that changed in Kafka trunk.

Ishiihara:
@ewencp FYI, just merged the PR.

Issue #3, opened 2015-05-13
Table whitelist/blacklist options


ewencp:
The user should be able to filter the set of tables down, either by a whitelist (if they only want a couple of tables) or blacklist (if they just want to not load a couple).

ewencp:
Fixed by #18

Issue #2, opened 2015-05-13
Allow overriding the default SELECT * queries


ewencp:
Currently SELECT * is hard coded. The user should be able to override this, allowing them to filter columns, join with other tables, etc. This probably means an alternate mode where the connector doesn't divvy up tables anymore since the query could potentially do things like join tables such that the result set is no longer tied to a single table.

This may affect some other features, e.g. if keys are generated from primary keys by default, a join query could break that functionality.

ewencp:
Fixed by #21

PR #10, opened 2015-10-03
Poll in background for changes the set of tables and trigger reconfiguration when tables are added or removed.

### LINKED ISSUE #5: Support dynamic sets of tables
If the user does a CREATE TABLE or DROP TABLE, the Connector should pick this up and reconfigure tasks appropriately. This requires some polling in the Connector on a background thread to periodically check whether the table set has changed.

As part of this, we should verify the Tasks gracefully handle the DROP TABLE case. This should just result in a SQLException which we need to handle anyway, but we might be able to handle this specific case more gracefully if we can narrow the exception to being caused by a dropped table.

ewencp:
Fixes #5.

### DIFF
--- a/src/main/java/io/confluent/copycat/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/copycat/jdbc/JdbcSourceConnector.java
@@ -39,9 +39,10 @@ public class JdbcSourceConnector extends SourceConnector {
 
   private static final Logger log = LoggerFactory.getLogger(JdbcSourceConnector.class);
 
-  Properties configProperties;
-  JdbcSourceConnectorConfig config;
-  Connection db;
+  private Properties configProperties;
+  private JdbcSourceConnectorConfig config;
+  private Connection db;
+  private TableMonitorThread tableMonitorThread;
 
   @Override
   public void start(Properties properties) throws CopycatException {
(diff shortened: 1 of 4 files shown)


ewencp:
@Ishiihara Care to review? This moves all the table monitoring at the connector level to a background thread that can be unit tested separately from the rest of the connector functionality. I left a TODO in there for additional filtering when #3 is implemented (and probably further adjustment will be needed for #2). Unfortunately no system tests for this yet, we'll want to add those to muckrake after we add the baseline tests with a fixed set of tables.

src/main/java/io/confluent/copycat/jdbc/JdbcSourceConnector.java
```diff
@@ -70,28 +75,30 @@ public void start(Properties properties) throws CopycatException {
...
     }
+    return taskConfigs;
   }
 
   @Override
   public void stop() throws CopycatException {
+    log.info("Stopping table monitoring thread");
+    tableMonitorThread.shutdown();
+    try {
+      tableMonitorThread.join(10000);
```
Ishiihara: Shall we define an instance variable for the timeout?

src/main/java/io/confluent/copycat/jdbc/TableMonitorThread.java
```diff
@@ -0,0 +1,119 @@
...
+          return;
+        }
+      } catch (InterruptedException e) {
+        log.error("Unexpected InterruptedException, ignoring: ", e);
+      }
+    }
+  }
+
+  public List<String> tables() {
+    final long TIMEOUT = 10000;
```
Ishiihara: 10000L?

ewencp:
@Ishiihara Anything else to comment on here or is this good to merge?

PR #11, opened 2015-10-07
Handle common logical types that Copycat supports: Decimal, Date, Time, Timestamp.

ewencp:


ewencp:
@Ishiihara Small update to fill in Decimal, Date, Time, and Timestamp support. Depends on [KAFKA-2622](https://github.com/apache/kafka/pull/285) so we shouldn't merge until that patch has landed, but it should be easy to review before that.

I think this covers all the logical types we need to handle for now, but it'd be good if you could also scan through the remaining unimplemented ones while reviewing this and make sure I haven't missed anything important. I think all the remaining ones are kind of obscure.

src/main/java/io/confluent/copycat/jdbc/DataConverter.java
```diff
@@ -204,27 +210,31 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
...
+        }
+        builder.field(fieldName, timeSchemaBuilder.build());
         break;
       }
 
       // Timestamp is a date + time
       case Types.TIMESTAMP: {
-        // Current approach uses UNIX time in milliseconds. May lose nanosecond precision
-        // available in source type
+        SchemaBuilder tsSchemaBuilder = org.apache.kafka.copycat.data.Timestamp.builder();
```
Ishiihara: Do we need the full qualified name here?

Ishiihara:
Left minor comments. Otherwise LGTM.

PR #13, opened 2015-11-10
Update to handle Copycat -> Kafka Connect renaming and a few API changes.

ewencp:


ewencp:
@Ishiihara or @miguno -- whoever gets to this first. Mostly straightforward renaming, but with the added `version()` API, I had to add some support for embedding the version number in the jar and then looking it up dynamically. Also a bunch of easy `Properties` -> `Map<String, String>` conversions.

ewencp:
@gwenshap Or you could review this if you have a chance. I pinged @miguno because I thought he needed it to unblock packaging, but he may not have gotten to these repos yet.

miguno:
LGTM.  Tests passed, too.

miguno:
Note that I also bumped the Kafka version to `0.9.0.0` in preparation of the Kafka 0.9.0.0 release (where the release branch, albeit it's still pre-release, already dropped the `-SNAPSHOT` suffix).

PR #14, opened 2015-11-12
Use database-specified quote characters instead of assuming double quotes.

ewencp:


ewencp:
@gwenshap Mind reviewing? Pretty straightforward and fixed an issue with the muckrake test I'm developing.

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
```diff
@@ -154,5 +155,17 @@ public static String getAutoincrementColumn(Connection conn, String table) throw
...
   }
+
+  /**
+   * Get the string used for quoting identifiers in this database's SQL dialect.
+   * @param connection the database connection
+   * @return the quote string
+   * @throws SQLException
+   */
+  public static String getIdentifierQuoteString(Connection connection) throws SQLException {
+    String quoteString = connection.getMetaData().getIdentifierQuoteString();
```
gwenshap: Neat! I didn't realize you can get this from the DB. I always made users configure this...

src/main/java/io/confluent/connect/jdbc/TimestampIncreasingTableQuerier.java
```diff
@@ -92,33 +94,51 @@ protected void createPreparedStatement(Connection db) throws SQLException {
       //  timestamp 1235, id 22
       //  timestamp 1236, id 23
       // We should capture both id = 22 (an update) and id = 23 (a new row)
-      builder.append(" WHERE (\"");
+      builder.append(" WHERE (");
+      builder.append(quoteString);
```
gwenshap: What I did in the past is have a utility function quoteIdentifier(String identifier) that returns a quoted string. 

I suspect it may and up more readable and less error-prone.

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
```diff
@@ -154,5 +155,17 @@ public static String getAutoincrementColumn(Connection conn, String table) throw
...
+    quoteString = quoteString == null ? "" : quoteString;
```
wushujames: If quote string is null, you use the empty string? I'm reading this from a phone, so I might be reading it wrong.

src/main/java/io/confluent/connect/jdbc/BulkTableQuerier.java
```diff
@@ -36,7 +36,8 @@ public BulkTableQuerier(String name) {
 
   @Override
   protected void createPreparedStatement(Connection db) throws SQLException {
-    stmt = db.prepareStatement("SELECT * FROM \"" + name + "\"");
+    String quoteString = JdbcUtils.getIdentifierQuoteString(db);
+    stmt = db.prepareStatement("SELECT * FROM " + quoteString + name + quoteString);
```
gwenshap: I'm not sure this will 100% work.
Table name can be "mytable" in which case your quoting is fine, but it can also be "myschema.mytable" which is quoted "myschema"."mytable".

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
ewencp: Yes, because this is always used to add quotes around identifiers, if the DB doesn't want quotes then it's harmless to use an empty string. It's easier to resolve the null issue here than have a bunch of null checks anywhere we use this.

gwenshap: http://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getIdentifierQuoteString()
"This method returns a space " " if identifier quoting is not supported."

Because databases make no sense at all :(

ewencp: Haha, as if you can trust it to be implemented properly in the first place. Some JDBC drivers apparently return null.

src/main/java/io/confluent/connect/jdbc/TimestampIncreasingTableQuerier.java
ewencp: Not sure which is better since the string builder also avoids lots of extra copying that we'll get with a quoteString helper, but this is rarely executed anyway so I've changed to to make it less error prone.

gwenshap:
LGTM.

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
```diff
@@ -131,7 +131,8 @@ public static String getAutoincrementColumn(Connection conn, String table) throw
     log.trace("Falling back to SELECT detection of auto-increment column for {}:{}", conn, table);
     Statement stmt = conn.createStatement();
     try {
-      rs = stmt.executeQuery("SELECT * FROM \"" + table + "\" LIMIT 1");
+      String quoteString = getIdentifierQuoteString(conn);
+      rs = stmt.executeQuery("SELECT * FROM " + quoteString + table + quoteString + " LIMIT 1");
```
wushujames: You don't need to worry about SQL injection at all, right? All these values like table name and timestamp.column.name are coming from config files, which you set yourself, so the worst you can do is SQL inject against yourself, likely against against a MySQL server that you already have permissions for?

Relevant XKCD post: https://xkcd.com/327/

ewencp: At the moment, these are all coming from a query against the database so there shouldn't be an issue. When I add table whitelist/blacklists, there will be user input, but that's a trusted user (you're also presumably giving the connection info, including password, via the same set of configs).

In general, while I think there are definitely security concerns in connectors, it's unlikely to be an issue in the first iteration of any connector since there are really only a small number of agents involved (Kafka, Kafka Connect, the other system, and the user submitting the connector).

PR #16, opened 2015-11-13
Initial systest fixes

ewencp:


ewencp:
@gwenshap Mind reviewing? There are 3 changes here, but since they're so small I figured they are worth wrapping into a single review.

Tests for most of these, are, unfortunately hard which is why I caught them with system tests. The unit tests use Derby and so cannot get perfect coverage of JDBC types. Doing anything time related that has to get the current time is going to be tricky to test with the DB... If you have ideas for how to add unit tests for any of these, I'm all ears :)

gwenshap:
LGTM.

Checking specific types in system tests makes sense.

PR #18, opened 2015-11-13
CC-14: Add whitelist/blacklist for tables.

ewencp:


ewencp:
@gwenshap Ready for review when you have a chance.

gwenshap:
LGTM.

If a Set type is generally useful, we could add it to ConfigDef and avoid the conversion.

PR #19, opened 2015-11-13
Make query mode a required configuration.

ewencp:


ewencp:
@gwenshap Found this out the hard way with some huuuuge log files from the output of running with defaults. I think this was just the default I stuck in because the bulk querier was the first implementation I added, but it's not a very good default since it usually isn't what you want and if you run with it accidentally, it just keeps dumping your table over and over again. Seems better to just make it required to be set just like the DB connection info.

gwenshap:
LGTM

PR #21, opened 2015-11-17
CC-12: Support user-specified queries.

ewencp:


ewencp:
@gwenshap This also adds a topic/topic-prefix setting because user specified queries don't exactly make it easy to come up with a topic to send data to. Should have had this anyway since table names often won't be unique, so being able to specify a prefix for table-based jobs is also obviously useful and important.

PR #22, opened 2015-11-18
Add check for NOT NULL columns on target tables when specified as increasing or timestamp column for offsets.

ewencp:
Because this is checked before making any queries, this is only checked for
table-based copying. Custom queries would either require specifying a
table.column format for the check, which wouldn't guarantee it even matched up
with the query, or require parsing or executing the query to perform the
validation before normal polling could start.


ewencp:
@gwenshap The most important limitation is described in the commit comment, let me know what you think. Without filling out more complete support for schema/table/column format support across all the settings, I'm not sure there's a good solution for user-specified queries.

gwenshap:
LGTM.

PR #23, opened 2015-11-19
Fix conversions of nulls for primitive types, add tests of null and non-null value conversions, and add missing tests of Date, Time, Timestamp, and Decimal.

ewencp:


ewencp:
@gwenshap Mostly straightforward. For primitive types, JDBC requires checking `wasNull` and we were missing that, so I filled in tests to validate handling of `NULL` for all types, as well as some missing types that somehow got skipped over when we added logical types.

hachikuji:
@ewencp LGTM

PR #24, opened 2015-11-30
CC-59: Add JDBC connector docs covering quickstart, key features, JDBC driver setup and config templates, and auto-generated config options.

ewencp:


ewencp:
@gwenshap @Ishiihara Docs for the JDBC connector. Anything missing a) in the template that all connector docs should have or b) anything specific for the JDBC connector?

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+
+Quickstart
+----------
+
+To see the basic functionality of the connector, we'll copy a single table from a local SQLite
+database. In this simple example, we'll assume each entry in the table is assigned a unique ID
+and is not modified after creation. Start by creating a database (you'll need to install SQLite
+if you haven't already)::
+
+   $ sqlite3 test.db
```
gwenshap: We are using SQLite for quickstart because thats the driver we can bundle? 
Because in terms of "quickstart" this is a database our users are least likely to be familiar with.

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+   connection.url=jdbc:sqlite:test.db
+   mode=increasing
+   increasing.column.name=id
+   topic.prefix=test-sqlite-jdbc-
+
+The first few settings are common settings you'll specify for all connectors. The ``connection.url``
+specifies the database to connect to, in this case a local SQLite database file. The ``mode``
+indicates how we want to query the data. In this case, we have an auto-incrementing unique
+ID, so we select ``increasing`` mode where each query uses the largest ID previously seen to limit
+new results to newly created rows. The ``increasing.column.name`` setting specifies the column in
```
gwenshap: I think the whole thing would be a lot clearer (and almost self-documenting) if we renamed "increasing" to "auto-incrementing" or "incrementing". We just spent 3 sentences explaining basically that increasing == incrementing :)

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+
+   $ ./bin/connect-standalone ./config/connect-standalone.properties quickstart-sqlite.properties
+
+You should see the process start up and log some messages, and then it will begin executing
+queries and sending the results to Kafka. In order to check that it has copied the data that was
+present when we started Kafka Connect, start a console consumer, reading from the beginning of
+the topic::
+
+   $ ./bin/kafka-avro-console-consumer --new-consumer --bootstrap-server localhost:9092 --topic test-sqlite-jdbc-accounts --from-beginning
+   {"id":1,"name":{"string":"alice"}}
```
gwenshap: I'm confused, is "id" a type here? Should we explain something about the output format here?

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+If you switch back to the console consumer, you should see the new record added (and,
+importantly, the old entries are not repeated)::
+
+   {"id":3,"name":{"string":"cathy"}}
+
+Note that the default polling interval is 5 seconds, so it may take a few seconds to show up.
+Depending on your expected rate up updates or desired latency, a smaller poll interval could be
+used to deliver updates more quickly.
+
+Of course, all the normal features of Kafka Connect, including offset management and fault
```
gwenshap: Link to KafkaConnect docs here so people will be able to check out the "normal features".

Also, s/normal/awesome/ :)

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+Kafka Connect uses `partitioned streams` as its core data model. In order to map to Kafka Connect
+records, the JDBC connector must model its data as a set of partitions, each with an ordered
+sequence of updates with unique offsets. In the case of a process restart or failure, these
+offsets are used to resume reading from the last position that the system can guarantee has been
+processed.
+
+To map to this model, the JDBC connector must map partitions and offsets. Partitions is
+handled on a per-table basis: each table represents a single partition and every insertion or
+update of a row corresponds to a single record in that partition. As a special case, custom
+queries (which disable table-based processing) are treated as a single-partition stream.
```
gwenshap: This is super confusing in the database context where "partitions" mean something else. 

Why is it important to explain the mapping to Kafka Connect Partitioned Streams? 

If we just need to explain the offsets, this can be much simpler. Something like:
"Kafka Connect tracks the latest record it retrieved from each table, so it can start in the correct location in the next iteration (or in case of a crash). custom-queries are treated as a single table." and from here explain the multiple options.

I think this is enough for a user-guide? 
We can have the more in-depth explanation at the end. I just don't want to lose anyone at this point ("oh, this is so academic and complex... I'll never figure this out, no point in reading on")

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+different approaches to defining offsets & queries for new or updated rows are supported.
+
+Incremental Query Modes
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* **Increasing Column**: In this mode, a single column is specified and contains a unique ID for
+  each row which increases with each insertion. Although it could be implemented differently, this
+  type of column is most frequently implemented as an ``AUTOINCREMENT`` column in a table where rows
+  are only inserted, but not updated. (Note that updates will *not* be handled properly in this
+  mode because they cannot be detected once the row has been processed for the first time.)
```
gwenshap: I'd add something to the effect of "this is a good fit when streaming fact tables in a data warehouse, since those are typically insert-only".

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+  tables. With a custom query, one of the other update automatic update modes can be used as long
+  as the necessary ``WHERE`` clause can be correctly appended (set the appropriate ``mode``
+  configuration). Alternatively, the specified query may handle filtering to new updates itself;
+  however, note that no offset tracking will be performed (unlike the automatic modes where
+  ``increasing`` and/or ``timestamp`` column values are recorded for each record), so the query
+  must track offsets itself.
+
+* **Bulk**: This mode is unfiltered and therefore not incremental at all. It will load all rows
+  from a table on each iteration. This can be useful if you want to periodically dump an entire
+  table where entries are eventually deleted and the downstream system can safely handle duplicates.
```
gwenshap: I'd add a note recommending that at least one of the tracking columns will have an index.

docs/index.rst
```diff
@@ -4,4 +4,315 @@ JDBC Connector
...
+Alternatively, you can set the ``CLASSPATH`` variable before running ``copycat-standalone`` or
+``copycat-distributed``. For example,::
+
+   $ CLASSPATH=/usr/local/firebird/* ./bin/copycat-distributed ./config/copycat-distributed.properties
+
+would add the JDBC driver for the Firebird database, located in ``/usr/local/firebird``, and allow
+you to use JDBC connection URLs like
+``jdbc:firebirdsql:localhost/3050:/var/lib/firebird/example.db``.
+
+Examples
```
gwenshap: Nice!

gwenshap:
Few comments, but looks great in general. I think the template is great.

docs/index.rst
ewencp: Yeah, but it's one which is trivial to setup and readily available pretty much everywhere. It's nice for demos since the database is just a file.

Anything DB we choose is going to make someone unhappy if it's not the one they prefer, so why not make as many people unhappy as possible!

But seriously, I've added a note about how you can swap out your favorite database.

docs/index.rst
ewencp: I don't think it was *that* long since I also wanted to specifically mention each JDBC connector config in the file. However, I've cleaned this up and renamed to incrementing since I agree that's probably clearer. We'll need a follow up patch in muckrake as well.

docs/index.rst
ewencp: Added explanation of the format -- definitely fair that not everyone will be familiar with the encoding . I've also adjusted the column for name to be nullable because SQLite has a kind of weird data model that causes the column to appear nullable even when marked as `NOT NULL`... This is probably good anyway because it's a good opportunity to discuss the Avro JSON encoding a bit, which I think often throws people off.

docs/index.rst
ewencp: Makes sense. Fixed to just give a rough intro explaining how the incremental querying works and then cleaned up the explanations of the different modes a bit to try to keep them as simple as possible.

docs/index.rst
ewencp: Great example, added.

docs/index.rst
ewencp: Added after the list of modes since it applies to all of them.

ewencp:
@gwenshap Updated. I made the change to `incrementing` since I think it makes a lot of sense. As a result, I also filed confluentinc/muckrake#74 to adjust the system test. Did you want to take another look?

gwenshap:
LGTM :+1:

PR #25, opened 2015-12-01
CC-51: Add SQLite, PostgreSQL, and MariaDB (and MySQL) JDBC drivers to packaged version so a few databases are supported out of the box.

ewencp:


ewencp:
@gwenshap @guozhangwang This adds the ones I think are reasonable and that work for licensing. The SQLite one is Apache 2, Postgres is BSD and MariaDB is LGPL and should cover MariaDB and MySQL.

This just pulls in the dependencies for packaging (you can run `mvn package` and validate they are in the right `share/java/` directory). We really need to also do validation once we've packaged them as well. We could potentially do that in muckrake. I think we could easily update the MySQL tests to use this one instead of the one we've manually pulled into muckrake until now. Postgres would require also creating a new Postgres ducktape service, so we may just need to manually validate for now.

gwenshap:
LGTM.

PR #26, opened 2015-12-04
CP-109: Remove old configs used for testing and replace with a quickstart-compatible file.

ewencp:


ewencp:
@gwenshap trivial change to address the configs that appeared broken when you ran through the quickstart. I think just having the quickstart one is enough. We give more examples in the docs, so I'm not worried about also having files for them in the packaged version.

docs/index.rst
```diff
@@ -62,7 +63,7 @@ table, the only output topic in this example will be ``test-sqlite-jdbc-accounts
 Now, run the connector in a standalone Kafka Connect worker in another terminal (this assumes
 Avro settings and that Kafka and the Schema Registry are running locally on the default ports)::
 
-   $ ./bin/connect-standalone ./config/connect-standalone.properties quickstart-sqlite.properties
+   $ ./bin/connect-standalone ./config/connect-standalone.properties etc/kafka-connect-jdbc/quickstart-sqlite.properties
```
gwenshap: while you are here... ./config/connect-standalone.properties does not exist.
We use etc/schema-registry/connect-avro-standalone.properties

ewencp: Good catch, fixed.

gwenshap:
LGTM

miguno:
@ewencp : Does the config change mean we need to update our corresponding smoke test, too?

miguno:
@ewencp Currently we have the following smoke test in packaging:

```bash
test_kafka_connect_jdbc() {
    machine=$1
    # Start a connector that uses simple SQLite settings.
    vagrant ssh $machine -- "connect-standalone /etc/schema-registry/connect-avro-standalone.properties /etc/kafka-connect-jdbc/sqlite-test.properties > /tmp/connect-jdbc.log"
    # In this case, we expect it to start up the connector, manage to make the
    # "connection" to the database (which it would create if it didn't exist),
    # but then fail because there aren't actually any tables in the database.
    sleep 5
    vagrant ssh $machine -- "grep 'Number of groups must be positive.' /tmp/connect-jdbc.log"
    # No cleanup necessary, the process exits due to connection errors. Validate it isn't running anymore.
    ! vagrant ssh $machine -- "ps ax | grep java | grep -i standalone | grep -v grep"
}
```

From what I understand (cf. smoke test snippet above) the smoke test does not need to be updated.

PR #37, opened 2016-01-18
Exposed Table Types as a config

Arttii:
Hi,

It seems the Table Types were not exposed as a config and the default query always used the default "Tables", so it was impossible to say extract data from Views.

Artyom


src/main/java/io/confluent/connect/jdbc/TableMonitorThread.java
```diff
@@ -102,7 +104,7 @@ private boolean updateTables() {
     synchronized (db) {
       final List<String> tables;
       try {
-        tables = JdbcUtils.getTables(db);
+        tables = JdbcUtils.getTables(db,table_types);
```
ewencp: If we're going to allow setting this, should we also support an empty list that gets all types? This might require a small change to `JdbcUtils.getTables()`, but otherwise seems simple to support.

ewencp:
@Arttii Basic patch looks good, but it'd be good to also have a unit test covering this functionality.

ewencp:
Also, it looks like the unit tests aren't updated to match the updated code -- at a minimum, `TableMonitorThreadTest` doesn't compile.

Arttii:
Sorry it seems i forgot to commit the tests. I will add it tomorrow morning.

Arttii:
Should be fixed now. The Unit tests seem to pass.

src/main/java/io/confluent/connect/jdbc/TableMonitorThread.java
Arttii: Isn't it easier to make the default setting to get all the types? Or is there a downside to that?

ewencp: I don't think that'll have the result many people expect. The goal is to enable broad copying by default (thus the reason for defaulting to copying all tables). Just considering views, it's not obvious you'd want to copy all views as well. But there are also other types of tables that I can't even think of good reasons for including -- synonym/aliases and system tables, for example.

In any case, at this point we also need to be careful to consider compatibility.

Arttii: I can only see a case of people wanting all tables and maybe some whitelisted/blacklisted views. Maybe just disallow an empty list and default to Tables if specified?

ewencp:
@Arttii Sorry, lost track of this. This does seem to fail some tests for me, e.g.

```
Running io.confluent.connect.jdbc.TableMonitorThreadTest
[2016-02-10 20:01:11,927] DEBUG Dropping database default by removing directory /Users/ewencp/confluent/copycat-jdbc.git/__test_database_default (io.confluent.connect.jdbc.EmbeddedDerby:171)
Exception in thread "Thread-9" java.lang.AssertionError:
  Unexpected method call getTables(org.apache.derby.impl.jdbc.EmbedConnection@1568657552 (XID = 169), (SESSIONID = 3), (DATABASE = __test_database_default), (DRDAID = null) , [TABLE]):
    getTables(org.apache.derby.impl.jdbc.EmbedConnection@1568657552 (XID = 169), (SESSIONID = 3), (DATABASE = __test_database_default), (DRDAID = null) ): expected: 1, actual: 0
	at org.easymock.internal.MockInvocationHandler.invoke(MockInvocationHandler.java:45)
	at org.powermock.api.easymock.internal.invocationcontrol.EasyMockMethodInvocationControl.invoke(EasyMockMethodInvocationControl.java:91)
	at org.powermock.core.MockGateway.doMethodCall(MockGateway.java:124)
	at org.powermock.core.MockGateway.methodCall(MockGateway.java:63)
	at io.confluent.connect.jdbc.JdbcUtils.getTables(JdbcUtils.java)
	at io.confluent.connect.jdbc.TableMonitorThread.updateTables(TableMonitorThread.java:107)
	at io.confluent.connect.jdbc.TableMonitorThread.run(TableMonitorThread.java:63)
```

It looks like there are a couple of tests with `expect` calls that are missing the new parameter.

Arttii:
Hi,

Well I also missed this entirely. I will take a look at the Test errors.

Artyom

gwenshap:
Also, since we are adding functionality, updating the doc with some information on the new config will be nice :)

Arttii:
So since the new platform was released I'll look into modifying the thing so the test pass finally.

Arttii:
So I changed the tests, had to add the types to the expect calls for some reason. I think this does not impact much, hopefully. 

I also added docs to the configs.

ewencp:
ok to test

docs/configuration_options.rst
```diff
@@ -98,4 +98,23 @@ Configuration Options
 
   * Type: boolean
   * Default: true
-  * Importance: low
\ No newline at end of file
+  * Importance: low
+
+``table.types``
```
ewencp: This file is actually automatically generated from the `ConfigDef`, so the changes here will get overwritten when someone regenerates this. Do you want to include these in the doc string too so it'll match? (You can generate the file by running the `main()` method in SchemaRegistryConfig)

ewencp:
@Arttii One last minor issue with the docs, but otherwise LGTM

PR #48, opened 2016-02-25
Delayed copy

gwenshap:


src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
```diff
@@ -74,6 +74,16 @@
       + "not be nullable.";
   public static final String TIMESTAMP_COLUMN_NAME_DEFAULT = "";
 
+  public static final String TIMESTAMP_DELAY_INTERVAL_MS_CONFIG = "timestamp.delay.interval.ms";
+  private static final String TIMESTAMP_DELAY_INTERVAL_MS_DOC =
+      "How long to wait after a row with certain timestamp appears before we include it in the result. "
+    + "You may choose to add some delay to allow transactions with earlier timestamp to complete. "
+    + "The first execution will fetch all available records (i.e. starting at timestamp 0) until current time minus the delay. "
+    + "every following execution will get data from the last time we fetched until current time minus a delay.";
```
Ishiihara: every -> Every. 
a delay ->  the delay

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
```diff
@@ -406,6 +444,7 @@ private void startTask(String timestampColumn, String incrementingColumn, String
     if (incrementingColumn != null) {
       taskConfig.put(JdbcSourceConnectorConfig.INCREMENTING_COLUMN_NAME_CONFIG, incrementingColumn);
     }
+    taskConfig.put(JdbcSourceConnectorConfig.TIMESTAMP_DELAY_INTERVAL_MS_CONFIG, delay.toString());
```
Ishiihara: We should check `delay!= null` before calling `toString()`.

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -60,15 +60,17 @@
   private Long timestampOffset;
   private String incrementingColumn;
   private Long incrementingOffset = null;
+  private Long timestampDelay;
```
ewencp: This never takes on a `null` value and isn't used as an `Object`, so should really be `long`.

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -60,15 +60,17 @@
...
   public TimestampIncrementingTableQuerier(QueryMode mode, String name, String topicPrefix,
                                            String timestampColumn, Long timestampOffset,
-                                           String incrementingColumn, Long incrementingOffset) {
+                                           String incrementingColumn, Long incrementingOffset, Long timestampDelay) {
     super(mode, name, topicPrefix);
     this.timestampColumn = timestampColumn;
     this.timestampOffset = timestampOffset;
     this.incrementingColumn = incrementingColumn;
     this.incrementingOffset = incrementingOffset;
+    this.timestampDelay = timestampDelay == null ? 0L : timestampDelay;
```
ewencp: Not a big deal, but if we're going to resolve this to 0 anyway, we could just force this with the option itself (which you already do) and pass in a `long` instead of `Long`. For the other fields, the fact that they are `null` is used to determine the mode we're in, whereas here it is not.

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -147,18 +149,25 @@ protected void createPreparedStatement(Connection db) throws SQLException {
   @Override
   protected ResultSet executeQuery() throws SQLException {
     if (incrementingColumn != null && timestampColumn != null) {
-      Timestamp ts = new Timestamp(timestampOffset == null ? 0 : timestampOffset);
-      stmt.setTimestamp(1, ts, UTC_CALENDAR);
-      stmt.setLong(2, (incrementingOffset == null ? -1 : incrementingOffset));
-      stmt.setTimestamp(3, ts, UTC_CALENDAR);
-      log.debug("Executing prepared statement with timestamp value = " + timestampOffset + " and incrementing value = " + incrementingOffset);
+      Timestamp startTime = new Timestamp(timestampOffset == null ? 0 : timestampOffset);
+      Timestamp endTime = new Timestamp(new Date().getTime() - timestampDelay);
```
ewencp: Hmm, so I thought about this when doing the initial implementation. I avoided using any timestamps from the node running this since then we're susceptible to time sync issues between the database and the host running the connector. And that is unfortunately pretty painful to unit test for correctness. Is it possible to rewrite this to do that computation via the SQL statement?

gwenshap: That was my first attempt too.
I failed to find syntax that works across the board, but will take another look there.

Any thoughts on whether we should also support "delayed ID"s? 
for similar reasons, transactions with newer sequences can finish later. On the other hand, unlike time, IDs don't move ahead uniformly so it can lead to some weird conditions.

ewencp: @gwenshap I figured that might be the problem. Do you know which syntax might work w/ different drivers? I'd imagine that long term we're going to need to handle quirks in different drivers (there's already a bit wrt sqlite iirc). If it's different for every one, that's one thing; if there's a small set we can use to get good coverage of different drivers, it might be worth trying to implement those couple of options.

Then again, system tests against popular dbs would really help here to figure out which work and which don't...

ewencp: How incredibly awful.

What about running a separate query to get the current time from the database? This definitely sucks since it's an extra query, but avoids sync issues, is a trivial query, and the bit of extra latency shouldn't really matter, right?

gwenshap: Yeah, that will work. It will also mean that if we want to support "ID delay" use-case, it won't be any extra overhead. 
Extra latency is much easier to deal with than time-sync issues, which are a huge time sink (see what I did there?)

miguno: > Extra latency is much easier to deal with than time-sync issues, which are a huge time sink (see what I did there?)

I just wrote this comment to say how awesome that line was, @gwenshap. :-)

gwenshap:
Updated to get time from DB. Since it is still fairly DB-specific, I'll be running tests on sqllite, mysql, Oracle to make sure this works.

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
wushujames: Coming out of lurk-mode just to applaud @gwenshap. Well done.

gwenshap:
This patch does not work with sqllite where it is impossible to use JDBC getTimestamp(). On the plus side, it isn't a regression - timestamp mode never worked in sqlite. Good thing our quickstart doc only includes incremental mode :)

gwenshap:
Manually validated with MySQL. (Couldn't get muckrake to work at the moment, but we'll see it in the nightly test after a commit...)

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
```diff
@@ -193,5 +196,41 @@ public static String getIdentifierQuoteString(Connection connection) throws SQLE
...
+
+    try {
+        rs = stmt.executeQuery(query);
+        if (rs.next())
+          return rs.getTimestamp(1, cal);
+        else
+          throw new ConnectException("Unable to get current time from DB using query " + query + " on database " + dbProduct);
+    } catch (SQLException e) {
+      log.error("Failed to get current time from DB using query " + query + " on database " + dbProduct, e);
+      throw e;
```
ewencp: Might want to wrap this in a ConnectException. In general, I think attempting to restrict exceptions that get thrown up to the framework to ConnectExceptions and then ensuring we have enough expressiveness in those will allow the framework to correclty respond is better than trying to handle arbitrary exceptions. (Of course we need to handle them since we can't guarantee that they won't be thrown, but trying to minimize them and also indicate intent will encourage correct handling of exceptions.)

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -144,21 +148,30 @@ protected void createPreparedStatement(Connection db) throws SQLException {
...
+              + " and incrementing value = " + incrementingOffset);
     } else if (incrementingColumn != null) {
       stmt.setLong(1, (incrementingOffset == null ? -1 : incrementingOffset));
       log.debug("Executing prepared statement with incrementing value = " + incrementingOffset);
     } else if (timestampColumn != null) {
-      Timestamp ts = new Timestamp(timestampOffset == null ? 0 : timestampOffset);
-      stmt.setTimestamp(1, ts, UTC_CALENDAR);
-      log.debug("Executing prepared statement with timestamp value = " + timestampOffset);
+      Timestamp startTime = new Timestamp(timestampOffset == null ? 0 : timestampOffset);
+      Timestamp endTime = new Timestamp(new Date().getTime() - timestampDelay);
```
ewencp: Missed updating this one -- should use `JdbcUtils.getCurrentTimeOnDB()` like the `timestamp+increasing` mode.

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
```diff
@@ -178,6 +179,40 @@ public void testTimestamp() throws Exception {
...
+    db.createTable(SINGLE_TABLE_NAME,
+            "modified", "TIMESTAMP NOT NULL",
+            "id", "INT");
+
+    db.insert(SINGLE_TABLE_NAME, "modified", JdbcUtils.formatUTC(new Timestamp(10L)), "id", 1);
+
+    startTask("modified", null, null, 4L);
+    verifyTimestampFirstPoll(TOPIC_PREFIX + SINGLE_TABLE_NAME);
+
+    Long currentTime = new Date().getTime();
```
ewencp: Is this leftover because one of the cases not changing from using `Date().getTime()`?

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
gwenshap: So, we are definitely propagating SQLExceptions from executeQuery (where this is called) all the way up to poll() where we log them as "Failed to run query"...

If we want to handle this failed query differently than others, than having a separate exception makes sense, otherwise, it may get a bit confusing (Why do some query failures throw different exception than others?)

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
gwenshap: The delay will be relative to the time on the DB (which, in the test, will be local time and not 1970), so we still need timestamps in the records that are more recent.

gwenshap:
@ewencp Mind taking another look (on Monday!) - I tested on MySQL and Oracle and resolved most of the issues you pointed out.

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
ewencp: Makes sense.

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -60,15 +61,18 @@
   private Long timestampOffset;
   private String incrementingColumn;
   private Long incrementingOffset = null;
+  private long timestampDelay;
+  private PreparedStatement currentTimeStmt;
```
ewencp: Unused

ewencp:
@gwenshap One trivial cleanup, otherwise LGTM!

gwenshap:
nitpicker :)

PR #50, opened 2016-03-04
Bump version to 3.0.0-SNAPSHOT and Kafka dependency to 0.10.0.0-SNAPSHOT

ewencp:


ewencp:
@miguno @ijuma @gwenshap @norwood Updating versions to clean up issues with system tests to address some issues in confluentinc/muckrake#90. Personally verified on that PR that it fixes issues w/ incompatible changes. These changes will be required for release anyway, modulo the SNAPSHOT suffix. Note that this upgrades all deps to 0.10.0.0-SNAPSHOT. Previous discussion left it unclear whether we might want to depend on older versions in some cases, but it simplifies things for now if we maintain one version across all packages.

ijuma:
LGTM

PR #51, opened 2016-03-08
Add config to disable non-null checks.

criccomini:


criccomini:
We have columns that are BIGINT (UTC seconds), and nullable. In some cases, it is impractical/impossible for us to ALTER the table to add NOT NULL. Without this patch, we can't use JDBC connector.

gwenshap:
LGTM. Thanks for contributing, @criccomini.

PR #53, opened 2016-03-16
adding checkstyle checks and the traditional minor fixes related

gwenshap:
The style is similar to Kafka, with small differences:
* No import controls
* indentation of 2
* No "unnecessary parenthesis" check - we use unnecessary parenthesis on occasion (mostly for trenary ops) and I think its good
* Different headers (duh)


ewencp:
@gwenshap Seems the `checkstyle.xml` file is missing?

ewencp:
And +1 to removing the unnecessary parentheses check -- that one drives me nuts in Kafka since I like it for clarity w/ ternary operator.

ewencp:
@gwenshap LGTM. This patch makes me very happy -- we should give this treatment to all our repos.

PR #55, opened 2016-03-18
Use new config definition

Ishiihara:


Ishiihara:
@ewencp JDBC connector using new config definition.

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
```diff
@@ -16,30 +16,40 @@
...
                                                      + "each table.";
   public static final int POLL_INTERVAL_MS_DEFAULT = 5000;
+  private static final String POLL_INTERVAL_MS_DISPLAY = "Poll Interval (ms)";
 
   public static final String BATCH_MAX_ROWS_CONFIG = "batch.max.rows";
   private static final String BATCH_MAX_ROWS_DOC =
       "Maximum number of rows to include in a single batch when polling for new data. This "
       + "setting can be used to limit the amount of data buffered internally in the connector.";
   public static final int BATCH_MAX_ROWS_DEFAULT = 100;
+  private static final String BATCH_MAX_ROWS_DISPLAY = "Batch Max Rows";
```
gwenshap: How many characters do we have? Perhaps "Maximum number of rows per batch"? or "Rows per batch limit"?
Display should probably be close to human english....

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
```diff
@@ -54,6 +64,7 @@
       + "  * timestamp+incrementing - use two columns, a timestamp column that detects new and "
       + "modified rows and a strictly incrementing column which provides a globally unique ID for "
       + "updates so each row can be assigned a unique stream offset.";
+  private static final String MODE_DISPLAY = "Mode";
```
gwenshap: Incremental mode?

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
```diff
@@ -90,16 +93,19 @@
       + "configurations to start polling for data in added tables or stop polling for data in "
       + "removed tables.";
   public static final long TABLE_POLL_INTERVAL_MS_DEFAULT = 60 * 1000;
+  private static final String TABLE_POLL_INTERVAL_MS_DISPLAY = "Table Poll Interval (ms)";
```
gwenshap: Data dictionary monitor interval (ms), or metadata monitor interval (ms). We are not actually polling a table here...

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
Ishiihara: We can also have bulk mode.  May be Table Updating Mode?

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
Ishiihara: The length of the display name should be close to the length of the typical value for the config.  How about "Max rows per batch" ?

gwenshap:
LGTM.

PR #59, opened 2016-04-04
Add schema evolution in doc

Ishiihara:
@ewencp @gwenshap @hachikuji

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -315,3 +315,45 @@ Configuration Options
   * Type: long
   * Default: 60000
   * Importance: low
+
+
+Schema Evolution
+----------------
+
+The JDBC connector supports schema evolution when the Avro converter is used. When there is a
+change in a database table schema, the JDBC connector can detect the change, create a new Kafka
+Connect schema and try to register a new Avro schema in the Schema Registry. Whether we can
+successfully register the schema or not depends on the compatibility level of the Schema Registry,
+which is backward by default.
+
+For example, if we remove a column from a table, the change is backward compatible and the
+corresponding Avro schema can be successfully registered in the Schema Registry. If we modify
+the database table schema to change a column type or add a column, when the Avro schema is
+registered to the Schema Registry, it will be rejected as the changes are not backward compatible.
+
+You can change the compatibility level of Schema Registry to allow incompatible schemas or other
+compatibility levels. There are two ways to do this:
+
+* Set the compatibility level for subjects which are used by the connector using
+  ``PUT /config/(string: subject)``. The subjects have format of ``topic-key`` and ``topic-value``
+  where the ``topic`` is determined by ``topic.prefix`` config and table name.
+
+* Configure the Schema Registry to use other schema compatibility level by setting
+  ``avro.compatibility.level`` in Schema Registry. Note that this is a global setting that applies
+  to all schemas in the Schema Registry.
+
+However, due to the limitation of the JDBC API, some compatible schema changes may be treated as
+incompatible change. For example, adding a column with default value is a backward compatible
+change. However, limitations of the JDBC API make it difficult to map this to default
+values of the correct type in a Kafka Connect schema, so the default values are currently omitted.
+The implications is that even some changes of the database table schema is backward compatible, the
+schema registered in the Schema Registry is not backward compatible as it doesn't contain a default
+value.
+
+If the JDBC connector is used together with the HDFS connector, there are some restrictions to schema
+compatibility as well. When Hive integration is enabled, schema compatibility is required to be
+backward, forward and full to ensure that the Hive schema is able to query the whole data under a
+topic. As some compatible schema change will be treated as incompatible schema change, those
+changes will not work as the resulting Hive schema will not be able to query the whole data for a
+topic.
+


docs/index.rst
```diff
@@ -315,3 +315,43 @@ Configuration Options
   * Type: long
   * Default: 60000
   * Importance: low
+
+
+Schema Evolution
+----------------
+
+The JDBC connector supports schema evolution. When there is a change in a database table schema,
+the JDBC connector can detect the change, create a new Kafka connect schema and try to register
```
ewencp: Kafka Connect should be capitalized

docs/index.rst
```diff
@@ -315,3 +315,43 @@ Configuration Options
...
+a new Avro schema in the schema registry. Whether we can successfully register the schema or not
+depends on the compatibility level of the schema registry which is backward by default.
```
ewencp: add a comma before "which"

docs/index.rst
```diff
@@ -315,3 +315,43 @@ Configuration Options
...
+
+For example, if we remove a column from the table, the change is a backward compatible and the
+corresponding Avro schema can be successfully registered in the schema registry. If we modify
+the database table schema to change a column type or add a column, when the Avro schema is
+registered to the schema registry, it will be rejected as the changes are not backward compatible.
+
+You can change the compatibility level of schema registry to allow non compatible schema or other
```
ewencp: non compatible schema -> incompatible schemas

docs/index.rst
```diff
@@ -315,3 +315,43 @@ Configuration Options
...
+compatibility levels. There are two ways to do this:
+
+* Configure the schema registry to use other schema compatibility level by setting
+  ``avro.compatibility.level`` in schema registry. Note that this is a global setting that applies
+  to all schemas in the schema registry.
+
+* Set the compatibility level for subjects which are used by the connector using
```
ewencp: We should probably put this one first since it is really the preferable way to change this setting.

docs/index.rst
```diff
@@ -315,3 +315,43 @@ Configuration Options
...
+  ``PUT /config/(string: subject)``. The subjects have format of ``topic-key`` and ``topic-value``
+  where the ``topic`` is determined by ``topic.prefix`` config and table name.
+
+However, due to the limitation of the JDBC API, some compatible schema changes may be treated as
+non compatible change. For example, adding a column with default value is a backward compatible
+change. However, there is no JDBC API to get the default value for a column. Even the change of
```
ewencp: This isn't strictly true -- there is COLUMN_DEF in the database metadata. We might need to phrase this differently. (I think there are limitations, e.g. it is the string form of the default, not sure what you'd do with defaults that are a function such as CURRENT_TIMESTAMP, etc.)

ewencp:
@Ishiihara In addition to other comments, we should make sure capitalization of Schema Registry is consistent with the rest of the documentation.

Ishiihara:
@ewencp Addressed the review comments and add more explanation on the complexity of getting the default table column values.

docs/index.rst
```diff
@@ -315,3 +315,48 @@ Configuration Options
   * Type: long
   * Default: 60000
   * Importance: low
+
+
+Schema Evolution
+----------------
+
+The JDBC connector supports schema evolution. When there is a change in a database table schema,
```
ewencp: Oh, one last point I realized -- do we want to mention that this is specific to using `AvroConverter`? If we're not using `AvroConverter`, then obviously a lot of this doesn't apply... The first set of statements up to the new Kafka Connect schema make sense, but everything after assumes AvroConverter.

docs/index.rst
```diff
@@ -315,3 +315,48 @@ Configuration Options
...
+the JDBC connector can detect the change, create a new Kafka Connect schema and try to register
+a new Avro schema in the Schema Registry. Whether we can successfully register the schema or not
+depends on the compatibility level of the Schema Registry, which is backward by default.
+
+For example, if we remove a column from the table, the change is a backward compatible and the
```
ewencp: change is a backward -> change is backward

docs/index.rst
```diff
@@ -315,3 +315,48 @@ Configuration Options
...
+  ``PUT /config/(string: subject)``. The subjects have format of ``topic-key`` and ``topic-value``
+  where the ``topic`` is determined by ``topic.prefix`` config and table name.
+
+* Configure the Schema Registry to use other schema compatibility level by setting
+  ``avro.compatibility.level`` in Schema Registry. Note that this is a global setting that applies
+  to all schemas in the Schema Registry.
+
+However, due to the limitation of the JDBC API, some compatible schema changes may be treated as
+incompatible change. For example, adding a column with default value is a backward compatible
+change. However, there is no unified JDBC API to get the default value from the ``ResultSetMetaData``.
```
ewencp: I don't know if we actually need that much detail about the implementation -- maybe just saying at a high level why we can't do it is enough. Maybe something like "For example adding a column with a default value is a backward compatible change. However, limitations of the JDBC API make it difficult to map this to default values of the correct type in a Kafka Connect schema, so the default values are currently omitted."

docs/index.rst
```diff
@@ -315,3 +315,48 @@ Configuration Options
...
+In some cases, we may be able to get the default value from ``COLUMN_DEF`` when reading the
+``ResultSet`` but not for all cases. For example, some default values are determined by calling
+functions in the database such as ``CURRENT_TIMESTAMP``. With all the complexities of getting
+the default values, the JDBC connector ignores the table column default values when creating the
+Kafka Connect schema. The implications is that even some changes of the database table schema is
+backward compatible, the schema registered in the Schema Registry is not backward compatible as it
+doesn't contain a default value.
+
+If the JDBC connector is used together with the HDFS connector, there are some limitations to schema
```
ewencp: Maybe say "restrictions on schema compatibility" rather than "limitations to schema compatibility". Limitations makes it sound like it is the connector's fault, but it's really a restriction imposed by Hive.

ewencp:
@Ishiihara Left some follow up items, but LGTM. Feel free to merge once those issues are resolved.

PR #60, opened 2016-04-05
Oracle NUMBER as incrementing column

tdpauw:
The implementation uses the scale of the NUMBER column to determine whether to create an `INT64_SCHEMA` or a `BYTES` schema.

In order to be able to have unit tests that are independent of a running Oracle instance, I made use of the HSQLDB in memory database using the Oracle compatibility mode.
As a result I had to create a second `EmbeddedHSQLDB` next to the existing `EmbeddedDerby` and an extra `JdbcSourceTaskTestOracleBase` that copies `JdbcSourceTaskTestBase`.

There is definitely room for refactoring like: 
* introducing a base class/interface for the embedded database classes,
* this way we could have one base `JdbcSourceTaskTestBase` class where every implementing class (i.e. `JdbcSourceTaskXXXTest`) can then define which database to use,
* lastly, the 2 big switch cases in `DataConverter` should be replaced by the use of polymorphism.

I did not want to attack these refactorings in this PR. Because I first wanted to have feedback about the provided implementation.

*Note*: this PR is created from the `v2.0.0.` tag.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,6 +59,7 @@
         <commons-io.version>2.4</commons-io.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
+        <hsqldb.version>2.3.3</hsqldb.version>
     </properties>
 
     <repositories>
@@ -131,6 +132,13 @@
             <version>${commons-io.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hsqldb</groupId>
+            <artifactId>hsqldb</artifactId>
+            <version>${hsqldb.version}</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
--- a/src/main/java/io/confluent/connect/jdbc/DataConverter.java
+++ b/src/main/java/io/confluent/connect/jdbc/DataConverter.java
@@ -65,8 +65,7 @@ public class DataConverter {
     Struct struct = new Struct(schema);
     for (int col = 1; col <= metadata.getColumnCount(); col++) {
       try {
-        convertFieldValue(resultSet, col, metadata.getColumnType(col), struct,
-                          metadata.getColumnLabel(col));
+        convertFieldValue(resultSet, col, struct, metadata);
       } catch (IOException e) {
         log.warn("Ignoring record because processing failed:", e);
       } catch (SQLException e) {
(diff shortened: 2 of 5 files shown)


ewencp:
@tdpauw Thanks for tackling this. A couple of quick questions before I dive into more detailed code review:

1. Is int64 guaranteed to be large enough? This is more a question of general SQL support than Oracle specifically.
2. This isn't a compatible change as is since it changes the schema, right? We need to think about compatibility, so this behavior might need to be controlled by an config option.
3. Is there a slightly more general set of conversions here? It seems like this isn't just about the case where `scale == 0`, rather its about whether you want a numeric representation at the cost of losing the exact schema. I think this might affect point 2, i.e. you might have the config support a couple of different modes, or at the least could probably also support converting columns with non-zero scale to doubles or something like that. It would be nice to avoid a huge number of configs (and compatibility challenges), so if we add a config for (2), it'd be nice if it can capture all the modes we think we might need.

tdpauw:
@ewencp Thank you for the feedback. You raised some good points there.

1. No Int64 is in fact not guaranteed to be large enough. Oracle NUMBER has a default precision of 38 (i.e. it has 38 digits). The biggest signed Int64 is 2^63 -1 = 9223372036854775807 which has 20 digits. So we loose data here.
2. Yes this is breaking compatibility. Because of point 1, I would therefore look for another implementation that does not break compatibility and still keep the Bytes schema instead of Int64. Now that we have unit tests, it'll be easier to find other solutions.
3. If I can tackle point 2, there will be no need for this point ;)

In addition I should add some test cases where the scale is not declared with the NUMBER declaration. Then it falls back to the default scale of zero. But I remember the JDBC driver then returned a negative value :/

shikhar:
I think #101 presents a better approach -- rely on precision in the metadata for `Types.NUMERIC` when the scale is 0 indicating integer value -- if the precision is < 20 we can safely map that to INT64 (and so on for the narrower int types).

tdpauw:
At time I submitted this PR this was a big problem for a customer who wanted to use Kafka Connect to stream database records to Kafka. In the mean time they've found another solution to achieve the same result. Therefore we didn't had any time to improve this PR. For which I am very sorry.
Therefore I will close this PR.

nik09541:
Hi @ewencp 

I have tried to configure Kafka JDBC connect in order to have streams from an existing table. The existing table unfortunately have a id field that has purging from another event(i.e the rows are not strictly incrementing). As of now we have created script to move the records to another custom table we created with a custom id and timestamp field. But this is not working.During the first insert into custom table with data from existing table there is no event seen on kafka topic. But on updating the timestamp column the event is noticed on connectror. I did try the bulk option, but then the processing goes in an endless loop while trying to read it from avro consumer (but the bulk table option is not going to take-off either as the audit tables are going to be huge).

insert into reference (select * from customer where customer_ref_id in (5004145,5004146,5004161,5004162,5004178));
Above insert registers no event in kafka connector as id id not strictly incrementing in customer table.

insert into reference (customer_ref_id ,updated_date) values (id_seq.nextval,current_timestamp);
This insert works as it is inserting data with a sequence into table. Event is noticed on kafka connector

I have tried the query mode also it also doesnt work.

Please help if we can poll out the existing data on kafka topic for a non incrementing exisiting id of a table which is having purging .

PR #67, opened 2016-04-22
Add support for single primary key column

gfodor:
Here's an attempt at adding support for the JDBC connect driver to send messages to Kafka with the primary key value for a table as the message key. I was not sure of the proper way to support composite primary keys, so the patch only adds support for tables that have a single primary key.

This is obviously a WIP, I wanted to get feedback on the composite primary key issue as well as the proper level of configurability for this feature. This seems like something that should be on by default, since it's likely the behavior of least surprise, but backwards-compatibility concerns obviously are there to consider as well.

This PR also adds the `incrementing.column.use_primary_key` option -- this is an option which will force the connector to assume the primary key is an incrementing column. For our use-case, this was a necessary addition, since our primary key columns are generated via a stored procedure (which generates an auto incremented value) but are not reported by the JDBC driver as auto incrementing, nor are all the primary key columns named the same so just setting `incrementing.column.name` cannot be used.


gfodor:
This actually isn't ready yet. Will re-open later.

gfodor:
[clabot:check]

ewencp:
ok to test

gfodor:
(I have been unable to get the test suite to run, FYI. Will work on adding tests etc for this but wanted to get feedback in the meantime.)

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
```diff
@@ -107,6 +107,12 @@
   public static final String TABLE_BLACKLIST_DEFAULT = "";
   private static final String TABLE_BLACKLIST_DISPLAY = "";
 
+  public static final String INCREMENTING_COLUMN_USE_PRIMARY_KEY_CONFIG = "incrementing.column.use_primary_key";
```
Ishiihara: incrementing.column.use_primary_key -> incrementing.column.use.primary.key

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
```diff
@@ -161,6 +167,8 @@ public static ConfigDef baseConfigDef() {
                 Importance.HIGH, MODE_DOC, MODE_GROUP, 1, Width.MEDIUM, MODE_DISPLAY, Arrays.asList(INCREMENTING_COLUMN_NAME_CONFIG, TIMESTAMP_COLUMN_NAME_CONFIG, VALIDATE_NON_NULL_CONFIG))
         .define(INCREMENTING_COLUMN_NAME_CONFIG, Type.STRING, INCREMENTING_COLUMN_NAME_DEFAULT, Importance.MEDIUM, INCREMENTING_COLUMN_NAME_DOC, MODE_GROUP, 2, Width.MEDIUM, INCREMENTING_COLUMN_NAME_DISPLAY,
                 MODE_DEPENDENTS_RECOMMENDER)
+        .define(INCREMENTING_COLUMN_USE_PRIMARY_KEY_CONFIG, Type.BOOLEAN, INCREMENTING_COLUMN_USE_PRIMARY_KEY_DEFAULT, Importance.MEDIUM, INCREMENTING_COLUMN_USE_PRIMARY_KEY_DOC, MODE_GROUP, 2, Width.MEDIUM, INCREMENTING_COLUMN_USE_PRIMARY_KEY_DISPLAY,
```
Ishiihara: We should change 2 to 3. The number represents the order of the config in the group.

gfodor:
Bump on this, I just pushed some additional tests and cleanup based on previous comments. We've been running this in production for the last month or so without issue.

src/main/java/io/confluent/connect/jdbc/BulkTableQuerier.java
```diff
@@ -78,6 +78,15 @@ public SourceRecord extractRecord() throws SQLException {
       default:
         throw new ConnectException("Unexpected query mode: " + mode);
     }
+
+    if (keyColumn != null && keySchema != null) {
```
ewencp: Check for both `keyColumn` and `keySchema` seems like it shouldn't be necessary -- if `keyColumn` isn't null, shouldn't we have thrown an exception earlier if we couldn't create the `keySchema`?

src/main/java/io/confluent/connect/jdbc/BulkTableQuerier.java
```diff
@@ -78,6 +78,15 @@ public SourceRecord extractRecord() throws SQLException {
...
+      Object key = record.get(keyColumn);
+
+      if (key != null) {
```
ewencp: Why is this structured such that you could fall through to the non-`keyColumn` case? If the key comes back `null` but the `keySchema` isn't optional, shouldn't that cause an error?

src/main/java/io/confluent/connect/jdbc/DataConverter.java
```diff
@@ -51,10 +51,17 @@
 
   public static Schema convertSchema(String tableName, ResultSetMetaData metadata)
       throws SQLException {
+    return convertSchema(tableName, null, metadata);
```
ewencp: Why the addition of the `columnName` to the other version of this method and passing `null` here? As far as I can tell, this seems like an unused addition.

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
```diff
@@ -107,6 +107,12 @@
   public static final String TABLE_BLACKLIST_DEFAULT = "";
   private static final String TABLE_BLACKLIST_DISPLAY = "";
 
+  public static final String INCREMENTING_COLUMN_USE_PRIMARY_KEY_CONFIG = "incrementing.column.use.primary.key";
```
ewencp: Here's another idea for how to expose this -- a mode that is `incrementing-primary-key` instead of a separate flag. I'm not sure if that's necessarily better -- I think we may end up with a bunch of combinations since we already have combinations of timestamps & primary keys.

So to make sure I understand the motivation here -- is the reason for the addition that you a) want to use `incrementing` mode, b) have many tables, and c) have primary key columns on those tables named differently? So this option allows you to capture in a smaller set of configs (or smaller set of connectors) all tables at once while the existing configs do not?

src/main/java/io/confluent/connect/jdbc/JdbcSourceTask.java
```diff
@@ -145,6 +149,35 @@ public void start(Map<String, String> properties) {
...
+      try {
+        primaryKeys = db.getMetaData().getPrimaryKeys(null, null, tableOrQuery);
+
+        if (primaryKeys.next()) {
+          String candidateKeyColumn = primaryKeys.getString(4);
+
+          if (!primaryKeys.next()) {
+            keyColumn = candidateKeyColumn;
+          } else {
+            log.warn("Composite primary key found in table {}, not supported", tableOrQuery);
```
ewencp: Should this be a warning or an error? Or maybe we need to hold on to this info until later to trigger errors for the case where you're using primary key as the incrementing column?

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -59,25 +59,31 @@
...
+    this.incrementingColumnUsePrimaryKey = incrementingColumnUsePrimaryKey;
   }
 
   @Override
   protected void createPreparedStatement(Connection db) throws SQLException {
-    // Default when unspecified uses an autoincrementing column
+    // Default when unspecified uses an autoincrementing column or primary key if primary key assumed to autoincrement.
     if (incrementingColumn != null && incrementingColumn.isEmpty()) {
-      incrementingColumn = JdbcUtils.getAutoincrementColumn(db, name);
+      if (incrementingColumn.isEmpty() && incrementingColumnUsePrimaryKey && keyColumn != null) {
```
ewencp: This duplicates the `incrementingColumn.isEmpty()` check that was already done for the `if` right above this.

Also, I think this is at least one location we could throw the exception if `incrementingColumnUsePrimaryKey` is true but `keyColumn` is false, right? Although maybe that's better done as a check in the constructor so we trigger the error as early as possible?

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -59,25 +59,31 @@
...
+        incrementingColumn = keyColumn;
```
ewencp: Unlike auto increment columns, primary key columns aren't actually guaranteed to have a compatible type for this use, right? Should we be doing anything to check the type of the column? (Although I suppose it might have to wait until we actually get the relevant schema back)

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
```diff
@@ -224,6 +230,15 @@ public SourceRecord extractRecord() throws SQLException {
       default:
         throw new ConnectException("Unexpected query mode: " + mode);
     }
+
+    if (keyColumn != null && keySchema != null) {
+      Object key = record.get(keyColumn);
+
+      if (key != null) {
```
ewencp: Same comments as for the `BulkTableQuerier` section containing the same code.

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
```diff
@@ -128,6 +128,29 @@ public void testManualIncrementing() throws Exception {
...
+
+    // Adding records should result in only those records during the next poll()
+    db.insert(SINGLE_TABLE_NAME, "id", 2);
+    db.insert(SINGLE_TABLE_NAME, "id", 3);
+
+    verifyPoll(2, "id", Arrays.asList(2, 3), false, true, TOPIC_PREFIX + SINGLE_TABLE_NAME);
+
+    PowerMock.verifyAll();
+  }
+
```
ewencp: Might be more relevant if exceptions are thrown due to missing primary keys, but should there be any failure tests?

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
```diff
@@ -191,7 +214,7 @@ public void testTimestampWithDelay() throws Exception {
 
     db.insert(SINGLE_TABLE_NAME, "modified", JdbcUtils.formatUTC(new Timestamp(10L)), "id", 1);
 
-    startTask("modified", null, null, 4L);
+    startTask("modified", null, null, null, 4l);
```
ewencp: Nit: the `L` is preferred since it's not easily mistakable for `1` like `l` can be.

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
```diff
@@ -418,16 +441,19 @@ public void testCustomQueryWithTimestamp() throws Exception {
   }
 
   private void startTask(String timestampColumn, String incrementingColumn, String query) {
-    startTask(timestampColumn, incrementingColumn, query, 0L);
+    startTask(timestampColumn, incrementingColumn, null, query, 0L);
   }
 
-  private void startTask(String timestampColumn, String incrementingColumn, String query, Long delay) {
+  private void startTask(String timestampColumn, String incrementingColumn, String usePrimaryKeyAsIncrementingColumn, String query, Long delay) {
```
ewencp: Is there a reason this is `String` instead of `boolean` (or `Boolean` to make the `toString()` conversion you might need easier)?

ewencp:
@gfodor This is looking good. I think the delta between this and supporting composite keys would actually be quite small -- basically just replacing the single `String keyColumn` with a `List<String>` (maybe wrapped up in a nicer class) and then converting that to a `Struct` schema while maintaining the degenerate single-column case as is.

src/main/java/io/confluent/connect/jdbc/TimestampIncrementingTableQuerier.java
gfodor: So actually I think it's ok to fall back on the default behavior if there was no primary key found -- I'm going to be opening another PR to enable support for views and since views typically don't have primary key metadata, this allows views to work since it falls back on a hardcoded incrementingColumn

ewencp: I see, that's reasonable. Might make sense to clarify that in the doc for `incrementing.column.use.primary.key` so the behavior is clear.

clumsy:
Any plans on merging this?

ewencp:
@clumsy Would like to, there's some stuff that needs to be cleaned up per the last review comments (and it looks like it now doesn't merge cleanly). Would love to get this in if @gfodor has some time to follow up on the remaining issues.

gfodor:
Yep sorry, I've been swamped at work. Going to try to get back to some of
these PRs soon!

src/main/java/io/confluent/connect/jdbc/BulkTableQuerier.java
gfodor: Yes, the schema check is redundant, I'll remove it.

src/main/java/io/confluent/connect/jdbc/BulkTableQuerier.java
gfodor: I think I was writing code defensively here in the case where the user specifies a key column but that column contains NULLs in the db for some reason. Obviously this seems unlikely but we need to make a decision on what to do. Should that be considered an error? I don't know if there is really a 'correct' behavior here, I guess it depends if we prefer the driver to be defensive or ensure all data gets loaded regardless of the configuration not making sense.

src/main/java/io/confluent/connect/jdbc/DataConverter.java
gfodor: Ah yes it's left over from a previous version of the code where I was calling `convertSchema` twice, once for the value schema and once for the key schema. I'll remove it.

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnectorConfig.java
gfodor: The reason I factored it this way is because the `mode` configuration seems orthogonal to the concern about what determines which column is the incrementing column. (Previously, there were two options for determining which column is the incrementing column: assume anything that is AUTO INCREMENT is the incrementing column, or specify a fixed incrementing column name.)

Ie, if we were to roll this into the `mode` config, you'd technically need two new options: `timestamp+incrementing-primary-key` and `incrementing-primary-key` -- this isn't terrible but if another mode were introduced that was in addition to `incrementing` and `incrementing+timestamp` that was incrementing column-dependent, you'd need two new `mode`s, etc.

src/main/java/io/confluent/connect/jdbc/JdbcSourceTask.java
gfodor: Yeah I'd argue that it's just a warning, since in general I don't think the JDBC driver should reject tables that have composite primary keys, it can simply fall back on the behavior it has now, where no key is specified for the record -- I will make sure there are appropriate errors if you specify the primary key as the incrementing column but the table does not have a primary key or has a composite primary key.

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
gfodor: Nope, I can refactor

src/test/java/io/confluent/connect/jdbc/JdbcSourceTaskUpdateTest.java
gfodor: Yes probably worth adding a test to cover the case where a composite primary key exists and use primary key for incrementing column is turned on, which should be a failure. Also based on your response above to the q about what to do in the case of NULL primary key values I'll add more.

gfodor:
Ugh, github swallowed all my comments on the PR due to the merge commit. I incorporated most of the feedback however there are still two sticking points:

First is that the configuration option I added for `incrementing.column.use.primary.key` likely should not be rolled into the `mode` since it seems to be an orthogonal concern. Currently there are two ways an incremently column is determined, either via the AUTO INCREMENT metadata or by a fixed name via `incrementing.column.name`, and this is a third. (If we were to roll this into the `mode`, we'd need two new modes, one for the `incrementing` mode and one for the `timestamp+incrementing` mode, which should show that this is a separate concern.) The current configuration is a bit strange, a cleaner, but non-backwards-compatible one would be to introduce `incrementing.column.type` which could be `auto_increment` (default), `primary_key`, or `fixed`, and `incrementing.column.name` is allowed if and only if `fixed` is specified.

The second set of unresolved issues in the PR are around error cases:
- I'd argue that a WARN is the proper logging for when a composite primary key is discovered in general, since it can fall back on the behavior of a table with no primary key, and not specify keys for the records. This may be surprising to users, but seems better than preventing users from syncing composite primary keyed tables, so a WARN seems to suffice. (I can update the log statement.)
- However if the user specifies that the primary key should be used for the incrementing column, but a singular primary key column is not found (either due to their not being one, or it being composite), this should be an error.
- Finally the question was brought up of what to do when there is a NULL value in a known primary key column. It seems the least surprising thing to do here is to still send that record to Kafka with a null key, but the alternative would be to emit an error.

Also I might have missed it but I didn't see any comments on the fact that this patch changes the semantics of records going to Kafka for any tables that have primary key columns: before this patch, they will have null keys, and after it they will have the key of the primary key column. If there is flexibility on backwards compatibility there, I feel strongly this is a positive change and is probably the expected behavior, but I wanted to make sure you're aware of the fact that users may run into issues where (for example) they are relying upon the random partitioning currently happening due to the null Kafka keys being sent for all records, regardless of their primary key.

gfodor:
Bump! @ewencp can you give me feedback on the above points so I can make a push for the finish line on this one?

nagarajG:
@gfodor I have pulled your changes and build the jar and using it. But i am not able deserialize the key. I tried various options like
   props.put("key.deserializer", "org.apache.kafka.common.serialization.IntegerDeserializer");
props.put("key.deserializer", "org.apache.kafka.common.serialization.LongDeserializer");
   props.put("key.deserializer", "org.apache.kafka.common.serialization.ByteArrayDeserializer");

But i am not able to deserialise the key from my consumer. Am i missing something? 

Thanks.

gfodor:
Another bump @ewencp can you please provide feedback on the above points so I can finish this up? Thanks!

gfodor:
Hi @nagarajG , I am not doing anything special -- they keys for me are
serialized as avro types. You'll want to make sure you create a key
serde, not a value serde. So you can use KafkaAvroSerializer. You'll
want to make sure you pass isKey=true to the configure method.

jozi-k:
Are there any plans to merge this into 3.1.0?

gfodor:
still waiting for feedback from someone at confluent on the q's above.

ewencp:
@gfodor Sorry, this is getting buried in my notifications with a million other GH updates.

1. You can also add the incrementing.column.type config in a backwards compatible way by having a null default instead and if that is used you fall back to the existing rules. That's also a path for deprecation of the old approach and eventually switching the default for incrementing.column.type -- include log warnings about the deprecation and then in a major release switch over to the new approach by default. This is how we would generally safely make config default changes.

2. WARN isn't great if they explicitly asked for use of the primary key and we're not going to be able to include it. That's the case I'd be most worried about since the user may not even want any records written if we can't give them the correct data (and unfortunately WARN is often ignored).

3. This is a good point. Probably this should be opt-in via config even if we eventually decide to change the default behavior in a major version. That said, the only way I can think of that people would be relying on the current behavior is if including the key would result in a bad distribution of records across partitions.

tegansnyder:
@gfodor if I'm reading your previous comments correctly are you implying that this PR would allow me to instruct the connector to use a non-auto-incrementing column by setting: `incrementing.column.type=fixed` and `incrementing.column.name=my_column`? This would be useful for scenarios where the source data doesn't contain a numeric auto-increment column.

larsduelfer:
Any progress on this topic? We are also looking for a solution to define a key for the messages generated by the JDBC Connector.

gharris1727:
Because this operation can be handled by SMTs (see [the article linked above](https://www.confluent.io/blog/simplest-useful-kafka-connect-data-pipeline-world-thereabouts-part-3/)) we wont support this operation in the connector itself.

PR #68, opened 2016-04-22
Re-create prepared statement on schema change for PostgreSQL

gfodor:


src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -150,6 +150,6 @@ public void stop() throws ConnectException {
 
   @Override
   public ConfigDef config() {
-    return JdbcSourceTaskConfig.config;
+    return JdbcSourceConnectorConfig.config;
```
gfodor: Not sure how this creeped into this patch, was not able to get rid of

Ishiihara:
@gfodor Can you rebase against master?

ewencp:
ok to test

src/main/java/io/confluent/connect/jdbc/TableQuerier.java
```diff
@@ -107,4 +121,11 @@ public int compareTo(TableQuerier other) {
       return this.name.compareTo(other.name);
     }
   }
+
+  private boolean exceptionIndicatesStalePreparedStatement(Connection db, SQLException e) throws SQLException {
+    // On PostgreSQL, queries with wildcards will be invalidated on schema changes
+    // and throw this error message.
+    return e.getMessage().contains("cached plan must not change result type") &&
```
ewencp: @gfodor This all seems reasonable. The thing I'm curious about is if we can do something better re: detecting the specific error that triggers this. First, is there a more specific exception we could catch that identifies the issue? And even if not, is there a better way to get the relevant info out of the exception that won't be sensitive to the specific formatting of the message? For example, https://github.com/rails/rails/commit/818d285305502cc6191a98400b43633f44394f6e seems to indicate there's a PG-specific error that (maybe?) we could detect.

gfodor: I took a look at the error code returned in the SQLException, and it looked to be zero. I'll see if maybe I can cast the exception to a PSQLException to get better info.

gfodor: Actually that probably won't work, because PSQLException isn't in scope of this library unless we want to add that dependency

gfodor:
Ping on this, any changes still to make? See note above re: not being able to get more granular.

src/main/java/io/confluent/connect/jdbc/TableQuerier.java
ewencp: I think the way you can make this work without requiring the runtime dependency for everyone is to add the dependency as `compile` or `provided` scope, check the class name at runtime and then only use the class in blocks where you've validated the class must be available at runtime. So roughly something like this:

```
catch (SQLException e) {
  if (e.getClass().getName() == "PSQLException") {
    PSQLException pge = (PSQLException) e;
    // check PSQLException-specific stuff
  }
}
```

Since you won't actually use the class unless the class name matches (which means it must be available in the classpath since it has been instantiated), this will work for instances that use PG and execute that block and other drivers which won't. This is basically a manual, runtime type check & validation that the class exists.

I'm not really sure if that's a good idea or not and how often this approach is used in practice. I think some users of libraries do similar things to detect which version of libraries they are running against to figure out what features they can use (or work around incompatibilities in the libraries). It just seems more reliable to me than relying on the message since messages seem more likely to change than the class name since the class name is used by "normal" users to catch exceptions.

I think I might have discussed this approach to handling cases like this with @gwenshap before. @gwenshap any input on this? Are there other examples you're aware of that use a similar solution?

src/main/java/io/confluent/connect/jdbc/TableQuerier.java
```diff
@@ -73,7 +77,17 @@ public boolean querying() {
...
       stmt = getOrCreatePreparedStatement(db);
-      resultSet = executeQuery();
+
+      try {
+        resultSet = executeQuery();
+      } catch (SQLException e) {
+        if (exceptionIndicatesStalePreparedStatement(db, e)) {
+          log.info("Stale prepared statement, re-creating.");
+          createPreparedStatement(db);
+          resultSet = executeQuery();
```
gfodor: leaving a note for myself here, throw e missing

daniel-shuy:
Any update on this PR?

daniel-shuy:
Its been more than 3 years. Any update on this?

PR #73, opened 2016-05-15
Update doc for CP 3.0

Ishiihara:


Ishiihara:
@ewencp @gwenshap Doc changes for CP 3.0

docs/index.rst
```diff
@@ -316,6 +316,20 @@ Configuration Options
   * Default: 60000
   * Importance: low
 
+``timestamp.delay.interval.ms``
+  How long to wait after a row with certain timestamp appears before we include it in the result. You may choose to add some delay to allow transactions with earlier timestamp to complete. The first execution will fetch all available records (i.e. starting at timestamp 0) until current time minus the delay. Every following execution will get data from the last time we fetched until current time minus the delay.
+
+  * Type: long
+  * Default: 0
+  * Importance: low
```
ewencp: Hmm, I don't think this was caught during the review when this option was added -- this was an oversight in the initial implementation and is probably pretty important. Should we raise the importance level here (at least to medium, maybe to high)?

ewencp:
@Ishiihara A couple of thoughts:

1. For other projects, after the first release we've added changelog notes: http://docs.confluent.io/2.0.1/schema-registry/docs/changelog.html They should be straightforward to generate.
2. Should we add a note somewhere at the end of the "Incremental Query Modes" section about `timestamp.delay.interval.ms`?

docs/index.rst
Ishiihara: I think we should raise it to high.

Ishiihara:
@ewencp Made suggested changes. PTAL. Thanks!

docs/jdbc_connector.rst
```diff
@@ -0,0 +1,376 @@
...
+  must track offsets itself.
+
+* **Bulk**: This mode is unfiltered and therefore not incremental at all. It will load all rows
+  from a table on each iteration. This can be useful if you want to periodically dump an entire
+  table where entries are eventually deleted and the downstream system can safely handle duplicates.
+
+Note that all incremental query modes that use certain columns to detect changes will require
+indexes on those columns to efficiently perform the queries.
+
+For incremental query modes that use timestamps, the JDBC connector uses a configuration
```
ewencp: nit: maybe "the wait after" -> "the waiting period after". Same in following sentence

ewencp:
@Ishiihara minor comment on text. For the reorganization, it doesn't seem like it quite follows the pattern of other projects. Should some of the sections like "Configuration Options" and "Schema Evolution" be on separate pages? I'm not really sure here since we don't match all the introductory sections either....

docs/index.rst
```diff
@@ -1,359 +1,10 @@
-.. Kafka Connect JDBC documentation master file
```
ewencp: It will be helpful to have a reference to link to here like many other sections do.

Ishiihara:
@ewencp Addressed the comments. Made some minor reorganization. PTAL. Thanks!

ewencp:
@Ishiihara LGTM, but I think we now want to do this against 3.x and then have it merged back to master. You can just submit a new PR against 3.x and merge immediately.

PR #83, opened 2016-06-07
  Feature/include views in tables

gfodor:
This PR adds support for SQL views. Views seem to be a better way to deal with custom queries since they are encapsulated and can evolve over time without having to push a new connect config.


ewencp:
add to whitelist

src/main/java/io/confluent/connect/jdbc/JdbcSourceTask.java
```diff
@@ -257,6 +257,10 @@ public void stop() throws ConnectException {
   private void validateNonNullable(String incrementalMode, String table, String incrementingColumn,
                                    String timestampColumn) {
     try {
+      // Views typically will not expose NOT NULL constraints on their columns, so don't
+      // check for nullability.
+      if (JdbcUtils.isView(db, table)) return;
```
ewencp: Is that going to be an issue for compatibility as schemas evolve? (This partly depends on the converter you're using and if you want to enforce compatibility, of course.)

src/main/java/io/confluent/connect/jdbc/JdbcUtils.java
```diff
@@ -51,7 +51,7 @@
    * The default only includes standard, user-defined tables.
    */
   public static final Set<String> DEFAULT_TABLE_TYPES = Collections.unmodifiableSet(
-      new HashSet<>(Arrays.asList("TABLE"))
+      new HashSet<>(Arrays.asList("TABLE", "VIEW"))
```
ewencp: This is potentially a compatibility issue since existing users will start seeing views show up in there. We actually already have a similar PR: https://github.com/confluentinc/kafka-connect-jdbc/pull/37 I kind of like the combination of the two patches -- this one has an actual test with Derby instead of just mocking `JdbcUtils` and the `validateNonNullable` tweak, but the other one includes the relevant config, maintains compatibility, and generalizes.

If I get the other one merged, would rebasing against that make sense for you? (The merge is trivial, I just want to make sure the combo would work for your use case.)

gfodor: Yup that's fine. We ended up not needing this patch for now, decided to use a real table instead for the case we had.

ewencp: Great, #37 has been merged. Hopefully the rebase is straightforward.

rnpridgeon:
I've marked this as closed for now with the resolution of #37. Feel free to reopen if the other PR does not satisfy this request

PR #84, opened 2016-06-07
New queries

slowenthal:
Features:

1. Add a new query.limit parameter
2. Refactor the query generation into a query generator interface. 
It allows DB-specific generators
  1. custom: Custom queries - no longer generates predicates for custom queries. Plan to add a feature to describe the parameters.  Internally it works like that.
  2. generic: generate the current queries
  3. postgres: generate queries that support paging predicates (timestamp, incrementing) > (?,?) and a LIMIT clause
  3. mssql: uses the postgres generator


slowenthal:
Note - It's off the 2.0.1 tag, and not the 2.x head right now.

mingfang:
This is PR is necessary for ingesting very large tabes.
How do we move forward with this?

Issue #44, opened 2016-02-16
BigDecimal has mismatching scale value for given Decimal schema


pradnyaalc:
Used the latest kafka-connect-jdbc-2.1.0-SNAPSHOT jar but I am unable to parse the Number format of oracle db. It gives the following exception:

org.apache.kafka.connect.errors.DataException: BigDecimal has mismatching scale value for given Decimal schema
        at org.apache.kafka.connect.data.Decimal.fromLogical(Decimal.java:69)
        at io.confluent.connect.avro.AvroData$5.convert(AvroData.java:218)
        at io.confluent.connect.avro.AvroData.fromConnectData(AvroData.java:318)
        at io.confluent.connect.avro.AvroData.fromConnectData(AvroData.java:450)
        at io.confluent.connect.avro.AvroData.fromConnectData(AvroData.java:267)
        at io.confluent.connect.avro.AvroConverter.fromConnectData(AvroConverter.java:90)
        at org.apache.kafka.connect.runtime.WorkerSourceTask.sendRecords(WorkerSourceTask.java:142)
        at org.apache.kafka.connect.runtime.WorkerSourceTask.access$600(WorkerSourceTask.java:50)
        at org.apache.kafka.connect.runtime.WorkerSourceTask$WorkerSourceTaskThread.execute(WorkerSourceTask.java:356)
        at org.apache.kafka.connect.util.ShutdownableThread.run(ShutdownableThread.java:82)

It seems that the latest version has fixed the issue but even after using the latest jar I am facing the same issue.

Please help..

erhwenkuo:
I am also encountered this error message when I try to use JDBC connector of oracle db. According to [Stackoverflow-Difference between number and integer datatype in oracle dictionary views](http://stackoverflow.com/questions/13494010/difference-between-number-and-integer-datatype-in-oracle-dictionary-views), I found out the "number" type in oracle table caused this issue. In my case, I have a field "ShipAmt" (Type: Number , Scale: 3) defined in oracle, and have one record "ShipAmt=10". From oracle jdbc resultset meta data, the scale (3) is correct but the decimal value using resultset.getBigDecimal() only show scale(0)  which cause value / schema mismatch validation error.
Below is a quick fix: Modify "DataConvert.convertFieldValue" method to enable "scale" check and force setting right scale number on BigDecimal object.

```
public class DataConverter {
private static void convertFieldValue(ResultSet resultSet, int col, int colType,
[...]

PR #89, opened 2016-06-14
Fix to "BigDecimal has mismatching scale value for given Decimal schema"

patelliandrea:
Fix to issue https://github.com/confluentinc/kafka-connect-jdbc/issues/44 as suggested in the comments

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/DataConverter.java
+++ b/src/main/java/io/confluent/connect/jdbc/DataConverter.java
@@ -27,6 +27,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.math.BigDecimal;
 import java.net.URL;
 import java.sql.Blob;
 import java.sql.Clob;
@@ -322,7 +323,12 @@ public class DataConverter {
 
       case Types.NUMERIC:
       case Types.DECIMAL: {
-        colValue = resultSet.getBigDecimal(col);
+        int scale = resultSet.getMetaData().getScale(col);
+        BigDecimal bigDecimalValue = resultSet.getBigDecimal(col);
+        if (bigDecimalValue == null)
+          colValue = null;
+        else
+          colValue = resultSet.getBigDecimal(col).setScale(scale);;
         break;
       }
 


patelliandrea:
[clabot:check

ewencp:
ok to test

src/main/java/io/confluent/connect/jdbc/DataConverter.java
```diff
@@ -322,7 +323,12 @@ private static void convertFieldValue(ResultSet resultSet, int col, int colType,
 
       case Types.NUMERIC:
       case Types.DECIMAL: {
-        colValue = resultSet.getBigDecimal(col);
+        int scale = resultSet.getMetaData().getScale(col);
```
ewencp: Can we test this change in JdbcSourceTaskConversionTest where we test a bunch of other conversions?

src/main/java/io/confluent/connect/jdbc/DataConverter.java
```diff
@@ -322,7 +323,12 @@ private static void convertFieldValue(ResultSet resultSet, int col, int colType,
...
+        BigDecimal bigDecimalValue = resultSet.getBigDecimal(col);
+        if (bigDecimalValue == null)
+          colValue = null;
+        else
+          colValue = resultSet.getBigDecimal(col).setScale(scale);;
```
ewencp: Double `;` is causing a checkstyle failure.

patelliandrea:
Hi, I've removed the double ;
The function passed all the tests in the JdbcSourceTaskConversionTest class

patelliandrea:
@ConfluentCLABot [clabot:check]

ewencp:
@patelliandrea I was curious about additional tests that actually trigger the need for this. I know we validate the scale in the logical type conversion which is why this is needed, but I'm not sure when it is occurring. `setScale` can throw an `ArithmeticException` if the scale is being reduced and the conversion would require rounding, and that isn't handled here. I would assume no database should be returning data with higher precision than requested, but since I don't know why/how this is happening in the first place I think it's probably better not to make an assumption like that.

patelliandrea:
@ewencp I've actually encountered the ArithmeticException because the conversion needed rounding so right now I'm using setScale with an hardcoded rounding mode, I'm working in order to make the rounding mode configurable.

PR #93, opened 2016-06-20
Fix thread-safety of date/time conversions in DataConverter

### LINKED ISSUE #92: convertRecord method in the DataConverter class is not thread-safe
The convertRecord method in the DataCoverter class is not thread-safe. A shared mutable Calendar instance is being used in the method to convert to the UTC time zone. So, if there are multiple tasks being used in the JDBC connector, this causes a race condition among the tasks trying to fetch data from timestamp or date type columns of different tables resulting in an incorrect column value being returned as a result.

shikhar:
Closes #92


ewencp:
LGTM

Ishiihara:
LGTM

PR #96, opened 2016-06-28
CC-243: use Long for nanos in offset map rather than Integer

shikhar:
if JSON converter is in use we can get a Long back even if an Integer is used at storage time


ewencp:
@shikhar LGTM.

I wonder if we should add some integration testing utilities to core Kafka to better support exercising some of this. I don't want to get to requiring the full Kafka cluster setup, but maybe something that can test approximately at the WorkerSourceTask/WorkerSinkTask level and mock out the producer/consumer stuff. Basically, enough that we can exercise data being converted, offsets being serialized, etc in a mostly realistic way.

shikhar:
@ewencp good point, going to think over that! system tests are nice as they cast a wide net but the issue here seems like something we should have been able to catch separately.

PR #100, opened 2016-07-26
CC-227: merge in JDBC Sink 

shikhar:
There are 2 commits to review here:

1. Preparatory -- move classes from source into a subpkg (except the Connector):
https://github.com/confluentinc/kafka-connect-jdbc/pull/100/commits/d0d47923c4c4712dfa20377dd55af61b81f5bb8c

2. The merge -- https://github.com/confluentinc/kafka-connect-jdbc/pull/100/commits/2c6cb8ced93aa874b41419a0d55cf71423529233 -- which introduces all the history. The merge-related changes worth reviewing are in: `checkstyle.xml` & `pom.xml`. Rest is purely code addition.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,9 @@
         <powermock.version>1.6.2</powermock.version>
         <derby.version>10.11.1.1</derby.version>
         <commons-io.version>2.4</commons-io.version>
+        <sqlite.version>3.8.11.2</sqlite.version>
+        <mockito.version>1.10.19</mockito.version>
+        <licenses.version>3.1.0-SNAPSHOT</licenses.version>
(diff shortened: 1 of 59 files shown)


pom.xml
```diff
@@ -91,6 +94,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.xerial</groupId>
```
ewencp: This is duplicated between `runtime` and `test` which causes maven to complain with a warning when you run it. The `runtime` one should cover the `test` classpath as well, so I think you just want to remove this second copy.

ewencp:
@shikhar One minor issue with the pom, but clean that up and it LGTM

shikhar:
Thanks @ewencp!

PR #102, opened 2016-07-27
JDBC Sink ConfigDef - use 0.10 fields; doc updates

shikhar:
Adding group, width, display string and also made some documentation updates along the way


ewencp:
@Ishiihara you actually know this better than I do, mind taking a pass?

shikhar:
Going to merge this into a docs PR as the config docs have undergone changes in that branch.

PR #104, opened 2016-07-29
fixes #101: More advanced NUMERIC data conversion

### LINKED ISSUE #101: More advanced NUMERIC data conversion
Hi,
I was trying to use kafka-connect-jdbc to feed from an Oracle DB.
Here are a few blocker issues I was able to identify:
1. Metadata does not always contain proper information
   For `NUMERIC` columns that are the result of aggregation it will say that the scale is 0 even though the result is a floating point - which will result in an Exception when trying to create a `BigDecimal` with scale 0 and some value with fraction. This happens because the default rounding policy for `BigDecimal` is `ROUND_UNNECESSARY` that throws the exception. There's already an issue raised for that: #44.
   Users should be advised to use a `CAST` function or an alternative to tackle such problems.
2. Oracle does not use the following datatypes ([details](http://docs.oracle.com/cd/B19306_01/gateways.102/b14270/apa.htm)): `BIT`, `TINYINY`, `SMALLINT`, `INTEGER`, `BIGINT`
   Instead it represents them with a `NUMBER(precision,scale)` which according to the current `DataConverter` implementation maps to `NUMERIC` that is handled by the `DECIMAL` conversion resulting in `BigDecimal` values.
   Using `BigDecimal` is an overkill to store the values are known to be in the range of the datatypes listed above.
   I suggest either providing a way of specifying a custom type mapper or changing the default one to be like this:

``` java
...
case Types.NUMERIC:
        int precision = metadata.getPrecision(col);
        if (metadata.getScale(col) == 0 && precision < 20) { // integer
          Schema schema;
          if (precision > 10) {
            schema = (optional) ? Schema.OPTIONAL_INT64_SCHEMA : Schema.INT64_SCHEMA;
          } else if (precision > 5) {
            schema = (optional) ? Schema.OPTIONAL_INT32_SCHEMA : Schema.INT32_SCHEMA;
          } else if (precision > 3) {
            schema = (optional) ? Schema.OPTIONAL_INT16_SCHEMA : Schema.INT16_SCHEMA;
          } else {
            schema = (optional) ? Schema.OPTIONAL_INT8_SCHEMA : Schema.INT8_SCHEMA;
          }
          builder.field(fieldName, schema);
          break;
        }
      case Types.DECIMAL: {
        SchemaBuilder fieldBuilder = Decimal.builder(metadata.getScale(col));
        if (optional) {
          fieldBuilder.optional();
        }
        builder.field(fieldName, fieldBuilder.build());
        break;
      }
...
```

This will also require the following modification to the value conversion:

``` java
...
case Types.NUMERIC:
        ResultSetMetaData metadata = resultSet.getMetaData();
        int precision = metadata.getPrecision(col);
        if (metadata.getScale(col) == 0 && precision < 20) { // integer
          if (precision > 10) {
            colValue = resultSet.getLong(col);
          } else if (precision > 5) {
            colValue = resultSet.getInt(col);
          } else if (precision > 3) {
            colValue = resultSet.getShort(col);
          } else {
            colValue = resultSet.getByte(col);
          }
          break;
        }
      case Types.DECIMAL: {
        BigDecimal bigDecimalValue = resultSet.getBigDecimal(col);
        if (bigDecimalValue != null) {
          colValue = bigDecimalValue.setScale(resultSet.getMetaData().getScale(col));
        } else {
          colValue = null;
        }
        break;
      }
...
```

I can provide a pull-request if you agree with the proposed change.

clumsy:
numeric.precision.mapping options is introduce to provide precision-based mapping for numeric columns

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
@@ -36,8 +36,8 @@ import io.confluent.connect.jdbc.util.JdbcUtils;
 public class BulkTableQuerier extends TableQuerier {
   private static final Logger log = LoggerFactory.getLogger(BulkTableQuerier.class);
 
-  public BulkTableQuerier(QueryMode mode, String name, String topicPrefix) {
-    super(mode, name, topicPrefix);
+  public BulkTableQuerier(QueryMode mode, String name, JdbcSourceTaskConfig config) {
+    super(mode, name, config);
   }
 
   @Override
@@ -63,7 +63,7 @@ public class BulkTableQuerier extends TableQuerier {
 
   @Override
   public SourceRecord extractRecord() throws SQLException {
-    Struct record = DataConverter.convertRecord(schema, resultSet);
+    Struct record = converter.convertRecord(schema, resultSet);
     // TODO: key from primary key? partition?
     final String topic;
     final Map<String, String> partition;
--- a/src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
@@ -45,7 +45,7 @@ import java.util.TimeZone;
  * Connect records.
  */
 public class DataConverter {
-  private static final Logger log = LoggerFactory.getLogger(JdbcSourceTask.class);
+  private static final Logger log = LoggerFactory.getLogger(DataConverter.class);
 
   private static final ThreadLocal<Calendar> UTC_CALENDAR = new ThreadLocal<Calendar>() {
     @Override
@@ -54,7 +54,13 @@ public class DataConverter {
     }
   };
 
-  public static Schema convertSchema(String tableName, ResultSetMetaData metadata)
+  private final boolean numericPrecisionMapping;
+
+  DataConverter(JdbcSourceTaskConfig config) {
+    this.numericPrecisionMapping = config.getBoolean(JdbcSourceConnectorConfig.NUMERIC_PRECISION_MAPPING_CONFIG);
+  }
+
+  public Schema convertSchema(String tableName, ResultSetMetaData metadata)
       throws SQLException {
     // TODO: Detect changes to metadata, which will require schema updates
     SchemaBuilder builder = SchemaBuilder.struct().name(tableName);
@@ -64,7 +70,7 @@ public class DataConverter {
     return builder.build();
   }
 
-  public static Struct convertRecord(Schema schema, ResultSet resultSet)
+  public Struct convertRecord(Schema schema, ResultSet resultSet)
       throws SQLException {
     ResultSetMetaData metadata = resultSet.getMetaData();
     Struct struct = new Struct(schema);
@@ -82,7 +88,7 @@ public class DataConverter {
   }
 
 
-  private static void addFieldSchema(ResultSetMetaData metadata, int col,
+  private void addFieldSchema(ResultSetMetaData metadata, int col,
                                      SchemaBuilder builder)
       throws SQLException {
     // Label is what the query requested the column name be using an "AS" clause, name is the
@@ -175,6 +181,23 @@ public class DataConverter {
       }
 
       case Types.NUMERIC:
+        if (numericPrecisionMapping) {
+          int precision = metadata.getPrecision(col);
+          if (metadata.getScale(col) == 0 && precision < 20) { // integer
+            Schema schema;
+            if (precision > 10) {
+              schema = (optional) ? Schema.OPTIONAL_INT64_SCHEMA : Schema.INT64_SCHEMA;
+            } else if (precision > 5) {
+              schema = (optional) ? Schema.OPTIONAL_INT32_SCHEMA : Schema.INT32_SCHEMA;
+            } else if (precision > 3) {
+              schema = (optional) ? Schema.OPTIONAL_INT16_SCHEMA : Schema.INT16_SCHEMA;
+            } else {
+              schema = (optional) ? Schema.OPTIONAL_INT8_SCHEMA : Schema.INT8_SCHEMA;
+            }
+            builder.field(fieldName, schema);
+            break;
+          }
+        }
       case Types.DECIMAL: {
         SchemaBuilder fieldBuilder = Decimal.builder(metadata.getScale(col));
         if (optional) {
@@ -262,7 +285,7 @@ public class DataConverter {
     }
   }
 
-  private static void convertFieldValue(ResultSet resultSet, int col, int colType,
+  private void convertFieldValue(ResultSet resultSet, int col, int colType,
                                         Struct struct, String fieldName)
       throws SQLException, IOException {
     final Object colValue;
@@ -326,6 +349,22 @@ public class DataConverter {
       }
 
       case Types.NUMERIC:
+        if (numericPrecisionMapping) {
+          ResultSetMetaData metadata = resultSet.getMetaData();
+          int precision = metadata.getPrecision(col);
+          if (metadata.getScale(col) == 0 && precision < 20) { // integer
+            if (precision > 10) {
+              colValue = resultSet.getLong(col);
+            } else if (precision > 5) {
+              colValue = resultSet.getInt(col);
+            } else if (precision > 3) {
+              colValue = resultSet.getShort(col);
+            } else {
+              colValue = resultSet.getByte(col);
+            }
+            break;
+          }
+        }
       case Types.DECIMAL: {
         colValue = resultSet.getBigDecimal(col);
         break;
@@ -432,5 +471,4 @@ public class DataConverter {
     // be faster than setting this by name?
     struct.put(fieldName, resultSet.wasNull() ? null : colValue);
   }
-
 }
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -53,6 +53,12 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
   public static final int BATCH_MAX_ROWS_DEFAULT = 100;
   private static final String BATCH_MAX_ROWS_DISPLAY = "Max Rows Per Batch";
 
+  public static final String NUMERIC_PRECISION_MAPPING_CONFIG = "numeric.precision.mapping";
+  private static final String NUMERIC_PRECISION_MAPPING_DOC =
+      "Whether or not to map NUMERIC values by precision";
+  public static final boolean NUMERIC_PRECISION_MAPPING_DEFAULT = false;
+  private static final String NUMERIC_PRECISION_MAPPING_DISPLAY = "Map Numeric Values By Precision";
+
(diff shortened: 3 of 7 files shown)


clumsy:
There, CLA signed.

ewencp:
ok to test

ewencp:
@clumsy Awesome, you are super responsive :) Re: NUMERIC(1,0) and Boolean, do you have an idea of which databases actually do this? It seems like it should be an unusual case. I'd figure that most DBs would treat boolean separately even if it can be reduced to a number. I'd be inclined to omit translation unless we know of a DB that ends up representing boolean as `NUMERIC(1,0)` via the JDBC driver.

clumsy:
I'm 100% sure Oracle has no built-in `BOOLEAN` type, everybody uses `NUMBER(1,0)` instead.
Plus I've seen an already merged commit cede49adecf2c003348c6e5ee962b42aaa38c0ce that has the following line for Oracle dialect in Sink:

```java
...
map.put(Schema.Type.BOOLEAN, "NUMBER(1,0)");
...
```

Not that one can do any other useful action with a `NUMBER(1,0)`...
I can add it, just let me know what you guys think.

shikhar:
@clumsy I think we should avoid that because storing a boolean in `NUMBER(1,0)` cannot loose any information, but the other way can.

On the other hand,
```
upto 2 digits i.e. [-99,99] -> byte i.e. [-128,128]
upto 4 digits i.e. [-9999,9999] -> short i.e. [-32768,32768]
upto 9 digits i.e. [-999999999,999999999] -> int i.e. [-2147483648,2147483648]
upto 19 digits i.e. [-9999999999999999999,9999999999999999999] -> long i.e. [-9223372036854775808,9223372036854775808]
```
Er I was going to write that these are all safe, but actually the last one isn't :-( So this was a good exercise... perhaps we should restrict it to 18-digit precision?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -53,6 +53,12 @@
   public static final int BATCH_MAX_ROWS_DEFAULT = 100;
   private static final String BATCH_MAX_ROWS_DISPLAY = "Max Rows Per Batch";
 
+  public static final String NUMERIC_PRECISION_MAPPING_CONFIG = "numeric.precision.mapping";
+  private static final String NUMERIC_PRECISION_MAPPING_DOC =
+      "Whether or not to map NUMERIC values by precision";
```
shikhar: "Whether or not to attempt mapping NUMERIC values by precision to integral types"

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
```diff
@@ -36,8 +36,8 @@
 public class BulkTableQuerier extends TableQuerier {
   private static final Logger log = LoggerFactory.getLogger(BulkTableQuerier.class);
 
-  public BulkTableQuerier(QueryMode mode, String name, String topicPrefix) {
-    super(mode, name, topicPrefix);
+  public BulkTableQuerier(QueryMode mode, String name, JdbcSourceTaskConfig config) {
```
shikhar: I think it would be cleaner to just add a boolean parameter for the new config property... the other parameters that remain are also derived from the config, so that feels more consistent.

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -175,6 +181,23 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
       }
 
       case Types.NUMERIC:
+        if (numericPrecisionMapping) {
+          int precision = metadata.getPrecision(col);
+          if (metadata.getScale(col) == 0 && precision < 20) { // integer
```
shikhar: To make this conversion totally safe, this should be `precision < 19` since INT64 cannot represent the full range of 19 digit precision integers.

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -54,7 +54,13 @@ protected Calendar initialValue() {
     }
   };
 
-  public static Schema convertSchema(String tableName, ResultSetMetaData metadata)
+  private final boolean numericPrecisionMapping;
+
+  DataConverter(JdbcSourceTaskConfig config) {
+    this.numericPrecisionMapping = config.getBoolean(JdbcSourceConnectorConfig.NUMERIC_PRECISION_MAPPING_CONFIG);
```
shikhar: I'd be in favor of leaving this as a class with purely static methods, and making the `numericPrecisionMapping` a boolean argument for the methods that depend on it.

shikhar:
@clumsy other than the comments, looking good! 👍 for tests

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
clumsy: I had this dilemma too, and spent a few minutes thinking how to make it both elegant and static.
But then I checked that the class is only needed from at max one object - that is only one instance is ever going to be created I decided to go with the instance field.
Later on it could be generified so that it's more like a pluggable strategy.

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
clumsy: Sorry, I do not follow this comment.
int8 (aka byte) = -128..127 that is 3 digits. How can we put it into NUMBER(2,0)?
similarly int16 (aka int) = -32768.. 32767 - 5 digits
...
int64 (aka long) = -9223372036854775808..9223372036854775807 - 19 digits

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
clumsy: 4 constructor parameters is too much for my liking. It's not that it's a big deal, but unlike others it's the original value that we use.
As for other parameters - they are not vanilla values, they are derived.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
clumsy: Fixed

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
shikhar: To take int8 as an example, for storing int8 `NUMBER(2,0)` does not work as you said since there are 3 digits of precision for `[-128..127]`, so if you were to create a db column for storing int8 `NUMBER(3,0)` makes sense.

But if you are getting a `NUMBER(3,0)` value back from the db, in theory the value could be anywhere in `[-999,999]`, so it would be unsafe to treat that as an int8.

clumsy: If one has `NUMBER(3,0)` which is our of the `int8` range - one will have to use a bigger precision.
It will be done explicitly most of the time.

Also in a situation when there's no metadata for aggregation columns, plus there are cases when one is summing up `NUMBER(3,0)`s it could theoretically result in a number outside of the precision of the components.

Thus in any case there's a work-around - manual cast with elected precision.
Seems like a better alternative than to loose half of the range available of values.

ewencp: @clumsy Not sure I understand your response. Isn't the whole point here that you want the `NUMBER` returned for all the `INT` types just work? The mappings you're giving will silently clamp some values that are in the valid range.

I think you might be suggesting that since you're seeing `NUMBER(3,0)` for `TINYINT`, you want to make sure that get's mapped back to `INT8` since that preserves the original width. But I see two problems with that. First, it means that you can lose information when the column actually *is* `NUMBER(3,0)`. Second, I'm pretty sure this isn't even accurate. Some quick googling seems to indicate that in Oracle, the various `INT` types vary between signed and unsigned, so you'd still need to . At least for `TINYINT`, it is actually an unsigned value 0-255, so mapping to `INT8` would actually lose data. It seems like it would be better to map to a type that is guaranteed to capture the full range of values. Lack of signed/unsigned is a tradeoff we made in choosing what to include in the data API based on what different serialization formats support, but it does sometimes mean promoting to a wider type to fully capture the set of possible values.

Unfortunately, I think there are just going to be limits to what we can support if Oracle doesn't export sufficient information to tell between "fake" `NUMBER` columns that are actually `INT` types vs "real" `NUMBER` columns that don't have a constrained set of values.

clumsy: Regarding the `TINYINT` comment.
Possible 0-255 values do not seem to stop the current implementation from handling explicit `TINYINT` metadata mapping it to Java byte type (-128..127):

```java
...
// ints <= 8 bits
      case Types.BIT:
      case Types.TINYINT: {
        if (optional) {
          builder.field(fieldName, Schema.OPTIONAL_INT8_SCHEMA);
        } else {
          builder.field(fieldName, Schema.INT8_SCHEMA);
        }
        break;
      }
...
// 8 bits int
      case Types.TINYINT: {
        colValue = resultSet.getByte(col);
        break;
      }
...
```

There are not that many options that we can get the column information from in the query result.
I guess we are limited to the metadata or the column name.
We can suffix the column name with the type - and to truncate it later in the Connect schema.
Or (and I think better) - use metadata.
In case the value cannot be represented in Connect for the given precision - just specify a bigger one.
E.g. if you have `NUMBER(3,0)` that are within Connect (and Java) range for byte (int8) - either use the column as is or cast to `NUMBER(3,0)`.
If your value is `NUMBER(3,0)` but is outside the Connect (and Java) range for byte (int8) - use a data type with a bigger precision, e.g. `NUMBER(5,0)` - short (int16).

In other words - using `NUMBER(3,0)` will be a way of telling kaka-connect-jdbc to map the column to `int8`.

shikhar: > Possible 0-255 values do not seem to stop the current implementation from handling explicit TINYINT metadata mapping it to Java byte type (-128..127):

That logic looks safe to me. `resultSet.getByte()` is JDBC driver-specific behavior for treating the column value as a byte, but the expectation is that it does the right thing and returns a signed byte. For example, if you try with SQLite JDBC a select query for "SELECT 255" and `getByte()` on the result correctly gives me `-1`.

shikhar: > In other words - using NUMBER(3,0) will be a way of telling kaka-connect-jdbc to map the column to int8.

I think we should err on the side of the wider type for now rather than add potentially unsafe conversions, and use the mappings I suggested in https://github.com/confluentinc/kafka-connect-jdbc/pull/104/files#r72742317.

One option if you need a specific type that the automatic conversions can't address is to implement a custom `Converter` that wraps one of the existing ones and do the down-casting in there. You'd want to hook into the `fromConnectData` since this is a source connector and that method will be called along the way from Connect -> Kafka. [ In future this may become more flexible as we design and integrate 'single-message transforms', e.g. there could be a transformation that allows you to project to a specific Avro schema you define even if it involves down-casting. ]

clumsy:
This logic is to dodgy. There should be a configurable alternative.

PR #106, opened 2016-08-01
CC-231: Add JDBC Sink Connector documentation

shikhar:
Restructue the docs to separate pages for the source and the sink

Also adding some 0.10.0 fields to the ConfigDef

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -8,6 +8,8 @@ Contents:
 .. toctree::
    :maxdepth: 3
 
-   jdbc_connector
-   configuration_options
+   source_connector
+   source_config_options
+   sink_connector
+   sink_config_options
    changelog
--- a/docs/sink_config_options.rst
+++ b/docs/sink_config_options.rst
@@ -1,0 +1,146 @@
+.. _sink-config-options:
+
+JDBC Sink Configuration Options
+-------------------------------
+
+Connection
+^^^^^^^^^^
+
+``connection.url``
+  JDBC connection URL.
+
+  * Type: string
+  * Default: ""
+  * Importance: high
+
+``connection.user``
+  JDBC connection user.
+
+  * Type: string
+  * Importance: high
+
+``connection.password``
+  JDBC connection password.
+
+  * Type: password
+  * Importance: high
+
+Writes
+^^^^^^
+
+``table.name.format``
+  A format string for the destination table name, which may contain '${topic}' as a placeholder for the originating topic name.
+
+  For example, ``kafka_${topic}`` for the topic 'orders' will map to the table name 'kafka_orders'.
+
+  * Type: string
+  * Default: "${topic}"
+  * Importance: medium
+
+``batch.size``
+  Specifies how many records to attempt to batch together for insertion, when possible.
+
+  * Type: int
+  * Default: 3000
+  * Importance: high
+
+
+``insert.mode``
+  The insertion mode to use. Supported modes are:
+
+  `insert`
+
+      Use standard SQL ``INSERT`` statements.
+
+  `upsert`
+
+      Use the appropriate upsert semantics for the target database if it is supported by the connector, e.g. ``INSERT OR IGNORE``.
+
+  * Type: string
+  * Default: "insert"
+  * Importance: medium
+
+.. _sink-pk-config-options:
+
+Primary Keys
+^^^^^^^^^^^^
+
+``pk.mode``
+  The primary key mode, also refer to ``pk.fields`` documentation for interplay. Supported modes are:
+
+  `none`
+
+      No keys utilized.
+
+  `kafka`
+
+      Kafka coordinates are used as the PK.
+
+  `record_key`
+
+      Field(s) from the record key are used, which may be a primitive or a struct.
+
+  `record_value`
+
+      Field(s) from the record value are used, which must be a struct.
+
+  * Type: string
+  * Default: "none"
+  * Importance: high
+
+``pk.fields``
+  List of comma-separated primary key field names. The runtime interpretation of this config depends on the ``pk.mode``:
+
+  `none`
+
+      Ignored as no fields are used as primary key in this mode.
+
+  `kafka`
+
+      Must be a trio representing the Kafka coordinates, defaults to ``__connect_topic,__connect_partition,__connect_offset`` if empty.
+
+  `record_key`
+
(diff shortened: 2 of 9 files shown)


shikhar:
@ewencp @Ishiihara ptal

docs/index.rst
```diff
@@ -8,6 +8,8 @@ Contents:
 .. toctree::
    :maxdepth: 3
 
-   jdbc_connector
-   configuration_options
+   source_connector
```
ewencp: Does adding hierarchy here make sense (i.e. Source connector -> (description, configs))? Or is that just not warranted for what we have so far? Just curious because the interleaved ordering seems a bit odd.

docs/sink_config_options.rst
```diff
@@ -0,0 +1,150 @@
...
+
+      Use the appropriate upsert semantics for the target database if it is supported by the connector, e.g. ``INSERT OR IGNORE``.
+
+  This config is overridable at the table-level by using a ``$table.`` prefix.
+
+  * Type: string
+  * Default: "insert"
+  * Importance: medium
+
+.. _sink-pk-config-options:
```
ewencp: Hmm, we may need to figure out how to get these references auto-generated. Otherwise we run the risk of accidentally losing them the next time someone updates this file.

docs/sink_connector.rst
```diff
@@ -0,0 +1,139 @@
+JDBC Sink Connector
+===================
+
+The JDBC sink connector allows you to export data from Kafka topics to any relational database with a JDBC driver.
+By using JDBC, this connector can support a wide variety of databases without requiring custom code for each one.
```
ewencp: Heh, might want to refine the end of this sentence as we technically have custom code for a bunch of different DBs. More like each database doesn't need its own dedicated connector implementation.

docs/sink_connector.rst
```diff
@@ -0,0 +1,139 @@
...
+The connector polls data from Kafka to write to the database, based on the topics subscription.
+It is possible to achieve idempotent writes with upserts.
+Auto-creation of tables, and limited auto-evolution is also supported.
+
+Quickstart
+----------
+
+To see the basic functionality of the connector, we'll be copying from a single topic to a local SQLite database.
```
ewencp: Maybe mention Avro data here to setup the Schema Registry requirement in the next sentence.

docs/sink_connector.rst
```diff
@@ -0,0 +1,139 @@
...
+This examples assume you are running Kafka and Schema Registry locally on the default ports.
```
ewencp: examples -> example

docs/sink_connector.rst
```diff
@@ -0,0 +1,139 @@
...
+
+Now if we query the database, we will see that the `orders` table was automatically created and contains the record.
+
+.. sourcecode:: bash
+
+    $ sqlite3 test.db
+    sqlite> select * from orders;
+    foo|50.0|100|999
+
+Features and Limitations
```
ewencp: The section title is accurate, but let's just leave it at "Features" to sound a bit more positive :)

docs/sink_connector.rst
```diff
@@ -0,0 +1,139 @@
...
+------------------------
+
+Data mapping
+^^^^^^^^^^^^
+
+The sink connector requires knowledge of record schemas, so you should use a suitable converter e.g. the Avro converter that comes with the schema registry,
+or the JSON converter with schemas enabled.
+Record keys if present can be primitive types or a struct, and the record value must be a struct.
```
ewencp: Hmm, some of this will be confusing to Connect newbies since they may not know the difference between Record vs struct. Maybe use "Kafka record" to clarify?

docs/sink_connector.rst
```diff
@@ -0,0 +1,139 @@
...
+Fields being selected from structs must be primitive.
+If the data in the topic is not of a compatible format, implementing a custom ``Converter`` may be necessary.
+
+Key handling
+^^^^^^^^^^^^
+
+The default is for primary keys to not be extracted with ``pk.mode`` set to "none",
```
ewencp: quoted values like `"none"` here should probably be in backticks.

docs/sink_connector.rst
```diff
@@ -0,0 +1,139 @@
...
+
+.. tip:: Make sure the JDBC user has the appropriate permissions for DDL.
+
+If ``auto.create`` is enabled, the connector can ``CREATE`` the destination table if it is found to be missing.
+The creation takes place online with records being consumed from the topic, since the connector uses the record schema as a basis for the table definition.
+Primary keys are specified based on the key configuration settings.
+
+If ``auto.evolve`` is enabled, the connector can perform limited auto-evolution by issuing ``ALTER`` on the destination table when it encounters a record for which a column is found to be missing.
+Since data-type changes and removal of columns can be dangerous, the connector does not attempt to perform such evolutions on the table.
+Addition of primary keys constraints is also not attempted.
```
ewencp: primary keys constraints -> primary key constraints

docs/sink_config_options.rst
shikhar: I actually had to create the sections in the config options manually too. It would be good to take into account the 0.10.0 group for each option in `toRst()` for dividing the page into sections. I'll ticket up these improvements.

docs/sink_connector.rst
shikhar: "Kafka record keys if present can be primitive types or a Connect struct, and the record value must be a Connect struct." seem ok?

docs/sink_config_options.rst
shikhar: https://issues.apache.org/jira/browse/KAFKA-4010

docs/index.rst
shikhar: Hmm, no strong preference from me but the current number of pages doesn't warrant adding an additional level of hierarchy I think. I'll change the ordering though so it's [source_connector, source_config_options, sink_connector, sink_config_options

docs/sink_connector.rst
ewencp: @shikhar Yeah, sounds good.

docs/index.rst
ewencp: Yeah, maybe why it looks a bit weird is that we get a level of hierarchy within the first page such that when you generate the docs you get a `+` symbol for that with a bunch of subsections, and then the configs are separate.

ewencp:
@shikhar LGTM. We'll just have to take care wrt lack of KAFKA-4010 until someone implements it.

PR #107, opened 2016-08-01
CC-266: use default values from Schema in DDL

shikhar:


shikhar:
@confluentinc/connect ptal

src/test/java/io/confluent/connect/jdbc/sink/dialect/DbDialectTest.java
```diff
@@ -16,9 +16,13 @@
 
 package io.confluent.connect.jdbc.sink.dialect;
 
+import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.junit.Test;
 
+import java.nio.charset.Charset;
```
ewencp: unused imports

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
```diff
@@ -107,13 +110,50 @@ protected void writeColumnSpec(StringBuilder builder, SinkRecordField f) {
...
+      formatColumnValue(builder, f.type, f.defaultValue);
+    } else if (f.isOptional) {
       builder.append(" NULL");
     } else {
       builder.append(" NOT NULL");
     }
   }
 
+  static void formatColumnValue(StringBuilder builder, Schema.Type type, Object value) {
+    switch (type) {
```
ewencp: Should we be checking for any of the logical types?

ewencp:
@shikhar One nit, one question about supporting logical types. Otherwise LGTM

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
shikhar: Good point. We don't support logical types elsewhere yet (in DDL `DbDialect.getSqlType()`, for insertion `PreparedStatementBinder.bindField`), so not worth starting here.

We should, though! Not sure if it's important for the initial release. I'll ticket it.

shikhar: https://confluentinc.atlassian.net/browse/CC-278

PR #108, opened 2016-08-11
CC-278: Sink: support Connect schema logical types

YAtOff:
Currently the sink doesn't support the logical types:

* `org.apache.kafka.connect.data.Date`
* `org.apache.kafka.connect.data.Time`
* `org.apache.kafka.connect.data.Timestamp`
* `org.apache.kafka.connect.data.Decimal`

The fields with such types have primitive schema type, but also a name.
The logical type is determined by the name.

When the converter see a logical type it converts the value to some higher level type:
 * `Date`: from `int` to `java.util.Date`
 * `Time`: from `int` to `java.util.Date`
 * `Timestamp`: from `long` to `java.util.Date`
 * `Decimal`: from `bytes[]` to `BigDecimal`

The problem is that the statement builder doesn't know that. For example it thinks that `Timestamp` is `long`, but it is `java.util.Date`.

I don't write much Java and maybe the code is not very good, but it fixes the problem.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
@@ -161,11 +161,11 @@ public class PreparedStatementBinder {
     } else {
       if (schema.name() != null) {
         LogicalTypeConverter logicalConverter = LOGICAL_CONVERTERS.get(schema.name());
-        if (logicalConverter != null)
+        if (logicalConverter != null) {
           value = logicalConverter.convert(schema, value);
+        }
       }
-      Schema.Type type = schema.type();
-      switch (type) {
+      switch (schema.type()) {
         case INT8:
           statement.setByte(index, (Byte) value);
           break;
@@ -202,7 +202,7 @@ public class PreparedStatementBinder {
           statement.setBytes(index, bytes);
           break;
         default:
-          throw new ConnectException("Unsupported source data type: " + type);
+          throw new ConnectException("Unsupported source data type: " + schema.type());
       }
     }
   }
--- a/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
@@ -16,20 +16,72 @@
 
 package io.confluent.connect.jdbc.sink;
 
+import org.apache.kafka.connect.data.Decimal;
+import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.data.Time;
+import org.apache.kafka.connect.data.Timestamp;
 import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.sink.SinkRecord;
 
+import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
+import java.util.HashMap;
 
 import io.confluent.connect.jdbc.sink.metadata.FieldsMetadata;
 import io.confluent.connect.jdbc.sink.metadata.SchemaPair;
 
 public class PreparedStatementBinder {
+
+  // Convert values in Connect form from their logical types. These logical converters are discovered by logical type
+  // names specified in the field
+  private static final HashMap<String, LogicalTypeConverter> LOGICAL_CONVERTERS = new HashMap<>();
+  static {
+    LOGICAL_CONVERTERS.put(Decimal.LOGICAL_NAME, new LogicalTypeConverter() {
+      @Override
+      public Object convert(Schema schema, Object value) {
+        if (!(value instanceof BigDecimal)) {
+          throw new DataException("Invalid type for Decimal, underlying representation should be BigDecimal but was " + value.getClass());
+        }
+        return Decimal.fromLogical(schema, (BigDecimal) value);
+      }
+    });
+
+    LOGICAL_CONVERTERS.put(Date.LOGICAL_NAME, new LogicalTypeConverter() {
+      @Override
+      public Object convert(Schema schema, Object value) {
+        if (!(value instanceof java.util.Date)) {
+          throw new DataException("Invalid type for Date, underlying representation should be java.util.Date but was " + value.getClass());
+        }
+        return Date.fromLogical(schema, (java.util.Date) value);
+      }
+    });
+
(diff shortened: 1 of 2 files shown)


YAtOff:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -85,28 +137,34 @@ public void bindRecord(SinkRecord record) throws SQLException {
...
   }
 
-  static void bindField(PreparedStatement statement, int index, Schema.Type type, Object value) throws SQLException {
+  static void bindField(PreparedStatement statement, int index, Schema schema, Object value) throws SQLException {
     if (value == null) {
       statement.setObject(index, null);
     } else {
+      if (schema.name() != null) {
+        LogicalTypeConverter logicalConverter = LOGICAL_CONVERTERS.get(schema.name());
+        if (logicalConverter != null)
```
shikhar: very minor nit, but the rest of the sink code is using braces even around single statement blocks

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -85,28 +137,34 @@ public void bindRecord(SinkRecord record) throws SQLException {
...
+          value = logicalConverter.convert(schema, value);
+      }
+      Schema.Type type = schema.type();
       switch (type) {
```
shikhar: another minor style nit, perhaps inline the value of type here

src/test/java/io/confluent/connect/jdbc/sink/PreparedStatementBinderTest.java
```diff
@@ -66,7 +77,18 @@ public void bindRecord() throws SQLException {
...
         .put("double", -2436546.56457)
         .put("bytes", new byte[]{-32, 124})
-        .put("age", 30);
+        .put("age", 30)
+        .put("decimal", new BigDecimal("1.5").setScale(0, BigDecimal.ROUND_HALF_EVEN));
+
+    Calendar calendar = Calendar.getInstance();
+    calendar.setTimeInMillis(0);
+    calendar.set(2000, 1, 1);
+    valueStruct.put("date", calendar.getTime());
```
shikhar: It should be cleaner to just instantiate `Date` directly here and a few lines below, e.g. `new Date(long)` and not mess with the `Calendar` instance.

src/test/java/io/confluent/connect/jdbc/sink/PreparedStatementBinderTest.java
```diff
@@ -108,57 +130,59 @@ public void bindRecord() throws SQLException {
...
+    verifyBindField(++index, Schema.INT16_SCHEMA, (short) 42).setShort(index, (short) 42);
+    verifyBindField(++index, Schema.INT32_SCHEMA, 42).setInt(index, 42);
+    verifyBindField(++index, Schema.INT64_SCHEMA, 42L).setLong(index, 42L);
+    verifyBindField(++index, Schema.BOOLEAN_SCHEMA, false).setBoolean(index, false);
+    verifyBindField(++index, Schema.BOOLEAN_SCHEMA, true).setBoolean(index, true);
+    verifyBindField(++index, Schema.FLOAT32_SCHEMA, -42f).setFloat(index, -42f);
+    verifyBindField(++index, Schema.FLOAT64_SCHEMA, 42d).setDouble(index, 42d);
+    verifyBindField(++index, Schema.BYTES_SCHEMA, new byte[]{42}).setBytes(index, new byte[]{42});
+    verifyBindField(++index, Schema.BYTES_SCHEMA, ByteBuffer.wrap(new byte[]{42})).setBytes(index, new byte[]{42});
+    verifyBindField(++index, Schema.STRING_SCHEMA, "yep").setString(index, "yep");
```
shikhar: missing `verifyBindField()` calls for`decimal`, `date`, `time`, and `timestamp` here

src/test/java/io/confluent/connect/jdbc/sink/PreparedStatementBinderTest.java
```diff
@@ -108,57 +130,59 @@ public void bindRecord() throws SQLException {
...
+    final List<Schema> nullableTypes = Arrays.asList(
+        Schema.INT8_SCHEMA,
+        Schema.INT16_SCHEMA,
+        Schema.INT32_SCHEMA,
+        Schema.INT64_SCHEMA,
+        Schema.FLOAT32_SCHEMA,
+        Schema.FLOAT64_SCHEMA,
+        Schema.BOOLEAN_SCHEMA,
+        Schema.BYTES_SCHEMA,
+        Schema.STRING_SCHEMA
```
shikhar: we should add the logical types' schemas here

shikhar:
Thank you for the patch @YAtOff! Looking good aside from the comments.

shikhar:
We also need support for logical types in `DbDialect` for `CREATE` & `ALTER` DDL statements, but that can be a separate PR.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -85,28 +137,34 @@ public void bindRecord(SinkRecord record) throws SQLException {
...
 
-  static void bindField(PreparedStatement statement, int index, Schema.Type type, Object value) throws SQLException {
+  static void bindField(PreparedStatement statement, int index, Schema schema, Object value) throws SQLException {
     if (value == null) {
       statement.setObject(index, null);
     } else {
+      if (schema.name() != null) {
+        LogicalTypeConverter logicalConverter = LOGICAL_CONVERTERS.get(schema.name());
+        if (logicalConverter != null)
+          value = logicalConverter.convert(schema, value);
```
shikhar: I think it might be better here to bind the logical types directly with 
`statement.setDate(..)`, `statement.setTime(..)`, `statement.setTimestamp(..), `statement.setBigDecimal(..)` here. But I haven't looked into this properly and this seems fine as a first pass at logical type support.

YAtOff:
Thanks for the feedack @shikhar.  I'm on vacation now. When I come back,  I'll fix it.

ewencp:
ok to test

shikhar:
@YAtOff out of curiosity, what database were you using the JDBC sink against?

YAtOff:
SQLite.
But I'm planning to user MySQL.
And here I have the following problem: with MySQL `DbMetadataQueries#getTableMetadata` cannot read the table columns. But I haven't investigated it.

shikhar:
Good to know, thanks @YAtOff. We have some system tests against MySQL and there hasn't been any issue with `DbMetadataQueries#getTableMetadata`. I'd welcome any detail you can provide  on the CP mailing list (https://groups.google.com/forum/#!forum/confluent-platform), or feel free to email me directly shikhar@confluent.io.

Issue #32, opened 2016-01-06
Database schema support 


tma625:
I have a Postgres database with multiple tables of the same name, but in different schemas. I'm not seeing a way to specify the schemas for the tables in the properties file, and if I provide the entire "schema.tablename" as the name in the properties file that doesn't seem to work either. 

It looks like there are a couple of problems causing this. First, `JdbcUtils.getTables` returns just a list of table names (without the schema name prepended). Second, also in JdbcUtils, whenever `DatabaseMetaData.getTables` is called, it's always called with `null` as the second parameter for the schema, which results in all of the schemas being searched. 

I think the fix would be allowing for the "schema.tablename" syntax in the table whitelist or blacklist, and then parsing schema out of this string. `TableQuerier` would also need an additional field for this schema in order to pass this schema along to `JdbcUtils`.

I've got a proof of concept working, but I was curious if this was a known issue or if support would be added soon?

Thanks

ewencp:
@tma625 You're correct that we are currently limited to a single schema. This was an intentional choice for the first release -- there's a limited number of features we could fit into this first round of development. We know that this will be important functionality and it will be added.

I actually don't think it will be that difficult to add -- as you pointed out, there are a couple of key places that we need to be careful about specifying or retrieving the schema. From there, hopefully most of the rest of the fallout is mostly mechanical changes to use schema + table rather than just table.

shikhar:
This was addressed with #109

PR #109, opened 2016-08-11
Ability to set the schema pattern for tables metadata retrieval

pqueixalos:
This adds the ability to configure jdbc schemaPattern from connector configuration.
Reported here https://github.com/confluentinc/kafka-connect-jdbc/issues/32

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -16,6 +16,12 @@
 
 package io.confluent.connect.jdbc;
 
+import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
+import io.confluent.connect.jdbc.source.JdbcSourceTask;
+import io.confluent.connect.jdbc.source.JdbcSourceTaskConfig;
+import io.confluent.connect.jdbc.source.TableMonitorThread;
+import io.confluent.connect.jdbc.util.StringUtils;
+import io.confluent.connect.jdbc.util.Version;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.connect.connector.Task;
@@ -35,13 +41,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
-import io.confluent.connect.jdbc.source.JdbcSourceTask;
-import io.confluent.connect.jdbc.source.JdbcSourceTaskConfig;
-import io.confluent.connect.jdbc.source.TableMonitorThread;
-import io.confluent.connect.jdbc.util.StringUtils;
-import io.confluent.connect.jdbc.util.Version;
 
 /**
  * JdbcConnector is a Kafka Connect Connector implementation that watches a JDBC database and
@@ -96,6 +95,7 @@ public class JdbcSourceConnector extends SourceConnector {
                                  + JdbcSourceConnectorConfig.TABLE_BLACKLIST_CONFIG + " are "
                                  + "exclusive.");
     String query = config.getString(JdbcSourceConnectorConfig.QUERY_CONFIG);
+    String schemaPattern = config.getString(JdbcSourceConnectorConfig.SCHEMA_PATTERN_CONFIG);
     if (!query.isEmpty()) {
       if (whitelistSet != null || blacklistSet != null)
         throw new ConnectException(JdbcSourceConnectorConfig.QUERY_CONFIG + " may not be combined"
@@ -104,7 +104,7 @@ public class JdbcSourceConnector extends SourceConnector {
       // query.
       whitelistSet = Collections.emptySet();
     }
-    tableMonitorThread = new TableMonitorThread(db, context, tablePollMs, whitelistSet,
+    tableMonitorThread = new TableMonitorThread(db, schemaPattern, context, tablePollMs, whitelistSet,
                                                 blacklistSet, tableTypesSet);
     tableMonitorThread.start();
   }
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -105,7 +105,7 @@ public class JdbcSourceConnector extends SourceConnector {
       // query.
       whitelistSet = Collections.emptySet();
     }
-    tableMonitorThread = new TableMonitorThread(db, schemaPattern, context, tablePollMs,
+    tableMonitorThread = new TableMonitorThread(db, context, schemaPattern, tablePollMs,
                                                 whitelistSet, blacklistSet, tableTypesSet);
     tableMonitorThread.start();
   }
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -16,12 +16,6 @@
 
 package io.confluent.connect.jdbc;
 
-import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
-import io.confluent.connect.jdbc.source.JdbcSourceTask;
-import io.confluent.connect.jdbc.source.JdbcSourceTaskConfig;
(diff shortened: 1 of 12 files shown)


pqueixalos:
CLA just signed.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -144,6 +143,9 @@
   public static final long TIMESTAMP_DELAY_INTERVAL_MS_DEFAULT = 0;
   private static final String TIMESTAMP_DELAY_INTERVAL_MS_DISPLAY = "Delay Interval (ms)";
 
+  public static final String SCHEMA_PATTERN_CONFIG = "schemaPattern";
```
shikhar: `schema.pattern` to be consistent with the config naming conventions

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -157,9 +155,9 @@ protected ResultSet executeQuery() throws SQLException {
       stmt.setLong(3, incOffset);
       stmt.setTimestamp(4, tsOffset, UTC_CALENDAR);
       log.debug("Executing prepared statement with start time value = {} end time = {} and incrementing value = {}",
-              JdbcUtils.formatUTC(tsOffset),
```
shikhar: minor: can you please remove the whitespace changes

shikhar:
@pqueixalos is it possible to just include the schema in the connection URL?

pqueixalos:
@shikhar, short answer : no. 
Settting the schema in URL configures the JDBC Driver and not all JDBC implementations supports this.
Also it would not change anything (I tried :)), the main issue is that JdbcUtils is aksing for database metadata without providing any schema pattern so every tables of every 'usable' schemas are retrieved.

Regarding the changes you asked (whitespace and parameter naming, sure, I will make those changes next week).

shikhar:
@pqueixalos would it work to update the code to use `connection.getSchema()` as the schema pattern? (in conjunction with the user specifying it as part of the connection URL)

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -16,6 +16,12 @@
 
 package io.confluent.connect.jdbc;
 
+import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
```
shikhar: let's avoid the import rearrangement

pqueixalos:
Good point @shikhar, getting the schema from the connection (when set in jdbc connection params) might work fine in some situations !
It won't work in my context for sure, as (silly) Redshift connector does not support it. Also, I don't know how much other jdbc connector implementations honor this, for instance MySQL would probably return null as it does not implement schema concept, connection.getCatalog() would need to be used instead.

pqueixalos:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -104,8 +105,8 @@ public void start(Map<String, String> properties) throws ConnectException {
       // query.
       whitelistSet = Collections.emptySet();
     }
-    tableMonitorThread = new TableMonitorThread(db, context, tablePollMs, whitelistSet,
-                                                blacklistSet, tableTypesSet);
+    tableMonitorThread = new TableMonitorThread(db, schemaPattern, context, tablePollMs,
```
shikhar: @pqueixalos sorry didn't notice this minor thing previously, but in terms of arg ordering `schemaPattern` makes sense after the context

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -73,6 +74,7 @@ public void start(Map<String, String> properties) {
       throw new ConnectException("Couldn't start JdbcSourceTask due to configuration error", e);
     }
 
+    this.schemaPattern = config.getString(JdbcSourceTaskConfig.SCHEMA_PATTERN_CONFIG);
```
shikhar: I don't think we need this as a field on this class.

shikhar:
@pqueixalos ah too bad Redshift JDBC does not support that. So we need this patch to make the JDBC source connector usable against Redshift then.

Left a couple more comments, otherwise LGTM

cc @ewencp in case you have any thoughts on this

shikhar:
ok to test

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -190,6 +192,7 @@ public static ConfigDef baseConfigDef() {
         .define(BATCH_MAX_ROWS_CONFIG, Type.INT, BATCH_MAX_ROWS_DEFAULT, Importance.LOW, BATCH_MAX_ROWS_DOC, CONNECTOR_GROUP, 2, Width.SHORT, BATCH_MAX_ROWS_DISPLAY)
         .define(TABLE_POLL_INTERVAL_MS_CONFIG, Type.LONG, TABLE_POLL_INTERVAL_MS_DEFAULT, Importance.LOW, TABLE_POLL_INTERVAL_MS_DOC, CONNECTOR_GROUP, 3, Width.SHORT, TABLE_POLL_INTERVAL_MS_DISPLAY)
         .define(TOPIC_PREFIX_CONFIG, Type.STRING, Importance.HIGH, TOPIC_PREFIX_DOC, CONNECTOR_GROUP, 4, Width.MEDIUM, TOPIC_PREFIX_DISPLAY)
+        .define(SCHEMA_PATTERN_CONFIG, Type.STRING, null, Importance.MEDIUM, SCHEMA_PATTERN_DOC, DATABASE_GROUP, 5, Width.SHORT, QUERY_DISPLAY)
```
ewencp: Is this in the right location? If it is part of the `DATABASE_GROUP` should we relocate it with that group? And I think the order might not be quite right since it looks like there are only 3 items in the `DATABASE_GROUP` so far. (It'd also be nice to include a DISPLAY setting as well for a more human-friendly form.)

ewencp:
@pqueixalos @shikhar Only took a quick scan through, but the changes seem fine. Major missing piece seems to be tests -- there were some updates to tests but it looks like they just use `null` for the schema pattern so we haven't actually tested the case where the user sets the schema pattern.

pqueixalos:
@ewencp I added a test in `JdbcUtilsTest`, but taht is probably not enough. Making sure that the configuration is well handled at `JdbcSourceConnector` level would be better. I will take care of this on a later commit.
I also improved the configuration documentation as you suggested, hoping I did well.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -108,6 +108,14 @@
   public static final String TABLE_BLACKLIST_DEFAULT = "";
   private static final String TABLE_BLACKLIST_DISPLAY = "Table Blacklist";
 
+  public static final String SCHEMA_PATTERN_CONFIG = "schema.pattern";
+  private static final String SCHEMA_PATTERN_DOC =
+      "Schema pattern to fetch tables metadata from:\n"
+      + "  * \"\" retrieves those without a schema,"
+      + "  * null (default) means that the schema name should not be used to narrow the search, all tables "
+      + "metadata would be fetched, regardless there schema.";
```
shikhar: typo: `s/there/their`

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -108,6 +108,14 @@
   public static final String TABLE_BLACKLIST_DEFAULT = "";
   private static final String TABLE_BLACKLIST_DISPLAY = "Table Blacklist";
 
+  public static final String SCHEMA_PATTERN_CONFIG = "schema.pattern";
+  private static final String SCHEMA_PATTERN_DOC =
+      "Schema pattern to fetch tables metadata from:\n"
```
shikhar: "from the database"

shikhar:
@pqueixalos a few minor comments about the doc, otherwise LGTM

shikhar:
Thanks for the contrib @pqueixalos!

PR #124, opened 2016-09-06
Schema name prefixes - fixes #123

### LINKED ISSUE #123: Support for schema prefixes.
The schemas that are currently generated are based on the table name. When using the [AvroConverter](https://github.com/confluentinc/schema-registry) from the [Confluent Schema Registry](https://github.com/confluentinc/schema-registry), the resulting avro schema does not have a namespace. This causes the [avro-maven-plugin](https://avro.apache.org/docs/1.8.1/gettingstartedjava.html) to attempt to place the generated source in default java package. The following exception is raised. 

```
 error: `package' expected but `public' found.
[INFO] public class eng_match extends org.apache.avro.specific.SpecificRecordBase implements org.apache.avro.specific.SpecificRecord {
[INFO] ^
[ERROR] /Users/raosa/gitprojects/examples/kafka-streams/target/generated-sources/eng_match.java:3619: error: identifier expected but eof found.
[INFO] }
[INFO]  ^
[ERROR] two errors found
```

jcustenborder:
Initial commits to add support for prefixing schema names. A little direction on where to add unit testing would be great. Currently setting the prefix to an empty string which ensures that the current behavior is maintained.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
@@ -36,8 +36,8 @@ import io.confluent.connect.jdbc.util.JdbcUtils;
 public class BulkTableQuerier extends TableQuerier {
   private static final Logger log = LoggerFactory.getLogger(BulkTableQuerier.class);
 
-  public BulkTableQuerier(QueryMode mode, String name, String topicPrefix) {
-    super(mode, name, topicPrefix);
+  public BulkTableQuerier(QueryMode mode, String name, JdbcSourceTaskConfig jdbcSourceTaskConfig) {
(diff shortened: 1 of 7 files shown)


shikhar:
@jcustenborder this needs rebase since #109 was merged

ewencp:
@jcustenborder As mentioned, needs rebase, but otherwise seems good to me. Testing can be pretty minimal as it is currently in the `JdbcSourceTaskConversionTest`.

gwenshap:
@jcustenborder this can now be handled with org.apache.kafka.connect.transforms.RegexRouter - I didn't try but I expect the config to be something like:
regex=(.*)
replacement=prefix.$1

Want to check if this does the trick? I can add it to the docs as an example if this works :)

Issue #125, opened 2016-09-08
Problem with sql server using JDBC Connector using Decimal for incrementingColumn


wegged:
Hi I am trying to use Kafka Connect with some vendor data in SQL Server.  The vendor is using DECIMAL(10,0) for their primary key.  this is resulting in the following exception.  Please advise.

We are using jtds 1.3.1 jdbc driver.

Thanks,
Eli

> 
> [2016-09-08 16:02:30,221] ERROR Task sql12Alloc-0 threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask:142)
> org.apache.kafka.connect.errors.ConnectException: Invalid type for incrementing column: BYTES
>         at io.confluent.connect.jdbc.TimestampIncrementingTableQuerier.extractRecord(TimestampIncrementingTableQuerier.java:191)
>         at io.confluent.connect.jdbc.JdbcSourceTask.poll(JdbcSourceTask.java:217)
>         at org.apache.kafka.connect.runtime.WorkerSourceTask.execute(WorkerSourceTask.java:155)
>         at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:140)
>         at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:175)
>         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
>         at java.util.concurrent.FutureTask.run(FutureTask.java:266)
>         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
>         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
>         at java.lang.Thread.run(Thread.java:745)
> [2016-09-08 16:02:30,236] ERROR Task is being killed and will not recover until manually restarted (org.apache.kafka.connect.runtime.WorkerTask:143)

shikhar:
We should be able to fix this and #31 by coercing the `Decimal` logical type in `TimestampIncrementingTableQuerier.extractRecord()` to `long`.

We could make sure that the value is less than `Long.MAX_VALUE` before the downcast from `BigDecimal` and otherwise throw an error.

Asim23:
@shikhar I have to import data from kafka to my local sql server (on my local machine). For kafka i am using landoop docker image. i have configured jdbc connector on http://127.0.0.1:3030/kafka-connect-ui
[...]

Issue #97, opened 2016-06-30
Add support for string incrementing columns


wkiser:
Would it be possible to add support for using a string column as the incrementing column? Looking at the code a little it seems like it might be as easy as making TimestampIncrementingTableQuerier generic.

ewencp:
@wkiser That might be possible with a simple change as you mentioned, as long as we have all the type info we need from the table metadata before creating the querier.

Out of curiousity, why strings? This seems like a pretty unusual use case unless you have some other type being encoded as a string and which ensures lexicographic ordering on the strings matches the ordering of the underlying type.

wkiser:
Hey @ewencp, you pretty much described the structure of the data that I am dealing with.

The tables I'm querying have a fixed width string primary key, which is a combination of numbers and letters that are lexicographically sortable. I have a temporary solution working where I create a new, encoded 'id' column at query time, but it is really inefficient. Normally I would just add an autoincrement column, but this table is very large.

PR #129, opened 2016-09-13
CC-311: support for Decimal logical type as incrementing column

### LINKED ISSUE #31: Issue with Oracle column type NUMBER set as "Incrementing Column" in Kafka Connect
Kafka-connect-jdbc program has an issue with Oracle Table that has a Primary Key of Type NUMBER and if we use that for Column as Incrementing (in the mode settings of the Kafka-Connect program).

It appears (from our debugging efforts) we see that NUMBER => java.math.BigDecimal => NUMERIC => BYTES in the kafka-connect-jdbc code. Essentially, the SchemaBuilder.java sets it to Type.BYTES.

And as result, we see the error "Invalid type of Incrementing column: BYTES" when we connect to the Oracle Database. This is a rather severe limitation for the Kafka Connect to be able to work with Oracle Database as most of the PK columns are of type NUMBER

Steps to Reproduce this Issue:

+ Set-up a table in Oracle 

<pre><code>create table TEST_KAFKA_CONNECT {
KAKFA_PK NUMBER not null, UPDATED_AT timestamp not null } </code></pre>

+ Configure the Kafka-Connect

<pre><code>mode=timestamp+incrementing
incrementing.column.name=KAKFA_PK
timestamp.column.name=UPDATED_AT </code></pre>

In the Kafka-Connect-JDBC use the Oracle's JDBC driver (tested with Oracle 11.2.0.2.0; driver corresponding Oracle JDBC driver available here: http://www.oracle.com/technetwork/apps-tech/jdbc-112010-090769.html

shikhar:
Only supported when the scale is 0 (i.e. it is an integer) and <= Long.MAX_VALUE, so we can continue storing it as a long for implementation simplicity

Fixes #31 and #125

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
@@ -16,12 +16,15 @@
 
 package io.confluent.connect.jdbc.source;
 
+import org.apache.kafka.connect.data.Decimal;
+import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.math.BigDecimal;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -55,6 +58,7 @@ public class TimestampIncrementingTableQuerier extends TableQuerier {
   private static final Logger log = LoggerFactory.getLogger(TimestampIncrementingTableQuerier.class);
 
   private static final Calendar UTC_CALENDAR = new GregorianCalendar(TimeZone.getTimeZone("UTC"));
+  private static final BigDecimal LONG_MAX_VALUE_AS_BIGDEC = new BigDecimal(Long.MAX_VALUE);
 
   private String timestampColumn;
   private String incrementingColumn;
@@ -145,8 +149,6 @@ public class TimestampIncrementingTableQuerier extends TableQuerier {
     stmt = db.prepareStatement(queryString);
   }
 
-
-
   @Override
   protected ResultSet executeQuery() throws SQLException {
     if (incrementingColumn != null && timestampColumn != null) {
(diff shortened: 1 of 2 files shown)


shikhar:

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -229,6 +202,56 @@ public SourceRecord extractRecord() throws SQLException {
...
+      if (incrementingColumnValue == null) {
+        throw new ConnectException("Null value for incrementing column of type: " + incrementingColumnSchema.type());
+      } else if (isIntegralPrimitiveType(incrementingColumnValue)) {
+        extractedId = ((Number) incrementingColumnValue).longValue();
+      } else if (incrementingColumnSchema.name() != null && incrementingColumnSchema.name().equals(Decimal.LOGICAL_NAME)) {
+        final BigDecimal decimal = ((BigDecimal) incrementingColumnValue);
+        if (decimal.compareTo(LONG_MAX_VALUE_AS_BIGDEC) > 0) {
+          throw new ConnectException("Decimal value for incrementing column exceeded Long.MAX_VALUE");
+        }
+        if (decimal.scale() != 0) {
```
ewencp: An alternative to these restrictions would be to just make `TimestampIncrementingOffset` use a `BigDecimal` and convert the other types (or even make it a composite where you can choose the `long` or the `BigDecimal`. Any reason not to be a bit more liberal by doing that? Performance is the only reason I can think of.

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -229,6 +202,56 @@ public SourceRecord extractRecord() throws SQLException {
...
+      assert incrementingOffset == -1L || extractedId > incrementingOffset || timestampColumn != null;
+    } else {
+      extractedId = null;
+    }
+
+    return new TimestampIncrementingOffset(extractedTimestamp, extractedId);
+  }
+
+  private boolean isIntegralPrimitiveType(Object incrementingColumnValue) {
+    return incrementingColumnValue instanceof Long
```
ewencp: Probably doesn't matter much in practice, but any reason we switched to checking the value rather than the schema?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
shikhar: I was hesitant to store `BigDecimal` when it's some other int type because wasn't sure if JDBC drivers would handle `statement.setBigDecimal()` well for such columns.

BTW, I tried to be more general in a local iteration by storing `Comparable`, in fact even including support for `String` types like in #97 but abandoned after complexity around default values. I have an idea now of how to handle that though, so I'll give this another go.

shikhar: Backwards compatibility is another concern to take care of... so we should probably always map to `Long` for any of the currently supported int types.

ewencp: Wait, this isn't going to get stored in the database, right? This is in the source so we are extracting it. I think the major difference would be that the offsets would change from their integer type to `Decimal`.

shikhar: > so I'll give this another go.

Er, re-abandoned. The issue is, if there is no existing stored offset, you have a `null` but no type. While a sensible default value can be determined ("" for string, -1 for numbers..), need type-info at bind-time.

I think happy with the solution for now. We could expand support to all `Number` types rather than just upto `Long.MAX_VALUE` if there is a request for it.

shikhar: > Wait, this isn't going to get stored in the database, right? This is in the source so we are extracting it.

It gets stored in the offset storage, and used for range queries

ewencp: Ok, makes sense.

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
shikhar: No good reason :-) Can't recall how I ended up switching to `instanceof`... shouldn't matter as you said

PR #131, opened 2016-09-14
CC-277: allow whitelisting of fields to be written to the db

shikhar:


shikhar:

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -52,26 +52,27 @@ public static FieldsMetadata extract(
...
       final JdbcSinkConfig.PrimaryKeyMode pkMode,
       final List<String> configuredPkFields,
+      final Set<String> fieldsWhitelist,
       final SchemaPair schemaPair
   ) {
-    return extract(tableName, pkMode, configuredPkFields, schemaPair.keySchema, schemaPair.valueSchema);
+    return extract(tableName, pkMode, configuredPkFields, fieldsWhitelist, schemaPair.keySchema, schemaPair.valueSchema);
   }
 
   public static FieldsMetadata extract(
```
shikhar: The logic in here is extremely hairy since tnhere is interplay of `pk.mode`, `pk.fields` and now `fields.whitelist` against what we actually find in the key & value schema of a record.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -131,16 +131,24 @@
...
       + "    Must be a trio representing the Kafka coordinates, defaults to ``" + StringUtils.join(DEFAULT_KAFKA_PK_NAMES, ",") + "`` if empty.\n"
       + "`record_key`\n"
-      + "    If empty, all fields from the key struct will be used, otherwise used to whitelist the desired fields - for primitive key only a single field name must be configured.\n"
+      + "    If empty, all fields from the key struct will be used, otherwise used to extract the desired fields - for primitive key only a single field name must be configured.\n"
       + "`record_value`\n"
-      + "    If empty, all fields from the value struct will be used, otherwise used to whitelist the desired fields.\n";
+      + "    If empty, all fields from the value struct will be used, otherwise used to extract the desired fields.";
   private static final String PK_FIELDS_DISPLAY = "Primary Key Fields";
 
+  public static final String FIELDS_WHITELIST = "fields.whitelist";
```
ewencp: Hmm, I'm realizing this is a bit inconsistent with the PK version since this uses `whitelist` whereas the PK one just says `pk.fields`. Should we try to unify them on `whitelist` or does that just add noise?

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -161,19 +160,29 @@ public static FieldsMetadata extract(
...
 
     }
 
+    final Set<String> nonKeyFieldNames = new LinkedHashSet<>();
     if (valueSchema != null) {
-      for (Field field : valueSchema.fields()) {
-        final boolean isKeyField = keyFieldNames.contains(field.name());
-        if (!isKeyField) {
-          nonKeyFieldNames.add(field.name());
+      for (Field field: valueSchema.fields()) {
```
ewencp: This might be easier to understand when expressed as set operations, i.e. valueSchema.fields() - keyFieldNames) and fieldsWhitelist. I haven't worked through exactly what that looks like w/ Java's set APIs, but might reduce this to a one-liner that's easier to understand.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -161,19 +160,29 @@ public static FieldsMetadata extract(
...
+        }
+        if (!fieldsWhitelist.isEmpty() && !fieldsWhitelist.contains(field.name())) {
+          continue;
         }
+
+        nonKeyFieldNames.add(field.name());
+
         final Schema fieldSchema = field.schema();
-        allFields.put(field.name(), new SinkRecordField(fieldSchema.type(), field.name(), isKeyField, !isKeyField && fieldSchema.isOptional(), fieldSchema.defaultValue()));
+        allFields.put(field.name(), new SinkRecordField(fieldSchema.type(), field.name(), false, fieldSchema.isOptional(), fieldSchema.defaultValue()));
```
ewencp: It seems like the behavior was changed here? `allFields` doesn't actually contain all the fields anymore (which is pretty clear since one of the flags is always `false`). Either the meaning of `allFields` changed and we should probably rename to clarify, or we may not have test coverage for something (it looks like it is used, for example, to determine if we need to alter the table).

shikhar: `allFields` is also being manipulated in the `switch (pkMode)` block above.

Previously, for pkMode = KAFKA, RECORD_KEY we were adding to `allFields` in the above block, but for RECORD_VALUE we were only adding to `allFields` over here along with all the non-key fields. Now, the key fields for RECORD_VALUE pkMode are also inserted into `allFields` in the switch block. Over here we only deal with the non-key fields. https://github.com/confluentinc/kafka-connect-jdbc/pull/131/files#diff-4b72122fa7b6eb8a030b667da3073382R163

ewencp: Ah, got it, makes sense.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
shikhar: `pk.fields` is I think a more descriptive name because that config is a bit overloaded. It is a whitelist in some cases and in others it specifies the names to use (e.g. for `pk.mode=kafka` it must be exactly 3 names for topic/partition/offset, and for `pk.mode=record_key` where you expect the record key to be a primitive type you can assign the name to use for the primitive).

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
shikhar: Java set API's don't make this any nicer IMO :-) (with some added cost from the temporary collections) With JDK8 streams it could be a lot cleaner though, when we can use it.

ewencp:
@shikhar Explanations all make sense. LGTM

shikhar:
Thanks @ewencp

PR #132, opened 2016-09-17
Fixes tests failing at clean up due to locked resources.

meadowlark-bradsher:
Some tests fail out of the box in my dev environment (Windows 7, Java 1.7.0_79, Intellij IDEA 15.0.6). The majority of these are solved by adding task.stop() or writer.closeQuietly() to the end of tests. In a few cases when exception were expected all passing through the method...

JdbcDbWriterTest.writeSameRecordTwiceExpectingSingleUpdate

... a try-with-resources was used to catch the exception, passing the expected SQLException and Assert.fail(String) on the others. Because Assert.fail(String) is deprecated in junit.framework.Assert the new package org.junit.Assert was used.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
@@ -30,7 +30,7 @@ import java.util.Map;
 
 import io.confluent.connect.jdbc.sink.dialect.DbDialect;
 
-public class JdbcDbWriter {
+public class JdbcDbWriter implements AutoCloseable{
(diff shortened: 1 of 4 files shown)


ewencp:
ok to test

shikhar:
@meadowlark-bradsher what do you think about making `writer` a field, which can be initialized in tests. Then in `tearDown()` we can have a block

```java
if (writer != null) writer.closeQuietly();
```

shikhar:
@meadowlark-bradsher I committed the above-mentioned approach in https://github.com/confluentinc/kafka-connect-jdbc/commit/325c59f31a9c4cdd54082b09cd074005212d82d1

Please report if you still run into failures in your env.

Issue #133, opened 2016-09-19
CC-278: Better support for logical types


shikhar:
In #108 we got basic support for logical types, which makes it so we use the underlying datatype when binding the record to a statement. There are some enhancements we should make:

– We need support for logical types in `DbDialect` and its implementations for `CREATE` & `ALTER` DDL statements, so that we default to an appropriate database-specific datatype definition which many databases provide. This would involve some refactoring as the current code looks up based on the `schema.type()` but for logical types the `schema.name()` needs to be considered.

– We should bind  with `statement.setDate(..)`, 
`statement.setTime(..)`, `statement.setTimestamp(..)`, `statement.setBigDecimal(..)` rather than the internal datatype used for representing it (ints for data/time types and bytes for decimal). In particular for Decimal, bytes is very unlikely to be the desired representation in the database.

Logical types included with Connect: `org.apache.connect.data.Date`, `org.apache.connect.data.Time`, `org.apache.connect.data.Timestamp`, `org.apache.connect.data.Decimal`

PR #138, opened 2016-09-26
CC-278: add logical type mappings for more sink dialects

shikhar:
Refactored dialect tests to make it easier to test the additional mappings, that's where most of the diff stems from


shikhar:
@willyhoang for PG
@santi81 for HANA

willyhoang:
LGTM for PG!

shikhar:
Linking #133 as the parent issue for this.

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
```diff
@@ -121,6 +138,22 @@ protected void writeColumnSpec(StringBuilder builder, SinkRecordField f) {
   }
 
   protected void formatColumnValue(StringBuilder builder, String schemaName, Map<String, String> schemaParameters, Schema.Type type, Object value) {
+    if (schemaName != null) {
+      switch (schemaName) {
+        case Decimal.LOGICAL_NAME:
+          builder.append(value);
```
ewencp: In some (maybe all?) databases, this could exceed the maximum value supported by the underlying type, right? e.g. in MySQL you have to map to a 65 bit maximum unscaled value. However, this isn't checking for that (nor are each of the dialect classes). Is this intentional and we just expect an error to be thrown, or did we just miss the check?

src/main/java/io/confluent/connect/jdbc/sink/dialect/SqliteDialect.java
```diff
@@ -36,6 +40,15 @@ public SqliteDialect() {
 
   @Override
   protected String getSqlType(String schemaName, Map<String, String> parameters, Schema.Type type) {
+    if (schemaName != null) {
+      switch (schemaName) {
+        case Decimal.LOGICAL_NAME:
+        case Date.LOGICAL_NAME:
+        case Time.LOGICAL_NAME:
+        case Timestamp.LOGICAL_NAME:
+          return "NUMERIC";
```
ewencp: Does SQLite not need a custom implementation of `formatColumnValue`? My (very quick) reading of [the docs](https://www.sqlite.org/datatype3.html#date_and_time_datatype) suggests you might need to use the datetime methods to get things in the desired format? In particular, because sqlite seems to support flexible column formats, I'm unclear whether passing in strings and marking the column type `NUMERIC` will actually get you data in the format you expect. In particular, look at [the rules](https://www.sqlite.org/datatype3.html#type_affinity) for what happens if you pass in `TEXT` for a `NUMERIC` column....

shikhar: Good catch!

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
shikhar: I was hoping to leave this to the relevant JDBC driver or database. The db's decimal type may have a limited precision/scale, and JDBC allows for `PreparedStatement.setBigDecimal()` where the `BigDecimal` has its own precison/scale, I trust there needs to be some mapping/validation there... perhaps something to exercise in a system test.

src/main/java/io/confluent/connect/jdbc/sink/dialect/SqliteDialect.java
shikhar: @ewencp fixed in 104225afe78405b1f554f4462fcca4927acc50fd. Also included another change in that commit to map to `INTEGER` for integral types rather than `NUMERIC` (there seems to be some casting nuance in which they are different and `INTEGER` is more accurate anyway). Since we are anyway doing integration-style testing for SQLite also  folded in logical type conversions into an existing test.

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
shikhar: Ah so over here we are just appending to the SQL directly not binding to a `PreparedStatement`... but this is only used for formatting default values into DDL so a low surface area for issues (and unlikely that decimal default values would push precision boundaries).

PR #142, opened 2016-10-11
Declaring SimpleDateFormat as thread-local

picadoh:
SimpleDateFormat is not thread-safe and this flow runs inside the sink tasks which run concurrently with each other, so this may incur in concurrency issues. 3 options for this would be 1) create a separate instance each time, 2) synchronise the access to the formatter 3) declare it as thread local.

I'm suggesting option 3) in this PR, as from past experience it's the one from the above with less performance impacts.

There are 2 more options, both are thread-safe but require an extra dependency to be added: 4) joda-time DateTimeFormatter, 5) commons-lang FastDateFormat.

Any feedback is welcome.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
@@ -41,31 +41,31 @@ import static io.confluent.connect.jdbc.sink.dialect.StringBuilderUtil.joinToBui
 import static io.confluent.connect.jdbc.sink.dialect.StringBuilderUtil.nCopiesToBuilder;
 
 public abstract class DbDialect {
+  private static final TimeZone UTC = TimeZone.getTimeZone("UTC");
 
   protected static final ThreadLocal<SimpleDateFormat> DATE_FORMAT = new ThreadLocal<SimpleDateFormat>() {
     protected SimpleDateFormat initialValue() {
-      return new SimpleDateFormat("yyyy-MM-dd");
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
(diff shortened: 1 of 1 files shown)


picadoh:
[clabot:check]

ewencp:
ok to test

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
```diff
@@ -42,15 +42,29 @@
...
+      return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
+    }
+  };
 
   static {
     final TimeZone utc = TimeZone.getTimeZone("UTC");
-    DATE_FORMAT.setTimeZone(utc);
-    TIME_FORMAT.setTimeZone(utc);
-    TIMESTAMP_FORMAT.setTimeZone(utc);
+    DATE_FORMAT.get().setTimeZone(utc);
```
ewencp: Since this will only run on the thread where static initialization happens, don't we instead need to move these `setTimeZone` calls up into the `ThreadLocal.initialValue` calls?

picadoh: Yep, makes perfect sense @ewencp . Fixed it. Thanks.

shikhar:
LGTM, thanks @picadoh 

Out of curiosity is this an issue you noticed while scanning the code, or something you ran into?

picadoh:
Hey @shikhar, thanks sir. I ran into a similar issue before in other code base and these are really tricky ones to find later. I've been doing some testing with this particular connector (Great job here btw) so I've been watching the repo and noticed the potential concurrency issue in one of the commits.

PR #149, opened 2016-10-17
CC-236: JDBC Sink retries test

shikhar:


shikhar:

src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkTaskTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import static org.easymock.EasyMock.expectLastCall;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class JdbcSinkTaskTest extends EasyMockSupport {
```
ewencp: Not sure if this was intentional, but there is now `SinkTaskTest` and `JdbcSinkTaskTest`. I think the former was just inconsistently named.

src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkTaskTest.java
```diff
@@ -0,0 +1,100 @@
...
+  public void retries() {
+    final int maxRetries = 2;
+    final int retryBackoffMs = 1000;
+
+    Set<SinkRecord> records = Collections.singleton(new SinkRecord("stub", 0, null, null, null, null, 0));
+    final JdbcDbWriter mockWriter = createMock(JdbcDbWriter.class);
+    SinkTaskContext ctx = createMock(SinkTaskContext.class);
+
+    try {
+      mockWriter.write(records);
```
ewencp: Doesn't much matter in this case since there's only one call, but I've also found that for these tests, adding a `throws SQLException` annotation to the test works fine in place of actually catching it and rethrowing since we'd never be throwing `SQLExceptions` ourselves anyway.

shikhar: good point

PR #144, opened 2016-10-12
CC-263: Prevent retrying queries with a broken connection

shikhar:
Fixes #38

PR #154, opened 2016-10-25
Re-check stop flag before continuing after sleep

shikhar:


shikhar:
This issue crept in with #144

Worst case is that the connection will be closed by the time we attempt to query on it due to a concurrent `stop()`.

shikhar:
@ewencp for review

ewencp:
@shikhar LGTM

PR #156, opened 2016-10-25
CC-348: Separate username & password config for JDBC source connector

### LINKED ISSUE #137: Create separate username/password configs for JDBC source connector
As reported in https://github.com/confluentinc/common/issues/38, all config fields are dumped to logs (password fields get masked). Currently the only way to provide the username & password for the source connector is as part of the `connection.url` config. We should make a separate fields for the username/password credentials, like in the JDBC sink, such that the password field is of `ConfigDef.Type.PASSWORD`.

shikhar:
Allows treating the password as `Type.PASSWORD` rather than having to embed it in the URL

Fixes #137


dbtucker:
Can we confirm the proper functionality of the PASSWORD type in Control Center 3.1.0   There was an issue with 3.0.x where the UI did not properly handle PASSWORD configuration values (they did not show up at all)

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -215,13 +226,15 @@ public JdbcSourceConnectorConfig(Map<String, String> props) {
...
+      String dbUser = (String) config.get(CONNECTION_USER_CONFIG);
+      Password dbPassword = (Password) config.get(CONNECTION_PASSWORD_CONFIG);
       String schemaPattern = (String) config.get(JdbcSourceTaskConfig.SCHEMA_PATTERN_CONFIG);
       if (dbUrl == null) {
         throw new ConfigException(CONNECTION_URL_CONFIG + " cannot be null.");
       }
       Connection db;
       try {
-        db = DriverManager.getConnection(dbUrl);
+        db = DriverManager.getConnection(dbUrl, dbUser, dbPassword == null ? null : dbPassword.value());
```
ewencp: Is `null` for the `user` parameter guaranteed to be safe? Since this goes to `DriverManager` I'd want to make sure it's generally safe, but I didn't see anything in the docs suggesting its safe. Did I miss somewhere suggesting it is safe, or should we be checking that both user and password are non-null?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -36,8 +37,16 @@
 public class JdbcSourceConnectorConfig extends AbstractConfig {
 
   public static final String CONNECTION_URL_CONFIG = "connection.url";
-  private static final String CONNECTION_URL_DOC = "JDBC connection URL for the database to load.";
-  private static final String CONNECTION_URL_DISPLAY = "Connection Url";
+  private static final String CONNECTION_URL_DOC = "JDBC connection URL.";
```
ewencp: Mentioning "Optionally including username and password, or use connection.user and connection.password"? Or is this unnecessary given other fields will be visible?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -178,11 +187,13 @@
   public static ConfigDef baseConfigDef() {
     return new ConfigDef()
         .define(CONNECTION_URL_CONFIG, Type.STRING, Importance.HIGH, CONNECTION_URL_DOC, DATABASE_GROUP, 1, Width.LONG, CONNECTION_URL_DISPLAY, Arrays.asList(TABLE_WHITELIST_CONFIG, TABLE_BLACKLIST_CONFIG))
-        .define(TABLE_WHITELIST_CONFIG, Type.LIST, TABLE_WHITELIST_DEFAULT, Importance.MEDIUM, TABLE_WHITELIST_DOC, DATABASE_GROUP, 2, Width.LONG, TABLE_WHITELIST_DISPLAY,
+        .define(CONNECTION_USER_CONFIG, Type.STRING, null, Importance.MEDIUM, CONNECTION_USER_DOC, DATABASE_GROUP, 2, Width.LONG, CONNECTION_USER_DISPLAY)
```
shikhar: Oversight, will make it consistent. I think HIGH makes more sense.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
shikhar: `null` is guaranteed to be safe, `DriverManager.getConnection(url, user, password)` behaves identically to `DriverManager.getConnection(url)` when user & password are `null` (based on reading the source).

Issue #28, opened 2015-12-11
Option to include key in SourceRecord


zcox:
Currently it appears that when a new `SourceRecord` is created, the `keySchema` and `key` fields are left null, which results in the message sent to the Kafka topic also having a null key. It would be useful to have the option for the JDBC Connector to use something as the message key, e.g. the primary key id value. If compaction is enabled on this Kafka topic, then the most recent changes to each database table row are replicated in the topic and consumer may then fully rebuild the current state of the table. Otherwise, the message will be removed from the Kafka topic after the retention period passes, and the database change will be lost from the topic.

To achieve this currently, a consumer would need to extract the primary key id out of the message value, and write the key & value to a 2nd Kafka topic.

I'm sure there is complexity around which columns to use as the message key and how to configure this, but it would be nice functionality to have.

ewencp:
@zcox Agreed, this is definitely functionality we want to support! When loading from tables, I think getting the primary key shouldn't be a problem since that information should be available via JDBC. Once you have the right set of columns, converting the values is exactly the same process as it would be for converting a row. One special case we might want to handle is if the primary key is a single column then we can pull it out of the struct and return the column type directly (e.g. if it is an int, don't wrap it in a single-field struct, just return the int with Schema.INT64_SCHEMA directly).

zcox:
@ewencp that all sounds great! :+1:

wushujames:
If you just used a Schema.INT64_SCHEMA directly, then you wouldn't have a column name for the primary key, right?

It just occurred to me that kafka-connect-jdbc has no way to capture deleted rows... You'd have to do soft-deletes (aka add an isDeleted=true column to your database).

zcox:
@wushujames that also occurred to me yesterday, that the jdbc connector cannot detect deleted rows. The only workaround I came up with was what you suggested: using a boolean column to mark a row deleted.

malonej7:
:+1:  Any chance this is going to happen soon?

gwenshap:
It will happen the moment anyone submits a patch :)

davidchen322:
@gwenshap What branch should we work off of if we plan on submitting a patch?  We are currently experimenting with Confluent 2.0.1.

nbuesing:
I am working on a PR for this enhancement, I have enough code in place to where I have POC working. With this being my first contribution to this connector (and confluentinc code in general) I have a few questions.

1)

[...]

PR #166, opened 2016-11-17
Allow setting of message key

whamilton42:
Fulfils #28 

It will (naturally) only work if the key converter is set to the string converter in the source connector config. I don't know if that explicitly needs to be called out.

For testing, the existing unit tests still pass. I didn't see anywhere that I could add a more meaningful test, given the tight coupling with Kafka Connect itself. For integration testing, I have tested that the entire stack works locally (like a charm).

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
@@ -36,8 +36,8 @@ import io.confluent.connect.jdbc.util.JdbcUtils;
 public class BulkTableQuerier extends TableQuerier {
   private static final Logger log = LoggerFactory.getLogger(BulkTableQuerier.class);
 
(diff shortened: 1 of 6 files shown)


whamilton42:
[clabot:check]

shikhar:
@whamilton42 thanks for the patch! #28 is certainly a gap we'd like to address and high on the list. 

However, I think it would be good to have a more complete solution:
1) support all the primitive types - as you mentioned only strings supported here
2) support multiple PKs - Struct schema if multiple PK, primitive if single PK
3) support detecting PKs from JDBC metadata - like in #67

We should be able to devote time to getting these in for the next release. If you have time to work on fleshing out this PR before we can get to it, I'd be happy to review.

shikhar:
If https://cwiki.apache.org/confluence/display/KAFKA/KIP-66%3A+Single+Message+Transforms+for+Kafka+Connect is adopted, the `ValueToKey` single-message-transform should address the functionality implemented by this PR.

Issue #160, opened 2016-11-04
Sporadic ConnectException leads to non processing of table data


vamossagar12:
I had setup a 2 node distributed kafka-connect process. Everything went well and I could see lot of data flowing into the relevant kafka topics.
After some time, JDBCUtils.getCurrentTimeOnDB threw a ConnectException with the following stacktrace:

> The last packet successfully received from the server was 792 milliseconds ago. The last packet sent successfully to the server was 286 milliseconds ago. (io.confluent.connect.jdbc.source.JdbcSourceTask:234)
> [2016-11-02 12:42:06,116] ERROR Failed to get current time from DB using query select CURRENT_TIMESTAMP; on database MySQL (io.confluent.connect.jdbc.util.JdbcUtils:226)
> com.mysql.jdbc.exceptions.jdbc4.CommunicationsException: Communications link failure
> The last packet successfully received from the server was 1,855 milliseconds ago. The last packet sent successfully to the server was 557 milliseconds ago.
> at sun.reflect.GeneratedConstructorAccessor51.newInstance(Unknown Source)
> at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
> at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
> at com.mysql.jdbc.Util.handleNewInstance(Util.java:411)
> at com.mysql.jdbc.SQLError.createCommunicationsException(SQLError.java:1117)
> at com.mysql.jdbc.MysqlIO.send(MysqlIO.java:3829)
> at com.mysql.jdbc.MysqlIO.sendCommand(MysqlIO.java:2449)
> at com.mysql.jdbc.MysqlIO.sqlQueryDirect(MysqlIO.java:2629)
> at com.mysql.jdbc.ConnectionImpl.execSQL(ConnectionImpl.java:2719)
> at com.mysql.jdbc.PreparedStatement.executeInternal(PreparedStatement.java:2155)
> at com.mysql.jdbc.PreparedStatement.execute(PreparedStatement.java:1379)
> at com.mysql.jdbc.StatementImpl.createResultSetUsingServerFetch(StatementImpl.java:651)
> at com.mysql.jdbc.StatementImpl.executeQuery(StatementImpl.java:1527)
> at io.confluent.connect.jdbc.util.JdbcUtils.getCurrentTimeOnDB(JdbcUtils.java:220)
[...]

PR #167, opened 2016-11-17
Connection.isValid() expects seconds not milliseconds

shikhar:
Discovered in #160


shikhar:

kkonstantine:
LGTM. Would you need a test eventually for this class?

shikhar:
This class does get exercised in integration tests against SQLite and Derby, but a unit test would be a good idea. Unfortunately wouldn't have caught this issue because of the incorrect assumption.

PR #168, opened 2016-11-17
Change default MySQL Schema.Type.TIMESTAMP mapping from TIMESTAMP to DATETIME

shikhar:
DATETIME supports a wider range by default, see http://dev.mysql.com/doc/refman/5.5/en/datetime.html


shikhar:

ewencp:
@shikhar LGTM, but is this going to be a compatibility issue for users that upgrade from an older version of the connector?

shikhar:
@ewencp no, since this would only apply for `CREATE`s going forward, or `ALTER`s that add columns.

Issue #165, opened 2016-11-14
Unsigned numeric fields not correctly supported


andybryant:
I am reading from a MariaDB database with an unsigned INT field to store IP address.
`DataConverter.java` ignores whether the field is signed or not when generating the Kafka Connect Schema, always converting `Types.INTEGER` into `Schema.OPTIONAL_INT32_SCHEMA` or `Schema.INT32_SCHEMA`.

When reading from the database this results in a `resultSet.getInt` call which will fail when the value is greater than `Integer.MAX_VALUE`. Here's an example of this happening...

```
[2016-11-11 15:58:49,574] WARN Ignoring record due to SQL error: (io.confluent.connect.jdbc.JdbcSourceTask:73)
java.sql.SQLException: Out of range value for column 'customers_ip_number' : value 3707109316 is not in class java.lang.Integer range
	at org.mariadb.jdbc.internal.queryresults.resultset.MariaSelectResultSet.rangeCheck(MariaSelectResultSet.java:3005)
	at org.mariadb.jdbc.internal.queryresults.resultset.MariaSelectResultSet.getInt(MariaSelectResultSet.java:1016)
	at org.mariadb.jdbc.internal.queryresults.resultset.MariaSelectResultSet.getInt(MariaSelectResultSet.java:956)
	at io.confluent.connect.jdbc.DataConverter.convertFieldValue(DataConverter.java:299)
	at io.confluent.connect.jdbc.DataConverter.convertRecord(DataConverter.java:68)
	at io.confluent.connect.jdbc.TimestampIncrementingTableQuerier.extractRecord(TimestampIncrementingTableQuerier.java:179)
	at io.confluent.connect.jdbc.JdbcSourceTask.poll(JdbcSourceTask.java:217)
	at org.apache.kafka.connect.runtime.WorkerSourceTask.execute(WorkerSourceTask.java:155)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:140)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:175)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[...]

PR #169, opened 2016-11-20
Fixes issue #165. Widen schema types for unsigned numeric types.

andybryant:
Note Derby DB doesn't support unsigned numeric types so unfortunately unable to add test cases for them.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
@@ -97,7 +97,6 @@ public class DataConverter {
         metadata.isNullable(col) == ResultSetMetaData.columnNullableUnknown) {
       optional = true;
     }
-    boolean signed = metadata.isSigned(col);
 
     switch (sqlType) {
       case Types.NULL: {
@@ -126,13 +125,13 @@ public class DataConverter {
 
       case Types.TINYINT: {
         if (optional) {
-          if (signed) {
+          if (metadata.isSigned(col)) {
             builder.field(fieldName, Schema.OPTIONAL_INT8_SCHEMA);
           } else {
             builder.field(fieldName, Schema.OPTIONAL_INT16_SCHEMA);
           }
         } else {
-          if (signed) {
+          if (metadata.isSigned(col)) {
             builder.field(fieldName, Schema.INT8_SCHEMA);
           } else {
             builder.field(fieldName, Schema.INT16_SCHEMA);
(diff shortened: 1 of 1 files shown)


andybryant:
Signed CLA now. BTW - would be good to link to the CLA page on those warnings as took me a while to find it (and they was by accident really).

ewencp:
@andybryant Seems straightforward. The major concern I would have is with compatibility since we'll be generating a new schema. Users that may otherwise have an unsigned column with values that fall within the safe range may see their schema change.

One option is to put this behind a config, make the default maintain the current behavior, include notes in the docs about the change, and warn that we will eventually flip this to be the default.

Another would be to evaluate whether widening fields is considered a compatible change in enough serialization formats that it would be safe to include this change. It is, for example, compatible with the schema projection support in Avro, and JSON doesn't have such type information. But I'm not sure it holds for protobufs and thrift, which are probably the other two most popular formats that have type information included. @shikhar You asked about this in the original bug report thread, did you ever find out the answer?

shikhar:
For Protobufs,
> int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility. If a number is parsed from the wire which doesn't fit in the corresponding type, you will get the same effect as if you had cast the number to that type in C++ (e.g. if a 64-bit number is read as an int32, it will be truncated to 32 bits).

For Thrift, I couldn't find doc about the compatibility but it does include the type alongwith the data in the binary encoding, so it would be unnecessarily restrictive if it wasn't compatible to widen a numeric type.

So I'd be in favor of merging as-is without another config flag, so long as we note it in the changelog.

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -289,19 +325,31 @@ private static void convertFieldValue(ResultSet resultSet, int col, int colType,
...
+        } else {
+          colValue = resultSet.getInt(col);
+        }
         break;
       }
 
       // 32 bits int
       case Types.INTEGER: {
-        colValue = resultSet.getInt(col);
+        if (resultSet.getMetaData().isSigned(col)) {
```
shikhar: I think you can use the `signed` var here as well?

shikhar: Alternately it may make more sense to inline all the usages to avoid it being unnecessarily evaluated in some of the cases (e.g. if a JDBC driver throws an exception if you call `resultSet.getMetaData().isSigned(col)` for textual data)

ewencp:
@shikhar Makes sense, and sounds good to me. This info will likely be useful in other connectors as well as I'm sure this issue will pop up occasionally.

vamossagar12:
@andybryant @shikhar Would this change be integrated? I have anyways gone ahead and used the changes at my end but because of this, I need to make a lot of changes at my domain model. Basically, all my columns were `unsigned` so after including this change in my kafka-connect-jdbc codebase, whichever columns were marked as int, now need to be long. 

Since it's going to be a bit of an effort, so I thought, I would ask if this is going to be integrated or there's anything else in the pipeline :)

shikhar:
@andybryant @vamossagar12 I had a comment about only evaluating `isSigned(col)` when necessary which I think it should be addressed, and then it should be good to merge.

andybryant:
I'll update the PR with that change

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
andybryant: updated accordingly

andybryant:
Updated to only evaluate `isSigned` where necessary

PR #171, opened 2016-11-28
CC-432: Missing PreparedStatement.close()

### LINKED ISSUE #170: Unending memory rise then Out of Memory on JDBC sink
We have a JDBC sink consuming 5 topics with smallish messages (5-10, 20 character strings) with a batch size of 100. Rate of messages is 10-15 per second. They are consuming into a remote MySQL DB. The process has a heap size of 4gigs. The memory for the sink grows until it uses all memory available on the machine. Then starts throwing memory errors and other errors.

Is this a bug in the sink or is there config to handle this? I don't need real-time sinking. The sink can lag a bit. The DB is very powerful and can handle large batches.

shikhar:
Fixes #170


shikhar:

ewencp:
LGTM. I'm assuming there's not really an effective way to write a test for this since it's just internal state of `BufferedRecords`?

shikhar:
@ewencp yes, I couldn't think of an effective way to test this

PR #177, opened 2016-12-13
Handle numeric types

cotedm:
@shikhar @ewencp do you agree with the approach here to handle ``NUMERIC`` types? I **think** we can also handle ``NUMBER`` in a similar fashion. Obviously better if we can add a config flag and tests (which I will add to this PR), but I want to verify you agree with the approach. Seems to work with ``NUMERIC`` types on postgres locally for me at this time.


src/main/java/io/confluent/connect/jdbc/DataConverter.java
```diff
@@ -169,7 +169,27 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
         break;
       }
 
-      case Types.NUMERIC:
+      case Types.NUMERIC: {
+        int precision = metadata.getPrecision(col);
+        if (metadata.getScale(col) == 0 && precision < 19) { // integer
```
ewencp: So I can review this properly and validate the choice of exponents, which behavior are we going with to control the range: allowing all valid values for a given integer width but exception/clamping when a value exceeds `<IntType>.MAX_VALUE` or only converting up to a certain range (i.e. disallowing some valid values for `<IntType>` to make sure we can always convert the value we receive to the target type)?

cotedm:
Added ``NUMBER`` now and make an effort to match the scale of the number as it comes in, but required me to use a deprecated ResultSet method, which I'm not crazy about. Seems to work with manual testing against Oracle express, so I'll add some unit tests tomorrow. @ewencp or @shikhar should I add a config flag to enable this kind of data type handling or should it just be always on?

ewencp:
@cotedm We need a flag -- it's a backwards incompatible change and databases that report the right types would be affected by it.

cotedm:
@ewencp I've added a config flag and unit test. The config flag is on by default for all unit tests because it shouldn't impact them and I decided it's better to do that then break out a whole separate test. If that doesn't make sense, I'm happy to change it.

The only thing I can't see how I could unit test is the problem where Oracle returns -127 for scale in a NUMERIC type with unspecified precision. Manual testing against Oracle seems to work for me, but derby doesn't behave this way, so looks like the framework in place can't support a unit test for that?

cotedm:
Rebased and merged manually to fix tests that had changed interfaces in the constructor. Tests are passing locally.

shikhar:
LGTM

cotedm:
thanks @shikhar!

PR #188, opened 2017-01-28
add rpm build to lifecycle 

rnpridgeon:
The current process for building rpm's is kind of a pain. Especially for something small like back porting a fix. Also RPM based hot fixes are a preferred distribution method over one off jars for quick fixes.  

There is also a deb plugin which can be utilized for the same.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,10 +59,10 @@
         <commons-io.version>2.4</commons-io.version>
         <mockito.version>1.10.19</mockito.version>
         <slf4j.version>1.7.5</slf4j.version>
-	<licenses.name>Apache License 2.0</licenses.name>
+		<licenses.name>Apache License 2.0</licenses.name>
         <licenses.version>${project.version}</licenses.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-	<project.package.home>target/${project.artifactId}-${project.version}-package</project.package.home>
+		<project.package.home>target/${project.artifactId}-${project.version}-package</project.package.home>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
     </properties>
 
@@ -236,34 +236,35 @@
 						<configuration>
 							<group>Applications/Internet</group>
 							<packager>Confluent Packaging</packager>
+							<license>${licenses.name}</license>
 							<needarch>false</needarch>
 							<requires>
 								<require>confluent-common</require>
 							</requires>
 							<mappings>
 								<mapping>
-									<directory>/usr/share/java/</directory>
+									<directory>/usr/share/java/${project.artifactId}</directory>
 									<sources>
 										<source>
-											<location>${project.package.home}/share/java</location>
+											<location>${project.package.home}/share/java/${project.artifactId}</location>
 										</source>
 									</sources>
 								</mapping>
 								<mapping>
 									<configuration>true</configuration>
-									<directory>/etc</directory>
+									<directory>/etc/${project.artifactId}</directory>
 									<sources>
 										<source>
-											<location>${project.package.home}/etc/</location>
+											<location>${project.package.home}/etc/${project.artifactId}</location>
 										</source>
 									</sources>
 								</mapping>
 								<mapping>
 									<documentation>true</documentation>
-									<directory>/usr/share/doc/</directory>
+									<directory>/usr/share/doc/${project.artifactId}</directory>
 									<sources>
 										<source>
-											<location>${project.package.home}/share/doc</location>
+											<location>${project.package.home}/share/doc/${project.artifactId}</location>
 										</source>
 									</sources>
 								</mapping>
@@ -338,8 +339,8 @@
                 </plugins>
             </build>
         </profile>
-        <profile>
-            <id>licenses-source</id>
+		<profile>
+			<id>licenses-source</id>
             <build>
                 <plugins>
                     <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -240,7 +240,7 @@
 							<needarch>noarch</needarch>
 							<targetOS>linux</targetOS>
 							<requires>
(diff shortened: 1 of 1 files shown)


rnpridgeon:
[clabot:check], interesting failure. Classic case of works on my machine

rnpridgeon:
not sure if it's advisable or possible but adding the rpm command to the image remedies this. 

aptitude install rpm

rnpridgeon:
After looking into this a bit more it's probably best to add the rpm goal as a profile. This way it's not a hard requirement to have rpmbuilder in place

rnpridgeon:
@ewencp assuming this doesn't break Jenkins again do you mind reviewing. I moved the rpm plugin into its own profile so it's not run without explicitly telling it to do so with -P rpm.

pom.xml
```diff
@@ -50,7 +50,7 @@
     </scm>
 
     <properties>
-        <confluent.version>3.3.0-SNAPSHOT</confluent.version>
+        <confluent.version>${project.version}</confluent.version>
```
ewencp: @aayars This seems like a good improvement (we should avoid duplicating the version number more places than necessary), but figure it's worth checking that this won't break the version bumping tooling.

pom.xml
```diff
@@ -59,8 +59,10 @@
         <commons-io.version>2.4</commons-io.version>
         <mockito.version>1.10.19</mockito.version>
         <slf4j.version>1.7.5</slf4j.version>
-        <licenses.version>3.3.0-SNAPSHOT</licenses.version>
+	<licenses.name>Apache License 2.0</licenses.name>
```
ewencp: indentation is messed up. in general it looks like a tabs vs spaces issue

pom.xml
```diff
@@ -216,6 +217,62 @@
...
+						<configuration>
+							<group>Applications/Internet</group>
+							<packager>Confluent Packaging</packager>
+							<needarch>false</needarch>
+							<requires>
+								<require>confluent-common</require>
+							</requires>
+							<mappings>
+								<mapping>
+									<directory>/usr/share/java/</directory>
```
ewencp: Is the plugin smart about directories it includes in the RPM. iirc the RPMs (or maybe it was debs...) need to have a single owner. Unless that's just the individual files? Anyway, it'd be good to verify we can use `/usr/share/java` and don't need to use the more specific `/usr/share/java/kafka-connect-jdbc`.

pom.xml
```diff
@@ -216,6 +217,62 @@
...
+						<version>2.1.5</version>
+						<executions>
+							<execution>
+								<id>generate-rpm</id>
+								<goals>
+									<goal>rpm</goal>
+								</goals>
+							</execution>
+						</executions>
+						<configuration>
```
ewencp: * I think we've lost the package description.
* Does the license info get pulled from elsewhere in the pom automatically?
* I assume version info is picked up automatically from elsewhere in the pom too?

pom.xml
aayars: This will work fine, we <3 properties.

pom.xml
rnpridgeon: good catch, I actually had this fixed in a previous commit and forgot to pull it down when I moved the plugin to it's own profile. 

the paths should have the suffix ${project.artifactId}

after putting it back we can see that only the proper paths are included with rpm -qpl

pom.xml
rnpridgeon: picks up the description and version info from the project, forgot to bring the license when moving to the profile. This is why I can't have nice things Ewen. 

rpm -qip kafka

Name        : kafka-connect-jdbc           Relocations: (not relocatable)
Version     : 3.3.0                             Vendor: Confluent, Inc.
Release     : SNAPSHOT20170201063236        Build Date: Wed Feb  1 01:32:56 2017
Install Date: (not installed)            Build Host: ryans-mbp.attlocal.net
Group       : Applications/Internet         Source RPM: kafka-connect-jdbc-3.3.0-SNAPSHOT20170201063236.src.rpm
Size        : 6413778                          License: Apache License 2.0
Signature   : RSA/SHA1, Wed Feb  1 01:32:56 2017, Key ID e64737809b06408d
Packager    : Confluent Packaging
URL         : http://confluent.io
Summary     : kafka-connect-jdbc
Architecture: noarch
Description :
A Kafka Connect JDBC connector for copying data between databases and Kafka.

pom.xml
rnpridgeon: locally I used one tab instead of two which looks a lot different than on here. Don't really have an excuse but the latest commit should fix this

rnpridgeon:
@ewen made 2 changes aside from the ones you commented on. 1. Made the requires clause more strict. 2. set the target os so builds generated on a mac in addition to a linux box

pom.xml
ewencp: Hmm, still seems off based on what I'm seeing in this diff? `project.package.home` and the entire profile below look wonky too.

pom.xml
```diff
@@ -216,6 +217,64 @@
...
+							<group>Applications/Internet</group>
+							<packager>Confluent Packaging</packager>
+							<license>${licenses.name}</license>
+							<needarch>noarch</needarch>
+							<targetOS>linux</targetOS>
+							<requires>
+								<require>confluent-common &gt;= ${project.version}</require>
```
ewencp: Based on discussion around the top level confluent-platform RPMs allowing use of a newer bug fix release of its dependencies, do we want to switch these to be just `=`?

rnpridgeon: Thought about that but I wasn't sure how strictly we would want to adhere to that. For the top level RPM it is an issue. For the individual components not so much. In fact that is currently how you get a way with installing the exact value you want. By install each individual rpm one by one. 

I'll let you make the final verdict though because I plan on adding this to all the maven derived projects once you give it the thumbs up :)

Issue #112, opened 2016-08-22
Support WHERE clauses in custom queries in incremental mode


shikhar:
Currently `TimestampIncrementingTableQuerier` appends a`WHERE` clause which causes invalid SQL if there is already one in the SQL.

One approach could be to detect if there is already a WHERE, but that seems error prone.

I propose we allow for the query to contain placeholder text like "${incrementalClause}". If that is present that gets replaced with the timestmap/ID conditions. If it is absent, we can append a "WHERE .. conditions .. " as currently.

ewencp:
Yeah, this seems like a reasonable solution if we go with string interpolation more generally across the framework/connectors. I had avoided handling this so far because I didn't really want to try to robustly handle the SQL syntax/parsing.

teq0:
I'd be very keen for this to go ahead, it would open up a lot of possibilities that just aren't possible right now. Can I suggest that maybe a few variables could be exposed, not just the ${incrementalClause}, which would only work for straight queries. Exposing, say, ${currentIncrementValue} and ${currentTimestampValue} would allow them to be passed to stored procedures as well.

yangfeiran:
https://github.com/confluentinc/kafka-connect-jdbc/pull/191, this is my pull request for the change I made to solve your problem.

teq0:
Could you please make it case-insensitive? It only gets called once for any task, the overhead shouldn't matter.

This doesn't solve the stored procedure problem though, that requires using the templated variables discussed above. I'll have a go at it myself when I get a minute.

miketzian:
Another way would be to add a 'query.condition' config parameter, which would replace the normal where clause generated by TimestampIncrementingTableQuerier, (in this case for a stored proc):

`query.condition=@from_ts = ?, @to_ts = ?`

The developer could then write whatever where clause they need based on their use of incrementing and/or timestamp columns.

kgeis:
There are (limited) workarounds for this, depending on the level of SQL support in the underlying database. Examples:

`SELECT * FROM (SELECT... WHERE...)`

```
WITH a AS
   SELECT * FROM b
    WHERE ...
SELECT * FROM a
```

yangfeiran:
you are absolutely right, but the right syntax is " SELECT * FROM ( SELECT * FROM table WHERE ...) as a"

kgeis:
@yangfeiran, either I'm absolutely right or I'm not. No buts! :)

[...]

PR #191, opened 2017-02-13
fix "Support WHERE clauses in custom queries in incremental mode #112"

yangfeiran:
This pull can easy solve this issue(https://github.com/confluentinc/kafka-connect-jdbc/issues/112). It allows people to add "where" condition in the query part of a jdbc source. So that, people can have any kind of offset for their jdbc source instead of querying data from the beginning of a table or view.
This change has been proved effective on our production environment.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
@@ -110,7 +110,11 @@ public class TimestampIncrementingTableQuerier extends TableQuerier {
       //  timestamp 1235, id 22
       //  timestamp 1236, id 23
       // We should capture both id = 22 (an update) and id = 23 (a new row)
-      builder.append(" WHERE ");
+      if (query.indexOf("where") != -1) {
+        builder.append(" and ");
+      } else {
+        builder.append(" WHERE ");
+      }
       builder.append(JdbcUtils.quoteString(timestampColumn, quoteString));
       builder.append(" < ? AND ((");
       builder.append(JdbcUtils.quoteString(timestampColumn, quoteString));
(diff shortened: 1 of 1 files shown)


yangfeiran:
How to sign your Contributor License Agreement in order to contribute? Through a web link?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -110,7 +110,11 @@ protected void createPreparedStatement(Connection db) throws SQLException {
       //  timestamp 1235, id 22
       //  timestamp 1236, id 23
       // We should capture both id = 22 (an update) and id = 23 (a new row)
-      builder.append(" WHERE ");
+      if (query.indexOf("where") != -1) {
```
arpanchaudhury: The predicate can be query.toLowerCase().contains("where") for case insensitivity

yangfeiran: sure, cool

gwenshap:
@yangfeiran The CLA is here: http://clabot.confluent.io/cla

yangfeiran:
have signed the CLA

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
sushilborhade: If in given query there is a column name which has 'where' string then below ` if (query.indexOf("where") != -1) {` checks will failed.
e.g. column name could be **whereabouts**
Q. Whereabouts in Madrid do you live?
The 'WHERE' string should prefix and suffix by space i.e. ` if (query.indexOf(" where ") != -1) {`.

sushilborhade:
If in given query there is a column name which has 'where' string then below if (query.indexOf("where") != -1) { checks will failed.
e.g. column name could be whereabouts
Q. Whereabouts in Madrid do you live?
The 'WHERE' string should prefix and suffix by space i.e. if (query.indexOf(" where ") != -1) {.

rhauch:
@sushilborhade, great catch. Would you mind creating a separate issue for this? This would be a very simple fix, so anyone interested could providing a simple PR.

PR #199, opened 2017-03-17
new load mode

bharatviswa504:
Adding a new load mode.
Load.once -- which will help to load existing tables in database to Kafka topic only once.
As existing bulk modes repeatedly inserts table data into Kafka topics, that is not helpful.
And this mode is also helpful if we want to do one-time loading of entire database tables to Kafka topic.


bharatviswa504:
[clabot:check]

bharatviswa504:
@ewencp @gwenshap Could you please take a look into this and provide your comments?

seanlindo:
I would like to see something like this merged in. If anyone stumbles upon this, make sure you read the documentation for setting poll.interval.ms. Setting it to something high and keeping an eye on the process is a workaround for now.

ewencp:
So the problem with using an approach like this is that Kafka Connect was meant for streaming data such that connectors run indefinitely. It isn't a scheduler for finite jobs, which means it isn't designed to clean up after a job that decides to terminate itself. So while a patch like this might roughly accomplish what you're looking for, it will also leave around extra state and, in the case of any rebalancing in distributed mode it will result in still re-running the bulk load when the connector is restarted on another node.

I think for something like this to be merged there would need to be additions to the framework itself to gracefully handle jobs that terminate.

PR #202, opened 2017-03-23
HP Vertica dialect

sauliusvl:
Adding dialect for HP Vertica:

- it only supports 64bit ints/floats, which are called `INT`/`FLOAT` hence the type mapping
- not sure what's the policy on the default `VARCHAR`/`VARBINARY` lengths, vertica does not have `VARCHAR(MAX)`, so `1024` looked reasonable
- no upsert, since the `MERGE` statement does not support subqueries and is not really a "zen" thing to do in vertica -- it's a columnar store, so it does not like updates/merges
- it does not support adding multiple columns in one statement, hence issuing multiple of them

Tried out all the functionality locally, seems to work, let me know if something's missing.


sauliusvl:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/sink/dialect/VerticaDialect.java
```diff
@@ -0,0 +1,83 @@
...
+  public VerticaDialect() {
+    super("\"", "\"");
+  }
+
+  @Override
+  protected String getSqlType(String schemaName, Map<String, String> parameters, Schema.Type type) {
+    if (schemaName != null) {
+      switch (schemaName) {
+        case Decimal.LOGICAL_NAME:
+          return "DECIMAL(18," + parameters.get(Decimal.SCALE_FIELD) + ")";
```
gwenshap: Why 18?

gwenshap:
Thanks! Added a small question. 

The maximum Varchar size for Vertica seems to be 65000. However, given that we are currently returning varchar(256) for MySQL and CLOB for Oracle, your decision seems rather reasonable.

Also, our docs have few sections that discuss types and upsert behavior for the different dialects - perhaps you want to mention Vertica there?

src/main/java/io/confluent/connect/jdbc/sink/dialect/VerticaDialect.java
sauliusvl: Figured it's a reasonable default, because according to the [Vertica docs](https://my.vertica.com/docs/7.2.x/HTML/index.htm#Authoring/SQLReferenceManual/Compatibility/Oracle-bill/DataTypeMappingsBetweenVerticaAndOracle.htm): ` Vertica's NUMERIC(p, s) data type is most effective for p=18, and increasingly expensive for p=37, 58, 67, etc., where p <= 1024.`

sauliusvl:
Documented the type mappings for Vertica. Since upserts are not implemented I guess it's covered by the `Other   *not supported*` line ?

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
```diff
@@ -230,6 +230,10 @@ public static DbDialect fromConnectionString(final String url) {
       return new HanaDialect();
     }
 
+    if (url.startsWith("jdbc:vertica")) {
```
bruce-szalwinski: Need to add a test over in DbDialectTest, detectVertica().

sauliusvl: test added

ewencp:
ok to test

src/main/java/io/confluent/connect/jdbc/sink/dialect/VerticaDialect.java
ewencp: For most DBs the question here is what the range of valid values is. 18 is unusually small. Something like 38 or 65 pop up commonly (see the table that you updated in the README). I think the question here will be whether optimizing for the common case (probably fits within p=18) or covering all cases (e.g. p=67 given what databases generally cover) is the right choice.

sauliusvl:
Right, it's not really clear what's the overall objective, e.g. strings becoming `VARCHAR(256)` in mysql suggests optimizing for a common case, but then decimals getting mapped to `DECIMAL(65,s)` suggests optimizing for all cases.

So you tell me what's the preferred way, I don't really mind bumping it.

ewencp:
@sauliusvl Yeah, the problem is that no matter what we make the default someone complains :) A bunch of people also want decimal to be converted to int64 in some cases because some DBs use DECIMAL even when it isn't needed. So basically there is no winning here...

I'm fine with the current version as is, I just didn't know where 18 came from (the other numbers tend to come from upper limits in the respective databases). The long term solution is to give people control over how many digits to preserve. One thing that might be worth considering is that you can always reduce the precision with [SMTs](http://kafka.apache.org/documentation.html#connect_transforms), but you can't restore it once it has been lost. (But at the same time, supporting the most common use case as the default is worthwhile.)

sauliusvl:
@ewencp I chose 18 because Vertica [claims](https://my.vertica.com/docs/7.2.x/HTML/index.htm#Authoring/SQLReferenceManual/DataTypes/Numeric/NUMERIC.htm) to work best with it: _Performance of the NUMERIC data type has been fine tuned for the common case of 18 digits of precision._

So shall I leave it as is ? As I said, I'm fine either way :)

bruce-szalwinski:
It's outside of this PR, but I'm wondering if the JdbcSink can be more config driven and have folks provide their desired mapping in their connector config.  The specific dialect can provide reasonable defaults but also provide a mechanism for overriding.

`final DbDialect dbDialect = DbDialect.fromConnectionString(config.connectionUrl);`

Currently, a dialect is found by passing the connectionUrl config to the `fromConnectionString` method.  What if the method took the whole config object instead and then passed the config object down to the specific DbDialect?  Config options for mapping logical names to dialect specific types would be introduced as well.  

There maybe some other options that are hard coded in the DbDialect (escapeStart / escapeStop) that would be good to bubble up to the config.

ewencp:
@bruce-szalwinski I'm not sure I understand what the goal for some of those customizations would be, e.g. why would escape characters be something you'd customize for each database? I'm not opposed to the idea, though I am skeptical of doing too much via config -- the dialects are generally a good way of concisely encoding different rules for different databases. Trying to do this via config gets messy fast, and most users would just get frustrated trying to figure out the right set of seemingly magic settings that work for their database.

Perhaps you could raise this in its own issue for further discussion and clarification?

Issue #198, opened 2017-03-15
Add in retry logic for Jdbc source connector if connection cannot be acquired 


sjdurfey:
The [CachedConnectionProvider#getValidConnection](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java#L49-L62) doesn't have retry logic to attempt to get a new connection if the first attempt fails. So, as soon as it fails the task is killed and it needs to be restarted to start again. Killing the task will eventually need to happen, but ideally there is more than one attempt.

mkwhitacre:
So to better articulate the problem @sjdurfey logged this issue for...

We have a situation where on a scheduled basis the database is restarted.  (please don't just we aren't in charge of it).  This causes problems for us in that those restarting the database neglect/forget the Kafka Connect instances requiring us to have to restart the Kafka Connect instances each time as well.  

Therefore the goal would be that Kafka Connect would be resilient to polling data during those temporary downtimes.  For this reason the PR attached to this was written to try and be resilient to this situation.  The behavior should now be:

* On startup/connect if the connection isn't valid then it will fail and stop the connector from running
* JdbcSinkConnector will behave the same way as it currently does (b/c we don't need it right now)
* JdbcSourceConnector + TableMonitorThread will not kill the connector instance if it encounters a connection problem but instead keep retrying at the expected intervals with lots of log messages.  

The last bullet point should therefore help the JdbcSourceConnector tasks recover when we've validated the connection properties are good (b/c done on startup) but when the DB instance is down for an unknown period.

Thoughts?

makeittotop:
Hi, has this been already merged into master?

rhauch:
Yes, it was merged.

PR #203, opened 2017-03-25
Issue #198: Improved the JdbcSourceConnector such that if there was a…

mkwhitacre:
… valid connection on startup but invalid while running it'll be resilient to that and keep trying.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -77,7 +77,10 @@ public class JdbcSourceTask extends SourceTask {
     final String dbUrl = config.getString(JdbcSourceConnectorConfig.CONNECTION_URL_CONFIG);
     final String dbUser = config.getString(JdbcSourceConnectorConfig.CONNECTION_USER_CONFIG);
     final Password dbPassword = config.getPassword(JdbcSourceConnectorConfig.CONNECTION_PASSWORD_CONFIG);
-    cachedConnectionProvider = new CachedConnectionProvider(dbUrl, dbUser, dbPassword == null ? null : dbPassword.value());
+    final int maxConnectionAttempts = config.getInt(JdbcSourceConnectorConfig.CONNECTION_ATTEMPTS_CONFIG);
+    final long connectionRetryBackoff = config.getLong(JdbcSourceConnectorConfig.CONNECTION_BACKOFF_CONFIG);
+    cachedConnectionProvider = new CachedConnectionProvider(dbUrl, dbUser,
+      dbPassword == null ? null : dbPassword.value(), maxConnectionAttempts, connectionRetryBackoff);
 
     List<String> tables = config.getList(JdbcSourceTaskConfig.TABLES_CONFIG);
     String query = config.getString(JdbcSourceTaskConfig.QUERY_CONFIG);
--- a/src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java
@@ -16,6 +16,7 @@
 
 package io.confluent.connect.jdbc.util;
 
+import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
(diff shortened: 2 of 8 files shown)


mkwhitacre:
[clabot:check]

mkwhitacre:
Just for linking purposes, this PR is for https://github.com/confluentinc/kafka-connect-jdbc/issues/198

gwenshap:
Thanks! It makes sense to handle connection errors the same way we handle SQL errors. 

Since there are no unit tests added, can you mention how you validated this?

mkwhitacre:
@gwenshap thanks for the review.  I just updated the commit to now include some unit tests.  We'll also do some validation to make sure this didn't cause any new exceptions/NPE with the changes.

ewencp:
ok to test

src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java
```diff
@@ -53,6 +53,7 @@ public synchronized Connection getValidConnection() {
       } else if (!connection.isValid(VALIDITY_CHECK_TIMEOUT_S)) {
         log.info("The database connection is invalid. Reconnecting...");
         closeQuietly();
+        connection = null;
```
ewencp: Isn't this redundant since `closeQuietly()` already does this? If the issue is that there is an exception during the call to `connection.close()` in that method, then perhaps the `connection = null;` line should simply be moved to a `finally` clause in that method?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -221,6 +221,10 @@ public void stop() throws ConnectException {
         log.error("Failed to run query for table {}: {}", querier.toString(), e);
         resetAndRequeueHead(querier);
         return null;
+      } catch (ConnectException e) {
```
ewencp: I don't think this is safe -- won't it mask other types of errors that have nothing to do with connectivity? And do so indefinitely since we'll never pass those `ConnectExceptions` along? We do want tasks to fail eventually, we just want them to retry for some time if they can.

mkwhitacre:
@ewencp  thanks for the feedback.  If I'm understanding your feedback correctly what you are proposing is:

* Add configuration for number of attempts + time between retries to connector and ultimately the CachedConnectionProvider
* Remove the handling of ConnectionException in the JdbcSourceTask such that if thrown there it should fail the task/connector.

If a consumer wants to be resilient to a multi-hour connectivity problem they'd need to configure the number of retries + pause between retries to be greater than the possible downtime?

Are there some standard properties for setting that config we should model or just "connection.retry.attempts"/"connection.backoff.ms"?

ewencp:
Regarding the two questions:

1. Could be a config, or could just be fixed. But yeah, I think a good, default, finite, fixed set of retries should be there rather than something infinite since indefinite retries will permanently mask failures.
2. My main concern here was that ConnectException might be too broad and catch a lot more than what was intended. Now that you mentioned `ConnectionException` I am wondering if there was a typo?

Re: long outage resiliency, yes, I think that would require configuring specifically to handle this case. But I think that's probably reasonable -- in general it is better to be notified/alerted proactively if data isn't being delivered, so configuring to mask outages of many hours is probably going to be unusual; most users would want the notification after, say, 15 or 30min.

As for standard properties, nothing specific. There's a `reconnect.backoff.ms` in Kafka, but other than that, nothing closely related to model this off of. I think a # of attempts + backoff period (possibly both linear and exponential) is a good starting point.

mkwhitacre:
@ewencp Updated to support configurable retries with reasonable defaults.  Also added some basic tests for the retrying.

``` in general it is better to be notified/alerted proactively if data isn't being delivered, so configuring to mask outages of many hours is probably going to be unusual; most users would want the notification after, say, 15 or 30min.```

I agree with this statement but in my case:
* The operators of Kafka Connect are separate from the DB support so resiliency in Kafka Connect is best so it is not forgotten.  Also avoids human coordination about the restarts 
* We use metrics and logs for notification of failures versus the entire task which needs to be restarted.  This should keep the concerns orthogonal.

mkwhitacre:
@ewencp is there something else I can help with regarding validation, missing tests, or other work to streamline getting this merged?

mkwhitacre:
I'd like to help get this merged if possible or address any follow up concerns.

PR #205, opened 2017-03-28
JDBC Sink Connector - Add insert.mode = UPDATE, that generates only update queries

mariaAbramiuc:
I needed a way to just update existing records from tables, and I added this UPDATE mode for JDBC Sink Connector. This creates UPDATE queries based on the primary keys.
I think this could be useful for other as well.

Thank you,
Maria


ewencp:
ok to test

PR #207, opened 2017-03-31
correct version added for kafka

abhisheknoldus:
Correct version added for Kafka version in POM. With the previous version Kafka-connect-JDBC was not able to build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
 
     <properties>
         <confluent.version>${project.version}</confluent.version>
-        <kafka.version>0.11.0.0-SNAPSHOT</kafka.version>
+        <kafka.version>0.10.2.0</kafka.version>
         <junit.version>4.12</junit.version>
         <easymock.version>3.0</easymock.version>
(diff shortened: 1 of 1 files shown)


abhisheknoldus:
[clabot:check]

ewencp:
@abhisheknoldus 0.11.0.0-SNAPSHOT is the correct value on master for the way we track Kafka release. We develop master against the unstable, unreleased version of Kafka so we can get features developed that depend on features in the unreleased version. If you're looking for the connector as released, there is a tag for the officially released version e.g. https://github.com/confluentinc/kafka-connect-jdbc/tree/v3.2.0 and also a branch for things like docs changes that we can release after the binary artifacts have been released e.g. https://github.com/confluentinc/kafka-connect-jdbc/tree/3.2.0-post. If you're trying to build a patched version pulling in a specific bug fix, that's what you'd want to use as the base to work from and then cherry pick any fixes (or write your own).

gwenshap:
I think there's nothing to do here and we can close.

Issue #214, opened 2017-04-26
The data with same timestamp is segmented into two partitions, lead to data loss.


mayuanxiaonong:
Consider "batch.max.rows" is 100, and the data in resultset is segmented to 30,120,200 by same timestamp, for example ts1,ts2,ts3.
First time read 100 data, contains 30 with ts1 and 70 with ts2, then set offset to ts2.
Second time the query will be " `timestampColumn` > ts2 and `timestampColumn` < ts3",  the rest 50 data with ts2 will be lost.

mayuanxiaonong:
Continue until all the data in the partition which parted by latest timestamp in current batch is read

mayuanxiaonong:
Is there anyone merge the pull request ? #215

mayuanxiaonong:
Sorry, it has no problem. My mistake.

PR #215, opened 2017-04-26
Fix issue #214: The data with same timestamp is segmented into two partitions, lead to data loss.

mayuanxiaonong:
When finished reading a batch, don't stop, continue reading until all the data in the partition which parted by latest timestamp in current batch is read.


mayuanxiaonong:
Sorry, it's not a bug. My data loss was an other situation.

PR #217, opened 2017-05-03
move to common cp pom

norwood:
🎉


checkstyle/suppressions.xml
```diff
@@ -0,0 +1,39 @@
...
+          checks="(JavaNCSS)"
+          files="(DataConverter|FieldsMetadata|JdbcSourceTask|TimestampIncrementingTableQuerier).java"
+  />
+
+  <suppress
+          checks="(NPathComplexity)"
+          files="(DataConverter|FieldsMetadata|JdbcSourceTask).java"
+  />
+
+  <suppress
```
norwood: this is the only thing i was unsure of. these seem like possible bugs? maybe its intended?

src/main/java/io/confluent/connect/jdbc/sink/dialect/HanaDialect.java
```diff
@@ -50,6 +54,8 @@ protected String getSqlType(String schemaName, Map<String, String> parameters, S
           return "DATE";
         case Timestamp.LOGICAL_NAME:
           return "TIMESTAMP";
+        default:
+          log.debug("unhandled schemaName={}", schemaName);
```
norwood: pretty annoyingly had to copy/paste this in all these `*Dialect` files :(

checkstyle/suppressions.xml
ewencp: I think it's definitely possible we have a bug, but the lines listed here don't seem to make sense. The starting line corresponds to a block where we definitely have a `break` in the block (and the structure is *simpler* than earlier case statements in the file).

Are these lines just conservative? Are the errors the same even if the `lines` are set to start later?

Given the end position, I think this may have to do with this block of code: https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/source/DataConverter.java#L390-L414 The case for `Types.NUMERIC` is not guaranteed to exit the case statement since the entire block is dependent on `mapNumerics`, and if that is false it will fall through.

Perhaps this gets resolved by refactoring the fallthrough case and handling the `else` case for when `!mapNumerics` is `true`?

src/main/java/io/confluent/connect/jdbc/sink/dialect/HanaDialect.java
ewencp: Is the fix for this refactoring the interface such that each type would be individually dispatched and the generic type would handle the dispatch step + default case? If so, file a follow up to make that cleaner (as long as it doesn't affect compatibility).

kkonstantine:
This task was accomplished by https://github.com/confluentinc/kafka-connect-jdbc/pull/287 after all. 
Thanks @norwood for taking the first stab on this!
I'm closing this PR.

PR #218, opened 2017-05-09
Add a config option, to use a specified timestamp when offset is null on first startup.

mayuanxiaonong:


mayuanxiaonong:
This is a very useful config option. When deploy a new environment to read only incremental data, this config option can specify the start position to read, avoid the large scale history data.

mayuanxiaonong:
No body merge pull requests ?

PR #251, opened 2017-07-23
Enable specification of timestamp and incrementing column alias

namoscato:
This is useful when specifying a custom query and there are columns across multiple tables with the same name. Including column aliases for specificity is currently less than ideal due to the way the column names are quoted.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
@@ -58,17 +58,22 @@ public class TimestampIncrementingTableQuerier extends TableQuerier {
   private static final BigDecimal LONG_MAX_VALUE_AS_BIGDEC = new BigDecimal(Long.MAX_VALUE);
 
   private String timestampColumn;
+  private String timestampColumnAlias;
   private String incrementingColumn;
+  private String incrementingColumnAlias;
   private long timestampDelay;
   private TimestampIncrementingOffset offset;
 
   public TimestampIncrementingTableQuerier(QueryMode mode, String name, String topicPrefix,
-                                           String timestampColumn, String incrementingColumn,
+                                           String timestampColumn, String timestampColumnAlias,
+                                           String incrementingColumn, String incrementingColumnAlias,
                                            Map<String, Object> offsetMap, Long timestampDelay,
                                            String schemaPattern, boolean mapNumerics) {
     super(mode, name, topicPrefix, schemaPattern, mapNumerics);
     this.timestampColumn = timestampColumn;
+    this.timestampColumnAlias = timestampColumnAlias;
     this.incrementingColumn = incrementingColumn;
+    this.incrementingColumnAlias = incrementingColumnAlias;
     this.timestampDelay = timestampDelay;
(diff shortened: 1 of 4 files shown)


namoscato:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -249,6 +258,22 @@ private boolean isIntegralPrimitiveType(Object incrementingColumnValue) {
...
+   * @param columnName column name
+   * @param columnAlias optional column alias
+   * @param quoteString string used to quote SQL identifiers
+   * @return the quoted column with a prefixed alias, if applicable
+   */
+  private String getQuotedColumn(String columnName, String columnAlias, String quoteString) {
+    String result = "";
+
+    if (!columnAlias.isEmpty()) {
+      result += JdbcUtils.quoteString(columnAlias, quoteString) + ".";
```
ewencp: It seems like the naming of the options might be a bit misleading. From what I can tell they aren't actually the aliases. Given the construction of the column name here, isn't the thing being prefixed actually the table name? In other words, you're adding options to allow specifying the fully qualified table name, right?

If that's the case, would it be better to just keep the existing options but extend them to allow support for fully qualified column names?

namoscato: Yeah, good point – it is the _table_ alias or name as you describe.

I would not be opposed to splitting the existing column name configuration values on period (`.`), still quoting the column name and table alias/name if applicable. Is that what you were thinking?

ewencp: Yeah, exactly. Just seems simpler and more intuitive for the user. My only concern is if we then also need to deal with the user doing their own quoting since it looks like, e.g., `.` is valid for table names as long as it is quoted.

namoscato: Yeah, that would be a weird identifier, but it is certainly possible. Aside from using some obscure delimiter character, which would be less than ideal, I am not sure how we would introduce this change with the existing configurations and maintain backwards compatibility.

ewencp: I think it would be possible, but a bit tricky. We'd have to extend the config to support pre-quoted values in the config. This might not be that bad -- we can check if it is quoted by just checking the first and last chars. Then it's just a matter of being able to unquote and split it up so we can get just the table name so we can process the JDBC ResultSet.

namoscato: Yeah, I could see that working. Just to be clear, the change would essentially enable users to prevent column name configuration values from being quoted if they have already quoted it themselves?

ewencp: Yes, exactly.

namoscato: Cool, I'll take a pass at that when I get a chance and open another pull request. Thanks for the feedback!

PR #253, opened 2017-07-26
fix "Support WHERE clauses in custom queries in incremental mode #112

yangfeiran:
This pull can easy solve this issue. It allows people to add "where" condition in the query part of a jdbc sink. So that, people can have any kind of offset for their jdbc sink instead of querying data from the beginning of a table or view.


dizzy0ny:
Should you check for " where " instead of "where" to avoid issues with column names that may have 'where' in the name.

dimon222:
Maybe a bit unrelated, but
Audit test fails during "mvn package" because of missing formatting whitespaces (before functional brackers, and arguments passed to ifWhereIncluded.

rhauch:
Feel free to rebase this PR on `master` and consider [the same suggestion I made on #259](https://github.com/confluentinc/kafka-connect-jdbc/pull/259#discussion_r149996111). But it is a duplicate of #258 / #259.

joewood:
Any update on this PR?

PR #254, opened 2017-07-27
Use Postgres-compatible boolean literals for default values in sink connector

ewencp:


src/main/java/io/confluent/connect/jdbc/sink/dialect/PostgreSqlDialect.java
```diff
@@ -97,4 +97,18 @@ public void apply(StringBuilder builder, String col) {
     return builder.toString();
   }
 
+  @Override
+  protected void formatColumnValue(StringBuilder builder, String schemaName, Map<String, String> schemaParameters, Schema.Type type, Object value) {
+    if (schemaName != null) {
+      super.formatColumnValue(builder, schemaName, schemaParameters, type, value);
+    }
```
kkonstantine: Is `return;` missing here? I can't say for sure because there's no `default:` clause in super method's switch on `schemaName`. But if we don't expect to hit one of the `case`s then we might hit the buggie `case BOOLEAN` within the base `formatColumnValue`. Does it make sense that `default:` is missing?

If it doesn't I think a `return;` is missing here.

ewencp: Good catch. I need to think about the best way to solve this. This method just mutates `builder` which makes it harder to override in the subclass. There's no default case in the superclass, so `return` is only valid if it matches one of the cases in the superclass, but it's not clear what the best way to determine whether the superclass handled the particular schema name or not is.

kkonstantine: What about just copying only the `switch` on `schemaName` from the super method? Not great but then the rest would be correct in this method, right?

PR #259, opened 2017-08-04
Fixed issue ##258.

sushilborhade:
Also added check wherein incrementingColumn should not declare only with
spaces..

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
 
     <properties>
         <confluent.version>${project.version}</confluent.version>
-        <kafka.version>0.11.1.0-SNAPSHOT</kafka.version>
+        <kafka.version>0.11.0.0</kafka.version>
         <junit.version>4.12</junit.version>
         <easymock.version>3.0</easymock.version>
         <powermock.version>1.6.2</powermock.version>
--- a/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
@@ -76,7 +76,7 @@ public class TimestampIncrementingTableQuerier extends TableQuerier {
   @Override
   protected void createPreparedStatement(Connection db) throws SQLException {
     // Default when unspecified uses an autoincrementing column
-    if (incrementingColumn != null && incrementingColumn.isEmpty()) {
+    if (incrementingColumn != null && incrementingColumn.trim().isEmpty()) {
       incrementingColumn = JdbcUtils.getAutoincrementColumn(db, schemaPattern, name);
     }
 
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -51,7 +51,7 @@
 
     <properties>
         <confluent.version>${project.version}</confluent.version>
-        <kafka.version>0.11.1.0-SNAPSHOT</kafka.version>
+        <kafka.version>0.11.0.0</kafka.version>
```
rhauch: Did you accidentally include this in the PR?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -146,7 +146,17 @@ protected void createPreparedStatement(Connection db) throws SQLException {
...
 
-  @Override
+  private void appendWhereIfNotPresent(StringBuilder builder) {
+	  if(query != null && !query.trim().isEmpty()){
+		  if(query.indexOf(" where ") != -1){
+			  builder.append(" AND ");
+		  }else{
+			  builder.append(" WHERE ");
+		  }
+	  }
```
rhauch: This no longer appends ` WHERE ` for the `TABLE` case (e.g., when `query == null || query.trim().isEmpty()`).

rhauch: Perhaps a more straightforward, more efficient, and less error prone approach is to use the switch statement starting on line 87 to determine whether these criteria should be prefixed with a `WHERE` or an `AND`. Something like:

    String criteriaStart = " WHERE ";
    switch (mode) {
       case TABLE:
         builder.append("SELECT * FROM ");
         builder.append(JdbcUtils.quoteString(name, quoteString));		         
         break;
       case QUERY:
         builder.append(query);
         if (query.toUpperCase().contains(" WHERE ")) {
             criteriaStart = " AND ";
         }
         break;
       default:
         throw new ConnectException("Unknown mode encountered when preparing query: " + mode.toString());
     }
 		
and then replace the:

    builder.append(" WHERE ");

lines with

    builder.append(CRITERIA_START);

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -76,7 +76,7 @@ public TimestampIncrementingTableQuerier(QueryMode mode, String name, String top
   @Override
   protected void createPreparedStatement(Connection db) throws SQLException {
     // Default when unspecified uses an autoincrementing column
-    if (incrementingColumn != null && incrementingColumn.isEmpty()) {
+    if (incrementingColumn != null && incrementingColumn.trim().isEmpty()) {
```
rhauch: This is good.

PR #267, opened 2017-08-23
Quoting of field names can be disabled by config

breadfan:
We use MySQL and we use the JDBC connector in QUERY mode to join tables. These tables have a primary key with the same name and though we need to qualifiy that name with the table name, which does not work if all field names are quoted which is the default for this connector.

Also as a n alternative aliases can not be used with MySQL since they are not known in the WHERE clause.


breadfan:
[clabot:check]

breadfan:
Closed because of https://github.com/confluentinc/kafka-connect-jdbc/pull/269

PR #269, opened 2017-08-23
Features for MySQL: Allowing to disable field name quoting, using MySQL LIMIT to prevent OOM exceptions and allowing specification of incrementing column label

breadfan:
3 Features to make the JDBC connector more valuable for MySQL:

* Allowing to disable field name quoting
* allowing specification of incrementing column label
* using MySQL LIMIT to prevent OOM exceptions 

* Allowing to disable field name quoting:
MySQL does not allow aliases/labels for columns (i.e. 'AS alias' in SELECT) to be used in WHERE clauses. So if you want query mode and you join two tables with identical field names (e.g. 'id') you need to fully qualify the field name with the table name: table.column. But if you quote that it will not work.

* allowing specification of incrementing column label
If you need to fully qualify your incrementing column name ('table.column') then you need an additional alias ('AS alias') for the field name in the Kafka source record, because dots are not allowed.  This alias/label must be the same as in the SELECT statements 'AS alias'.

* using MySQL LIMIT to prevent OOM exceptions 
The MySQL JDBC driver per default fetches all rows. This makes it impossible to use with big tables because you will get OutOfMemoryExceptions

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -118,6 +118,15 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
   public static final String INCREMENTING_COLUMN_NAME_DEFAULT = "";
   private static final String INCREMENTING_COLUMN_NAME_DISPLAY = "Incrementing Column Name";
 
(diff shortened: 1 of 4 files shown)


vgw-rhysc:
Do these PRs intend on getting merged? This is a 6 month old PR for some pretty fundamental fixes... 😞

PR #271, opened 2017-08-23
use BYTEA for BYTES for postgresql dialect

jbeluch:


rhauch:
This PR has conflicts and hasn't been updated to work with the changes since 5.0.0, so I create a different PR (#535) with the suggested changes in the new dialects.

rhauch:
Closed in favor of #535, which works with the latest dialect design added in 5.0.0.

PR #273, opened 2017-08-24
CC-1065 Source connector’s recommender uses table type filter and caches results

rhauch:
The JDBC source connector’s recommender for table names can take a long time when the database has a large number of tables. And, the recommender is used for both the table whitelist and blacklist properties.

This PR makes a number of improvements. First, the recommender now uses the table types to reduce the number of results that must be processed when determining the names of the available tables. This may help to reduce the number of tables returned by the recommender if a different set of tables is required. For example, if only views are to be used, then the table types property can be set to `VIEW` and the recommender will correctly return only the available views.

Secondly and more importantly, the table name recommender used for both the whitelist and blacklist properties will now cache results so that the list of table names need only be performed once. To accomplish this, the class no longer uses a static recommender but instead creates a new caching recommender each time a ConfigDef instance is returned by the connector. This change alone will effectively halve the time required for the recommender to be used during configuration validation.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -25,6 +25,9 @@ import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.types.Password;
+import org.apache.kafka.common.utils.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
@@ -35,8 +38,12 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class JdbcSourceConnectorConfig extends AbstractConfig {
+
+  private static final Logger LOG = LoggerFactory.getLogger(JdbcSourceConnectorConfig.class);
 
   public static final String CONNECTION_URL_CONFIG = "connection.url";
   private static final String CONNECTION_URL_DOC = "JDBC connection URL.";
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -252,6 +258,31 @@ public boolean visible(String name, Map<String, Object> config) {
...
+   * A recommender for table names that caches the table names for the last configuration.
+   */
+  private static class CachingTableRecommender extends TableRecommender {
+
+    private Map<String, Object> lastConfig;
+    private List<Object> results;
+
+    @Override
+    public List<Object> validValues(String name, Map<String, Object> config) {
+      if (lastConfig != null && lastConfig.equals(config)) {
```
ewencp: Shouldn't this have a timeout or something as well? Even if the config doesn't change, the DB might...

rhauch: The `CachedTableRecommender` instance is per `ConfigDef` instance, and it seems like every validation attempt acquires a new `ConfigDef` instance. Is that right? If so, then each recommender instance should only be around for a short period of time and only to validate a single configuration.

rhauch: Just to clarify, the recommender is no longer static.

ewencp: The connector returns a static instance https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java#L156 https://github.com/rhauch/kafka-connect-jdbc/blob/a8dfbdb43794652dc490dd769ee68d6ddb93365e/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java#L224 so the same recommender will always be used, no?

rhauch: Okay, that makes sense. I can add a short timeout and make it threadsafe, but would it also be worthwhile to change the `JdbcSourceConnector.config()` method to not reuse the static instance? That's currently the only place the static `CONFIG_DEF` instance is currently used.

ewencp: yeah, that would work too. i think we have this pattern most places because generally the configdef is immutable

rhauch: Making the caching recommender be threadsafe and have an expiration should be enough. Just about to push a second commit that makes those changes, makes the table recommender static again (albeit it's now a caching recommender wrapper), and adds quite a few more tests. Thanks for the feedback.

rhauch:
@ewencp care to review again after my latest commits?

PR #278, opened 2017-09-07
added delete support

andybryant:
DON'T MERGE!!!


mancini77:
Hi all, I'm hoping somebody can clarify the status of this ticket - it says it is closed but the comment above warns against merging the changeset into 3.3.X ? This functionality would sure make a lot of people happy!

PR #287, opened 2017-09-15
CC-1112: Use common pom as parent and fix checkstyle issues.

ewencp:


yx91490:
It seems that io.confluent:common pom will include some unnecessary dependencies like zookeeper, why not use io.confluent:common-parent?

PR #288, opened 2017-09-18
Remove unnecessary surefire configuration overrides.

ewencp:


ewencp:
This override shouldn't be setting the version and only included a configuration that is already included in the common pom but didn't include some other configs which enable recording of coverage, so we weren't getting proper coverage reporting in Jenkins.

PR #291, opened 2017-09-26
Log out actual sql exceptions

qiao-meng-zefr:
Currently when there's error from db side, the error log shows as follow:
```
java.sql.BatchUpdateException: Batch entry 0 `INSERT INTO ....` was aborted.  Call getNextException to see the cause.
```

Instead, it should show the actual error that retrieved from `getNextException`.

This PR add that error message to log.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
@@ -71,9 +71,11 @@ public class JdbcSinkTask extends SinkTask {
       log.warn(
           "Write of {} records failed, remainingRetries={}",
           records.size(),
-          remainingRetries,
-          sqle
+          remainingRetries
       );
+      for (Throwable e: sqle) {
+        log.warn(e.getMessage());
+      }
       if (remainingRetries == 0) {
         throw new ConnectException(sqle);
       } else {


qiao-meng-zefr:
This should also help on #264

qiao-meng-zefr:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -68,6 +68,7 @@ public void put(Collection<SinkRecord> records) {
     try {
       writer.write(records);
     } catch (SQLException sqle) {
+      log.warn(sqle.getNextException().getMessage());
       log.warn(
           "Write of {} records failed, remainingRetries={}",
           records.size(),
```
rhauch: Why output this before the log message for the original? Also, it's possible that the next exception also has a next exception, so it'd be better to log all of them in the correct order _after_ the existing log message:

```
for (Throwable next : sqle) {
    log.warn(next);
}
```

WDYT?

qiao-meng-zefr: Very good point. Just updated.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -71,9 +71,11 @@ public void put(Collection<SinkRecord> records) {
...
-          remainingRetries,
-          sqle
```
qiao-meng-zefr: BTW: this will log the first exception in sqle, which is now covered in the following foreach loop. So I removed it here to prevent duplicate log.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
qiao-meng-zefr: Just updated, please take another look. Thank you.

qiao-meng-zefr:
@rhauch Could you take another look? Thank you!

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -71,9 +71,11 @@ public void put(Collection<SinkRecord> records) {
...
+          remainingRetries
       );
+      for (Throwable e: sqle) {
+        log.warn(e.getMessage());
+      }
```
rhauch: Ok, even though I suggested this approach, it now occurs to me that the separate warning messages may be difficult to reconcile if there are other concurrent log messages from other tasks, connectors, or the worker. What do you think about building a string of these messages delimited by the system line delimiter, and doing this before line 71 and include it in the one log message, something like:

```
Write of {} records failed, remaining retries={}, with {} reasons: {}
```

Thoughts?

qiao-meng-zefr: Do we have any concern on the size of this error message string? On another hand, this log entry is already separated from other log entry in this task, so we might already have this issue anyway.

What if we generate a task_id and bind it to the logger? So that all messages for this task has same task id, users can just filter by task_id.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
rhauch: This will prevent the log message from containing the stack trace for `sqle`. It'd be better to include that here and to exclude it from the list below.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
rhauch: It's possible. I don't know what cardinalities are typical. An alternative is to keep them separate log messages but to clearly denote on each log message (including the first one on line 71) info about the number of messages that are all related. So, if there are `n` exceptions, something like:

```
Write of {} records failed, remaining retries={}, with {} reason 1/{}: {}
```
and add `n` and `sqle.getMessage()`, before `sqle` in the log message parameters. The other log messages might be:

```
Write record failure reason {}/{}: {}
```
with the appropriate parameters (e.g., message #, total message count, error message).

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
qiao-meng-zefr: Yeah I'll add it back.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
qiao-meng-zefr: I tried the approach concatenating all messages together and log them all at once. It works for me. I'm sure memory usage increased a little bit, but it's not causing any obvious problem or OOM.

I'll make change accordingly.

qiao-meng-zefr:
Can't commit to this PR anymore (the forked repo was deleted). I opened a new PR #306

PR #305, opened 2017-11-06
Use PowerMock replay/verify to handle all mocks in table monitor invalid connection test

ewencp:


ewencp:
Hmm, was hoping this was the issue with jenkins build since the error `org.powermock.api.easymock.internal.invocationcontrol.EasyMockMethodInvocationControl cannot be cast to org.powermock.api.easymock.internal.invocationcontrol.EasyMockMethodInvocationControl` couldn't be reproduced locally, but it seems it is something else.

PR #306, opened 2017-11-08
Log out actual sql exceptions (fix for #291)

qiao-meng-zefr:
The original forked repo for #291 was deleted by accident so I cannot commit to that PR anymore. This PR is the final fix.

======== Original PR Note ========

Currently when there's error from db side, the error log shows as follow:
```
java.sql.BatchUpdateException: Batch entry 0 `INSERT INTO ....` was aborted.  Call getNextException to see the cause.
```

Instead, it should show the actual error that retrieved from `getNextException`.

This PR add that error message to log.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
@@ -74,14 +74,18 @@ public class JdbcSinkTask extends SinkTask {
           remainingRetries,
           sqle
       );
+      String sqleAllMessages = "";
+      for (Throwable e : sqle) {
+        sqleAllMessages += e + System.lineSeparator();
+      }
       if (remainingRetries == 0) {
-        throw new ConnectException(sqle);
+        throw new ConnectException(sqleAllMessages);
       } else {
         writer.closeQuietly();
         initWriter();
         remainingRetries--;
         context.timeout(config.retryBackoffMs);
-        throw new RetriableException(sqle);
+        throw new RetriableException(sqleAllMessages);
       }
     }
     remainingRetries = config.maxRetries;
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
@@ -79,13 +79,13 @@ public class JdbcSinkTask extends SinkTask {
         sqleAllMessages += e + System.lineSeparator();
       }
       if (remainingRetries == 0) {
-        throw new ConnectException(sqleAllMessages);
+        throw new ConnectException(new SQLException(sqleAllMessages));
       } else {
         writer.closeQuietly();
         initWriter();
         remainingRetries--;
         context.timeout(config.retryBackoffMs);
-        throw new RetriableException(sqleAllMessages);
+        throw new RetriableException(new SQLException(sqleAllMessages));
       }
     }
     remainingRetries = config.maxRetries;


qiao-meng-zefr:
@rhauch Sorry I have to open another PR because I couldn't commit to that PR anymore. Could you take a look at this PR and let me know? Thank you.

qiao-meng-zefr:
@kkonstantine Could you take a look at this PR?

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -74,14 +74,18 @@ public void put(Collection<SinkRecord> records) {
           remainingRetries,
           sqle
       );
+      String sqleAllMessages = "";
+      for (Throwable e : sqle) {
+        sqleAllMessages += e.getMessage() + "\n";
```
kkonstantine: We should probably use the `toString` method of the throwable, because if it doesn't have a cause "null" will be printed. So this would be: 

`sqleAllMessages += e + "\n";`

If `cause` is null, it'll at least print the classname of the exception.

kkonstantine: Also, I've assumed that we expect a small number of exceptions linked within the `SQLException`, that's why we concatenate the string in a loop. If that's not always the case we should use a `StringBuilder` instead.

qiao-meng-zefr: That makes sense. I'll change it.

Yes we deployed this fix to our prod system, there are usually 2, 3 exceptions linked in the SQLException. Not a lot of them.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -76,7 +76,7 @@ public void put(Collection<SinkRecord> records) {
       );
       String sqleAllMessages = "";
       for (Throwable e : sqle) {
-        sqleAllMessages += e.getMessage() + "\n";
+        sqleAllMessages += e + "\n";
```
rhauch: Might be better to use `System.lineSeparator()` rather than `\n`.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -74,14 +74,18 @@ public void put(Collection<SinkRecord> records) {
           remainingRetries,
           sqle
       );
+      String sqleAllMessages = "";
+      for (Throwable e : sqle) {
+        sqleAllMessages += e + "\n";
```
rhauch: Might be better to use `System.lineSeparator()` rather than `\n`.

qiao-meng-zefr: Great. Thanks for pointing out. Done!

rhauch:
@kkonstantine, yes those errors were fixed on master several weeks ago. @qiao-meng-zefr, would you mind rebasing your PR on the latest `master`, just to be sure? Then we can wait for a clean build and merge.

Everything else looks great. Thanks, @qiao-meng-zefr!

qiao-meng-zefr:
Just rebased from `master`. Looks like still failing..

rhauch:
Did you merge or rebase? Rebasing should not have brought in all of those extra commits.

qiao-meng-zefr:
I used rebase. Here's what I did:

```
git remote add origin-master git@github.com:confluentinc/kafka-connect-jdbc.git
git fetch origin-master
git checkout log_all_msgs
git rebase origin-master/master
git pull
git push
```

Is this the right way to do it?

qiao-meng-zefr:
I changed the base branch to another and then back to master on github. That fixed the extra commits from others..

rhauch:
No, I would have done the following:

```
git remote add upstream git@github.com:confluentinc/kafka-connect-jdbc.git
git checkout master
git pull upstream master
git checkout log_all_msgs
git rebase master
git push
```
Note that I used `upstream` rather than `origin-master` simply because I think it's difficult to have a remote name that includes the name of the branch.

qiao-meng-zefr:
I think I figured it out.. after rebase my local will have conflicts with github. I made a mistake doing `git pull`, I should just have done `git push -f`. That'll keep a clean rebase history.

So now I cleaned up the history on this PR and did another push. It looks good now.

However the test still fails. Could you guys take another look? thank you for the help!

rhauch:
The build failure is a JDBC sink connector test, so its unrelated:

    java.lang.NullPointerException
        at io.confluent.connect.jdbc.sink.JdbcSinkTaskTest.retries(JdbcSinkTaskTest.java:239)

rhauch:
@qiao-meng-zefr, this is the test method that is failing: [JdbcSinkTest.retries()](https://github.com/qiao-meng-zefr/kafka-connect-jdbc/blob/f7c722e5d1e5933fbe10887f955535178c16c2f3/src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkTaskTest.java#L188) is throwing a NPE on [this line](https://github.com/qiao-meng-zefr/kafka-connect-jdbc/blob/f7c722e5d1e5933fbe10887f955535178c16c2f3/src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkTaskTest.java#L239) since it expects a non-null cause.

You should be able to run this locally with your changes, with `mvn clean install`.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -74,14 +74,18 @@ public void put(Collection<SinkRecord> records) {
           remainingRetries,
           sqle
       );
+      String sqleAllMessages = "";
+      for (Throwable e : sqle) {
+        sqleAllMessages += e + System.lineSeparator();
+      }
       if (remainingRetries == 0) {
-        throw new ConnectException(sqle);
+        throw new ConnectException(SQLException(sqleAllMessages));
```
kkonstantine: Not sure if we didn't notice before, or this change happened along with rebasing, but this leads to compilation error: 
SQLException needs construction with `new`

qiao-meng-zefr: Yeah that was me trying to fix the unit test. Forgot the new keyword. Fixed now. And looks like tests passed.

kkonstantine:
Rebased on top of recent changes. Thanks for the fix @qiao-meng-zefr !

PR #307, opened 2017-11-10
CC-1266: Corrected build by changing PowerMock uses

rhauch:
A [recent change to the Kafka dependencies](https://github.com/apache/kafka/pull/3881) included an upgrade of the PowerMock library to 2.0.0-beta5 for JDK9 reasons, but this appears to have broken several tests in this JDBC connector repository.

Apparently, with the new 2.0.0-beta5 version it is better to use `@Mock` annotations on member fields rather than to manually create the (local or member field) mock instances via `PowerMock.createMock(MyClass.class)`. Unfortunately this requires running the concrete test classes via `@RunWith(PowerMockRunner.class)` and several other class-level annotations, which is more verbose.

### DIFF
--- a/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfigTest.java
+++ b/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfigTest.java
@@ -24,7 +24,12 @@ import org.easymock.EasyMock;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.powermock.api.easymock.PowerMock;
+import org.powermock.api.easymock.annotation.Mock;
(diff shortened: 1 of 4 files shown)


rhauch:
Note that although the PowerMock dependency was changed before the Kafka 1.0.0 release (e.g., it is changed on Kafka `1.0.x` and `trunk`), the Common POM overrides the version and includes PowerMock 2.0.0-beta5 _only on the `master` branch, which is why this change only needs to be applied to `master`.

rhauch:
@ewencp thanks for the quick review. I'm not sure why there is a difference, but the previous code definitely failed every time. My guess is that PowerMock 2.0.0-beta5 is not yet stable and correct.

ijuma:
The PowerMock author told me to use the annotations in the AK PR you linked to, so my understanding is that it's the recommended approach.

Issue #211, opened 2017-04-19
SQL Syntax error when current time is retrieved on DB2 UDB for AS/400


Protoss78:
The error happens when using mode=timestamp+incrementing or mode=timestamp on a connected DB2 UDB for AS/400 database.
The exception is:
`[2017-04-19 16:09:11,379] DEBUG executing query select CURRENT_TIMESTAMP; to get current time from database (io.confluent.connect.jdbc.util.JdbcUtils:198)
[2017-04-19 16:09:11,388] ERROR Failed to get current time from DB using query select CURRENT_TIMESTAMP; on database DB2 UDB for AS/400 (io.confluent.connect.jdbc.util.JdbcUtils:205)
com.ibm.as400.access.AS400JDBCSQLSyntaxErrorException: [SQL0104] Token ; ungültig. Gültige Token: + - AS <IDENTIFIER>.
	at com.ibm.as400.access.JDError.createSQLExceptionSubClass(JDError.java:891)
	at com.ibm.as400.access.JDError.throwSQLException(JDError.java:692)
	at com.ibm.as400.access.AS400JDBCStatement.commonPrepare(AS400JDBCStatement.java:1836)
	at com.ibm.as400.access.AS400JDBCStatement.executeQuery(AS400JDBCStatement.java:2459)
	at io.confluent.connect.jdbc.util.JdbcUtils.getCurrentTimeOnDB(JdbcUtils.java:199)
	at io.confluent.connect.jdbc.source.TimestampIncrementingTableQuerier.executeQuery(TimestampIncrementingTableQuerier.java:154)
	at io.confluent.connect.jdbc.source.TableQuerier.maybeStartQuery(TableQuerier.java:84)
	at io.confluent.connect.jdbc.source.TimestampIncrementingTableQuerier.maybeStartQuery(TimestampIncrementingTableQuerier.java:55)
	at io.confluent.connect.jdbc.source.JdbcSourceTask.poll(JdbcSourceTask.java:200)
	at org.apache.kafka.connect.runtime.WorkerSourceTask.execute(WorkerSourceTask.java:162)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:139)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:182)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #308, opened 2017-11-11
Broader fix for issue #211

k1th:
Target all product names starting with "DB2" regardless of platform.


src/main/java/io/confluent/connect/jdbc/util/JdbcUtils.java
```diff
@@ -243,7 +243,7 @@ public static Timestamp getCurrentTimeOnDB(
     String dbProduct = conn.getMetaData().getDatabaseProductName();
     if ("Oracle".equals(dbProduct)) {
       query = "select CURRENT_TIMESTAMP from dual";
-    } else if ("Apache Derby".equals(dbProduct) || "DB2 UDB for AS/400".equals(dbProduct)) {
+    } else if ("Apache Derby".equals(dbProduct) || dbProduct.startsWith("DB2")) {
```
kkonstantine: Should we be super cautious and add: 
`(dbProduct != null && dbProduct.startsWith("DB2"))` instead?

A matter of style more than anything else, rather than a possibility of an implementation of `getDatabaseProductName` ever returning `null`

k1th:
While where at it:
This probably won't make it into 4.0.0 confluent, will it?

In order to use the docker image for connect distributed, will we need to create our own patched version of the image or is there a way to overwrite the provided jdbc source. I tried and couldn't find a way to do that…

miguno:
Thanks @k1th for re-submitting this PR.  Sorry that our auto-workflow closed the original one of yours.

k1th:
what's up with the jenkins build? Can't see details of the error due to server not answering…

kkonstantine:
It was just checkstyle errors. I fixed them, but when I pushed to your branch, this PR was closed. 

Reopened with your changes for one last time here: https://github.com/confluentinc/kafka-connect-jdbc/pull/319

Indeed, it's too late for 4.0.0, it will appear in the next bug fix for 4.0 (that is 4.0.1) and any forthcoming new releases. Merging the PR is pending the release of 4.0.0.

PR #316, opened 2017-11-21
CC-45: Added support for multiple timestamp columns

mikkin:
With this PR, users can specify multiple timestamp columns to address the scenario where they may have "modified" and "created" timestamp columns. They will need to list the column names in the order of priority. 
For e.g.
modified, created - will look at modified column first and only if that column is null, it will proceed to created

There are checks in place to ensure that **_at least_** one of the timestamp column is not null.

Finally, added tests to validate the changes made.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -360,7 +360,7 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
         MODE_DEPENDENTS_RECOMMENDER
     ).define(
         TIMESTAMP_COLUMN_NAME_CONFIG,
-        Type.STRING,
+        Type.LIST,
         TIMESTAMP_COLUMN_NAME_DEFAULT,
         Importance.MEDIUM,
         TIMESTAMP_COLUMN_NAME_DOC,
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -115,8 +115,8 @@ public class JdbcSourceTask extends SourceTask {
         = config.getString(JdbcSourceTaskConfig.SCHEMA_PATTERN_CONFIG);
     String incrementingColumn
         = config.getString(JdbcSourceTaskConfig.INCREMENTING_COLUMN_NAME_CONFIG);
-    String timestampColumn
-        = config.getString(JdbcSourceTaskConfig.TIMESTAMP_COLUMN_NAME_CONFIG);
+    List<String> timestampColumn
+        = config.getList(JdbcSourceTaskConfig.TIMESTAMP_COLUMN_NAME_CONFIG);
     Long timestampDelayInterval
         = config.getLong(JdbcSourceTaskConfig.TIMESTAMP_DELAY_INTERVAL_MS_CONFIG);
     boolean validateNonNulls
@@ -151,6 +151,12 @@ public class JdbcSourceTask extends SourceTask {
       boolean mapNumerics
           = config.getBoolean(JdbcSourceTaskConfig.NUMERIC_PRECISION_MAPPING_CONFIG);
 
+      TimestampHelper timestampHelper = new MultiColumnTimestampHelper(timestampColumn);
+
+      if (!mode.equals(JdbcSourceTaskConfig.MODE_BULK) && timestampColumn.size() == 1) {
+        timestampHelper = new SingleColumnTimestampHelper(timestampColumn.get(0));
+      }
+
       if (mode.equals(JdbcSourceTaskConfig.MODE_BULK)) {
         tableQueue.add(new BulkTableQuerier(queryMode, tableOrQuery, schemaPattern,
                 topicPrefix, mapNumerics));
@@ -160,11 +166,11 @@ public class JdbcSourceTask extends SourceTask {
                 timestampDelayInterval, schemaPattern, mapNumerics));
       } else if (mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP)) {
         tableQueue.add(new TimestampIncrementingTableQuerier(
-            queryMode, tableOrQuery, topicPrefix, timestampColumn, null, offset,
+            queryMode, tableOrQuery, topicPrefix, timestampHelper, null, offset,
                 timestampDelayInterval, schemaPattern, mapNumerics));
       } else if (mode.endsWith(JdbcSourceTaskConfig.MODE_TIMESTAMP_INCREMENTING)) {
         tableQueue.add(new TimestampIncrementingTableQuerier(
-            queryMode, tableOrQuery, topicPrefix, timestampColumn, incrementingColumn,
+            queryMode, tableOrQuery, topicPrefix, timestampHelper, incrementingColumn,
                 offset, timestampDelayInterval, schemaPattern, mapNumerics));
       }
     }
@@ -267,7 +273,7 @@ public class JdbcSourceTask extends SourceTask {
       String schemaPattern,
       String table,
       String incrementingColumn,
-      String timestampColumn
+      List<String> timestampColumn
   ) {
     try {
       final Connection connection = cachedConnectionProvider.getValidConnection();
@@ -283,7 +289,7 @@ public class JdbcSourceTask extends SourceTask {
       }
       if ((incrementalMode.equals(JdbcSourceConnectorConfig.MODE_TIMESTAMP)
            || incrementalMode.equals(JdbcSourceConnectorConfig.MODE_TIMESTAMP_INCREMENTING))
-          && JdbcUtils.isColumnNullable(connection, schemaPattern, table, timestampColumn)) {
+          && JdbcUtils.areAllColumnsNullable(connection, schemaPattern, table, timestampColumn)) {
         throw new ConnectException("Cannot make incremental queries using timestamp column "
                                    + timestampColumn + " on " + table + " because this column is "
                                    + "nullable.");
--- a/src/main/java/io/confluent/connect/jdbc/source/MultiColumnTimestampHelper.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/MultiColumnTimestampHelper.java
@@ -1,0 +1,69 @@
+package io.confluent.connect.jdbc.source;
+
+import io.confluent.connect.jdbc.util.JdbcUtils;
+import io.confluent.connect.jdbc.util.StringUtils;
+import java.sql.Timestamp;
+import java.util.List;
+import org.apache.kafka.connect.data.Struct;
+
+public class MultiColumnTimestampHelper implements TimestampHelper {
+
+  private List<String> timestampColumns;
+
+  public MultiColumnTimestampHelper(List<String> timestampColumns) {
+    this.timestampColumns = timestampColumns;
+  }
+
+  @Override
+  public void incrementingWhereClauseBuilder(StringBuilder builder,
+      String incrementingColumn, String quoteString) {
+    List<String> quotedList = JdbcUtils.quoteList(timestampColumns, quoteString);
+    String quotedTimestampColumn = StringUtils.join(quotedList, ",");
+    builder.append(" WHERE COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") < ? AND ((COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") = ? AND ");
+    builder.append(JdbcUtils.quoteString(incrementingColumn, quoteString));
+    builder.append(" > ?");
+    builder.append(") OR COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") > ?)");
+    builder.append(" ORDER BY COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append("),");
+    builder.append(JdbcUtils.quoteString(incrementingColumn, quoteString));
+    builder.append(" ASC");
+  }
+
+  @Override
+  public void buildWhereClause(StringBuilder builder, String quoteString) {
+    List<String> quotedList = JdbcUtils.quoteList(timestampColumns, quoteString);
+    String quotedTimestampColumn = StringUtils.join(quotedList, ",");
+    builder.append(" WHERE COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") > ? AND COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") < ? ORDER BY COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") ASC");
+  }
+
+  @Override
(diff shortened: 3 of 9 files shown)


src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -289,12 +270,11 @@ private boolean isIntegralPrimitiveType(Object incrementingColumnValue) {
...
   public String toString() {
-    return "TimestampIncrementingTableQuerier{"
-           + "name='" + name + '\''
-           + ", query='" + query + '\''
-           + ", topicPrefix='" + topicPrefix + '\''
-           + ", timestampColumn='" + timestampColumn + '\''
-           + ", incrementingColumn='" + incrementingColumn + '\''
-           + '}';
+    return "TimestampIncrementingTableQuerier{" +
+        "incrementingColumn='" + incrementingColumn + '\'' +
```
kkonstantine: I was expecting checkstyle to complain about `+` not being at the start of the line. Checking at the jenkins build I see it does indeed (along with a few more errors).

src/main/java/io/confluent/connect/jdbc/util/JdbcUtils.java
```diff
@@ -202,6 +202,20 @@ public static boolean isColumnNullable(
     return false;
   }
```
kkonstantine: Looks like the other public methods in this class have a javadoc. Should we add one here too?

src/main/java/io/confluent/connect/jdbc/util/JdbcUtils.java
```diff
@@ -224,6 +238,14 @@ public static String quoteString(String orig, String quote) {
     return quote + orig + quote;
   }
```
kkonstantine: Should we include javadoc?

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -183,11 +182,38 @@ public void testTimestamp() throws Exception {
...
+    // Manage these manually so we can verify the emitted values
+    db.createTable(SINGLE_TABLE_NAME,
+        "modified", "TIMESTAMP",
+        "created", "TIMESTAMP NOT NULL",
+        "id", "INT");
+    db.insert(SINGLE_TABLE_NAME, "created", DateTimeUtils.formatUtcTimestamp(new Timestamp(10L)), "id", 1);
+    startTask("modified, created", null, null);
+    verifyMultiTimestampFirstPoll(TOPIC_PREFIX + SINGLE_TABLE_NAME);
+
+
```
kkonstantine: extra line

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -183,11 +182,38 @@ public void testTimestamp() throws Exception {
...
+    db.insert(SINGLE_TABLE_NAME, "modified", DateTimeUtils.formatUtcTimestamp(new Timestamp(13L)), "created", DateTimeUtils.formatUtcTimestamp(new Timestamp(10L)), "id", 2);
+    db.insert(SINGLE_TABLE_NAME, "created", DateTimeUtils.formatUtcTimestamp(new Timestamp(11L)), "id", 3);
+    db.insert(SINGLE_TABLE_NAME, "created", DateTimeUtils.formatUtcTimestamp(new Timestamp(12L)), "id", 4);
+
+    verifyPoll(3, "id", Arrays.asList(2, 3, 4), false, false, true, TOPIC_PREFIX + SINGLE_TABLE_NAME);
+
+    PowerMock.verifyAll();
+
```
kkonstantine: Few extra lines.

src/main/java/io/confluent/connect/jdbc/source/MultiColumnTimestampHelper.java
```diff
@@ -0,0 +1,69 @@
+package io.confluent.connect.jdbc.source;
+
+import io.confluent.connect.jdbc.util.JdbcUtils;
+import io.confluent.connect.jdbc.util.StringUtils;
+import java.sql.Timestamp;
+import java.util.List;
+import org.apache.kafka.connect.data.Struct;
+
```
kkonstantine: Class javadoc?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -151,6 +151,12 @@ public void start(Map<String, String> properties) {
       boolean mapNumerics
           = config.getBoolean(JdbcSourceTaskConfig.NUMERIC_PRECISION_MAPPING_CONFIG);
 
+      TimestampHelper timestampHelper = new MultiColumnTimestampHelper(timestampColumn);
```
kkonstantine: Should we just set to `null` initially followed by an if-else? Maybe the instantiation here is not worth it, given we might discard it below.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -267,7 +273,7 @@ private void validateNonNullable(
       String schemaPattern,
       String table,
       String incrementingColumn,
-      String timestampColumn
+      List<String> timestampColumn
```
wicknicks: should this variable be called `timestampColumns`?

src/main/java/io/confluent/connect/jdbc/source/TimestampHelper.java
```diff
@@ -0,0 +1,34 @@
...
+import java.sql.Timestamp;
+import org.apache.kafka.connect.data.Struct;
+
+/**
+ * TimestampHelper is an interface to construct and operate on timestamp based queries. The
+ * implementation handles two types of queries: single timestamp or multiple timestamps.
+ */
+public interface TimestampHelper {
+
+  public void addWhereClause(StringBuilder builder, String quoteString, String incrementingColumn);
```
rhauch: This is pretty minor, but it'd be nice to have JavaDoc for these methods to be clear about expectations and constraints, such as whether `incrementingColumn` (or any of the other parameters) might ever be null.

mikkin: Makes sense. I meant to add that and then slipped through.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -151,21 +151,28 @@ public void start(Map<String, String> properties) {
       boolean mapNumerics
           = config.getBoolean(JdbcSourceTaskConfig.NUMERIC_PRECISION_MAPPING_CONFIG);
 
+      final TimestampHelper timestampHelper;
+      if (timestampColumns.size() == 1) {
```
kkonstantine: Empty list is allowed to be returned by the `ConfigDef` if the `String` representation of the list is empty. What should we do if users have supplied empty list of `timestampColumns`?`

rhauch: Yeah, that's a bit confusing. It is an option a parameter, but the helper is only used if the mode is timestamp or timestamp-incrementing. Are we missing a validator to make sure that at least one timestamp column is provided when the mode is timestamp or timestamp-incrementing?

src/main/java/io/confluent/connect/jdbc/source/MultiColumnTimestampHelper.java
```diff
@@ -0,0 +1,90 @@
...
+    builder.append(" ASC");
+  }
+
+  @Override
+  public void addWhereClause(StringBuilder builder, String quoteString) {
+    List<String> quotedList = JdbcUtils.quoteList(timestampColumns, quoteString);
+    String quotedTimestampColumn = StringUtils.join(quotedList, ",");
+    builder.append(" WHERE COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") > ? AND COALESCE(");
```
wicknicks: COALESCE returns the first non null timestamp in this case. Wouldn't we need the largest timestamp for every row? In the (created-date, modified-date) example, COALESCE could return created-date for every query, while the modified-date is the column which is being updated.

rhauch: We have to be able to pull out the timestamp value that satisfied the criteria, and if there are multiple timestamps there's no guarantee we know which one actually did. So the current solution is to have the column names be ordered and to grab the first non-null timestamp value, and this corresponds to the `COALESCE` function.

So, a user whose table contains multiple timestamp columns might need to make sure they order the timestamp columns with the preferred one first.

wicknicks: Got it. Should we clear up the documentation to reflect this more clearly?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -120,8 +120,8 @@
 
   public static final String TIMESTAMP_COLUMN_NAME_CONFIG = "timestamp.column.name";
   private static final String TIMESTAMP_COLUMN_NAME_DOC =
-      "The name of the timestamp column to use to detect new or modified rows. This column may "
-      + "not be nullable.";
+      "Comma separated list of one or more timestamp columns to detect new or modified rows in "
+      + "the order specified. At least one column may not be nullable.";
```
rhauch: To address @wicknicks' suggestion, we might change this to something like:

> Comma separated list of one or more timestamp columns to detect new or modified rows. The order is important: the first non-null timestamp found in each row will be used. At least one column should not be nullable.

src/main/java/io/confluent/connect/jdbc/source/MultiColumnTimestampHelper.java
rhauch: +1. I'll add a suggestion in a line comment.

mikkin: I do feel that @wicknicks has a good point here in that this does make it error prone and easy for someone to make a mistake.

rhauch: @mikkin we need to extract the same timestamp value from the record that satisfied the criteria. If we use `COALESCE` in the criteria, then that will select a row if and only if the first non-null column value is greater than the timestamp we provide when querying. That means that when we extract the value, we have to pull the first non-null timestamp column value from the record.

However, if we were to instead do something a bit different and build the criteria with `(x IS NOT NULL AND x > ?)` ORed together for each timestamp column, then we would have to do the same thing to extract the timestamp from the record: loop through the timestamp column values and look for the first non-null value _that is greater than the timestamp used in the criteria_.

Using `COALESCE` is nice because we still only have a single `?` parameter in our prepared statement, whereas using the `(x IS NOT NULL AND x > ?)` criteria for each timestamp column means we have one `?` parameter _for each timestamp column_, and that will change how we set the values on the prepared statement just before we execute it. 

I think for simplicity we should stick with `COALESCE` because it's easier to understand and explain, and because the implementation is simpler. That means that the current logic is correct, although we still need to improve the documentation to make the behavior clear.

rhauch: Also, `COALESCE` is used in the `ORDER BY` clause to ensure the rows are ordered correctly, so that's another reason to use `COALESCE`.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
rhauch: Or

> Comma separated list of one or more timestamp columns to detect new or modified rows using the `COALESCE` SQL function. Rows whose first non-null timestamp value is greater than the largest previous timestamp value seen will be discovered with each poll. At least one column should not be nullable.

wicknicks: Looks like we are assuming that there exists an order to how the timestamps are updated. Could there be cases where this is not true? For example, in an `employee` table we could have columns like `created_date`, `name_update_date` and `address_update_date`, then we might not be able to prescribe an order suitable to the `COALESCE` function.

rhauch: What would a query against that table look like, especially the ordering clause?

src/main/java/io/confluent/connect/jdbc/source/MultiColumnTimestampHelper.java
wicknicks: I see. We want the DB to order by timestamps. Fair enough. Any reason why we are not using this `GREATEST` function?  (https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions060.htm)

mikkin: `GREATEST` returns `NULL` if any value is null.

wicknicks: You can just protect it with `GREATEST ( COALESCE (date1, 0) )` (https://stackoverflow.com/a/9831901)

rhauch: We should really use `ORDER BY ASC NULLS LAST`, but not DBMSes support that `NULLS LAST|FIRST` clause.

From what I understand, only Oracle and MySQL have `GREATEST` function.

mikkin:
@wicknicks, @kkonstantine, @rhauch please review

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
wicknicks: We leave this as `timestampColumn`?

src/main/java/io/confluent/connect/jdbc/source/MultiColumnTimestampHelper.java
```diff
@@ -0,0 +1,90 @@
...
+    builder.append(JdbcUtils.quoteString(incrementingColumn, quoteString));
+    builder.append(" > ?");
+    builder.append(") OR COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append(") > ?)");
+    builder.append(" ORDER BY COALESCE(");
+    builder.append(quotedTimestampColumn);
+    builder.append("),");
+    builder.append(JdbcUtils.quoteString(incrementingColumn, quoteString));
+    builder.append(" ASC");
```
wicknicks: This is more FYI.. But if we use the String concatenation operator (`+`), then the java compiler will optimize it to use StringBuilder. (ref: https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.18.1).

wicknicks:
Thanks for the PR, LGTM.

mikeakala:
Is there a plan to release this feature?

PR #317, opened 2017-11-22
Pull fixed tableNamePattern to configuration

mrcmatuszak:
`TableMonitorThread` calling the `getTables` method is causing dead lock on our database.
The instance of DB contains thousands o tables in one schema with many DML operations. 
Once getTables is called the MS JDBC driver calls sp_tables which ends in deadlock.

This pull request pulls `TableNamePattern` (which was originally flexed to `'%'`) to configuration to narrow the query.


mrcmatuszak:
[clabot:check]

andrewegel:
Causing CI to break due to the "unknown repository" - This needs to be closed to get around that. If this change is still required, please re-do it.

PR #331, opened 2017-12-29
add a note about SSL with JDBC

cotedm:
This is to update the docs to explain SSL config a bit more for this connector.


docs/sink_config_options.rst
```diff
@@ -3,6 +3,19 @@
...
 
+Security
+^^^^^^^^
+
+In the connector configuration you will notice there are no security parameters. This is because
+SSL is not part of the JDBC standard and will depend on the JDBC driver in use. In general, you
+will need to configure SSL via the ``connection.url`` parameter. For example, with mysql it would
+look like:
+
+  ..sourcecode bash:
```
jcustenborder: You can use properties instead of bash for the highlighting if you want. LGTM

jcustenborder:
@cotedm One small thing but LGTM.

docs/sink_config_options.rst
```diff
@@ -3,6 +3,19 @@
 JDBC Sink Configuration Options
 -------------------------------
 
+Security
+^^^^^^^^
+
+In the connector configuration you will notice there are no security parameters. This is because
+SSL is not part of the JDBC standard and will depend on the JDBC driver in use. In general, you
+will need to configure SSL via the ``connection.url`` parameter. For example, with mysql it would
```
ewencp: nit: mysql should either be monotype font to indicate protocol in the connection url or MySQL to match the product name

docs/sink_config_options.rst
```diff
@@ -3,6 +3,19 @@
 JDBC Sink Configuration Options
 -------------------------------
 
+Security
```
ewencp: nit: maybe emphasize which connection this security refers to? Kafka security is via worker config, but new users may not always realize that. maybe "Database Connection Security" or something like that?

docs/source_config_options.rst
```diff
@@ -1,6 +1,19 @@
 JDBC Source Configuration Options
 ---------------------------------
 
+Security
+^^^^^^^^
+
+In the connector configuration you will notice there are no security parameters. This is because
```
ewencp: if you're going to repeat text, it is possible to include the same text via an [include](http://docutils.sourceforge.net/docs/ref/rst/directives.html#include). @joel-hamill recently started using these to make repeated text more manageable. seems reasonable to do here, and he might have more guidance on how it makes best sense to manage this going forward.

cotedm: yeah this makes sense, I gave the ``include`` directive a shot in the latest commit. I can already see just in making a couple of changes how mistake prone this repeated text would be if copied in multiple places. If it looks ok to @joel-hamill then I'll merge.

cotedm:
@joel-hamill I added you as a reviewer to make sure I'm using this ``include`` directive like you would expect for consistency.

docs/db_connection_security.rst
```diff
@@ -0,0 +1,12 @@
+Database Connection Security
```
joel-hamill: @cotedm for the include, the convention is to nest in an `/includes` folder (e.g. `connect-jdbc/includes/db_connection_security.rst`). also, not an issue here, but eventually if you have more than one string to include, you can do this by referencing the specific line numbers for your include (see https://github.com/confluentinc/docs#content-reuse).

joel-hamill: i've added more explanation here as well https://github.com/confluentinc/docs/pull/650

cotedm: thanks @joel-hamill! how's this look? If it's good, I think we're ok to merge :)

docs/includes/db_connection_security.rst
```diff
@@ -0,0 +1,12 @@
+Database Connection Security
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In the connector configuration you will notice there are no security parameters. This is because
+SSL is not part of the JDBC standard and will depend on the JDBC driver in use. In general, you
+will need to configure SSL via the ``connection.url`` parameter. For example, with MySQL it would
+look like:
+
+  ..sourcecode properties:
```
ewencp: this should be `.. sourceode:: properties`. right now it is rendering weirdly because it's not being interpreted as an rst command

joel-hamill: yeah, the output is messed up 
image

ewencp:
@joel-hamill I pushed a fix for the formatting issue if you want to verify. Then I think we're good to merge.

PR #333, opened 2018-01-03
CC-1193: Refactored with a new dialect framework for DBMS-specific behavior

### LINKED ISSUE #35: support postgresql boolean/BIT
I know you are already aware of this, as it's mentioned in the [code](https://github.com/confluentinc/kafka-connect-jdbc/blob/49530e16ee3f5bdd905c5073bb3a102f59899fbc/src/main/java/io/confluent/connect/jdbc/DataConverter.java#L278) but it'd be nice to handle the BIT type for postgresql.

Indeed, boolean columns get seen as BIT types and fail while being converted:

```
[2016-01-08 16:09:48,292] WARN Ignoring record due to SQL error: (io.confluent.connect.jdbc.JdbcSourceTask:73)
org.postgresql.util.PSQLException: Bad value for type byte : f
    at org.postgresql.jdbc2.AbstractJdbc2ResultSet.getByte(AbstractJdbc2ResultSet.java:2100)
    at io.confluent.connect.jdbc.DataConverter.convertFieldValue(DataConverter.java:281)
    at io.confluent.connect.jdbc.DataConverter.convertRecord(DataConverter.java:68)
    at io.confluent.connect.jdbc.TimestampIncrementingTableQuerier.extractRecord(TimestampIncrementingTableQuerier.java:165)
    at io.confluent.connect.jdbc.JdbcSourceTask.poll(JdbcSourceTask.java:211)
name=control-info-pgstaging
    at org.apache.kafka.connect.runtime.WorkerSourceTask$WorkerSourceTaskThread.execute(WorkerSourceTask.java:353)
    at org.apache.kafka.connect.util.ShutdownableThread.run(ShutdownableThread.java:82)
```

The resulting avro message ends up with all boolean values being null (and also I think these failures might slow down the import process quite a lot).

### LINKED ISSUE #303: Add DBMS-specific dialects to source connector
Change the JDBC source connector to use dialect components, and create the first default dialect that has all of the existing functionality. The framework should determine which dialect to use based upon the JDBC connection URL string and/or the JDBC Connection instance. (Note the dialect is not involved in establishing the connection.) Include an abstract test class that can be used to easily verify the functionality of individual dialects.

This change will make it easier to correct DBMS-specific issues, including but not limited to:
* #230 
* #231 
* #101 
* #103 
* #213 
* #221 
* #299 
* #101 
* #103 
* #116 
* #81 
* #211 
* #46

Feel free to mention this issue on other issues that will be easier to fix once the source connector has DBMS-specific dialects.

### LINKED ISSUE #421: Field type warning is of limited use and could be modified to include field name.
https://github.com/confluentinc/kafka-connect-jdbc/blob/a1a8360a11d855917e16d7ffc402adc25cf666ee/src/main/java/io/confluent/connect/jdbc/source/DataConverter.java#L104

Regarding this line,
It produces logs such as:
[2018-05-31 10:49:45,316] WARN JDBC type 1111 not currently supported (io.confluent.connect.jdbc.source.JdbcSourceTask:352)

I suggest that line 104 ought to be changed to something like:
`log.warn("JDBC type {} not currently supported for column {}", sqlType, fieldName);`

This would let users know the name of the field which is causing the unsupported type error to be thrown.

rhauch:
The source and sink JDBC connectors were refactored to use dialects for all interaction with the database, allowing for the customization of behavior when working with specific DBMSes. Dialects are extensible and auto-discovered via Java’s service loader mechanism. Separate dialects were added for each major DBMS, though only DBMS-specific behavior already in the connector was extracted into those dialects; additional customization can be done on the dialects to handle DBMS-specific issues.

Prior to this change, the sink connector made use of an earlier dialect framework that was not very extensible, and it only encapsulated a limited subset of the interaction with the database. This new dialect framework is completely extensible (dialects can be added at runtime by simply including in the plugin or on the classpath one or more JARs with the custom dialects), and it completely encapsulates all interaction with the database so that dialects can customize any of the behavior.

A generic dialect is included as a complete implementation that uses JDBC and standard SQL, and custom dialects should extend this generic implementation so that the dialect overrides only the behavior that needs to be customized. The majority of the code within a dialect is needed by the source and sink logic, so there is no longer a distinction between source and sink dialects; each dialect supports all features of both source and sink connectors.

The old sink dialect framework was removed (rather than just deprecated), since it was difficult for users to provide their own custom sink dialect implementations, as the method that determined which sink dialect to use was static and thus not easily customized. Also refactored the sink’s database metadata queries and utilities to reuse components already used in the source.
  
Fixes #303
Fixes #421 
Fixes #35


rhauch:
Rebased and corrected the checkstyle errors.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialects.java
```diff
@@ -0,0 +1,198 @@
...
+   * @param jdbcUrl the JDBC connection URL; may not be null
+   * @param config  the connector configuration used to create the dialect; may not be null
+   * @return the {@link DatabaseDialect} instance with the greatest score; never null, but possibly
+   *     the {@link DatabaseDialect default DatabaseDialect}
+   * @throws ConnectException if there is a problem with the JDBC URL
+   */
+  public static DatabaseDialect findBestFor(
+      String jdbcUrl,
+      AbstractConfig config
+  ) throws ConnectException {
```
rhauch: This class and in particular this method is the starting point for using dialects. Given a JDBC URL and a source or sink connector configuration, this method iterates through all of the registered `DatabaseDialectProvider` instances and has each provider score the given JDBC URL. The provider with the highest score will be used as a factory to create a new `DatabaseDialect` instance with the supplied connector configuration.

Although `DatabaseDialectProvider` implementations can be registered manually, this class uses Java's Service Provider mechanism to automatically find and register all `DatabaseDialectProvider` implementations that are accessible via this class' classloader. That means that anyone can implement a custom `DatabaseDialectProvider`, build a properly-constructed JAR file w/ provider manifest file, and add that JAR file to the JDBC connector plugin directory. (See the JavaDoc for the `DatabaseDialectProvider` interface for specifics.)

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialectProvider.java
```diff
@@ -0,0 +1,215 @@
...
+    return dialectName();
+  }
+
+  /**
+   * An abstract base class for a provider that will return a
+   * {@link DatabaseDialectProvider#PERFECT_MATCHING_SCORE}
+   * if any of the supplied JDBC subprotocols match the subprotocol from the JDBC connection URL, or
+   * {@link DatabaseDialectProvider#NO_MATCH_SCORE} otherwise.
+   */
+  public abstract static class SubprotocolBasedProvider extends DatabaseDialectProvider {
```
rhauch: Here's an abstract subclass of `DatabaseDialectProvider` that we used in all of the DBMS-specific dialect providers. Simply provide the subprotocols, and URLs that use one of those subprotocols will be scored with a "perfect" score; anything else gets a "no match" score.

src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
```diff
@@ -0,0 +1,162 @@
...
+  public static class Provider extends SubprotocolBasedProvider {
+    public Provider() {
+      super(Db2DatabaseDialect.class.getSimpleName(), "db2", "db2j", "ibmdb");
+    }
+
+    @Override
+    public DatabaseDialect create(AbstractConfig config) {
+      return new Db2DatabaseDialect(config);
+    }
+  }
```
rhauch: Here's an example of how a dialect typically defines a custom `DatabaseDialectProvider` implementation class. It extends the `SubprotocolBasedProvider` so that it perfectly matches any JDBC URL with the `db2`, `db2j`, or `imdb` subprotocols, and it provides a factory method for the actual `Db2DatabaseDialect` implementation class.

Once again, this class is never directly used or referenced anywhere. Instead, the fully-qualified name of this provider class is specified in the `META-INF/services/io.confluent.connect.jdbc.dialect.DatabaseDialectProvider` file

src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Create a new dialect instance with the given connector configuration.
+   *
+   * @param config the connector configuration; may not be null
+   */
+  public Db2DatabaseDialect(AbstractConfig config) {
+    super(config, new IdentifierRules(".", "\"", "\""));
```
rhauch: The `GenericDatabaseDialect` serves as a base class that implements everything necessary for the JDBC source connector and most methods necessary for the JDBC sink connector. Thus, dialect implementations only implement the methods that are necessary to customize the behavior of the `GenericDatabaseDialect` class.

One of the things the `GenericDatabaseDialect` class does is use the supplied "identifier rules" that dictate the segment delimiter string, leading quote string, and trailing quote string used in the DBMS' fully-qualified identifiers. DB2 uses the period (`.`) for delimiters, and double quotes for leading and trailing quotes. With these rules, the `GenericDatabaseDialect` can build fully-qualified table names used in SQL statements as well as parse strings into fully-qualified names, allowing the connector user to now support either simple table names and fully-qualified names in the whitelist and blacklist filters; if needed, the names can optionally be quoted using the DBMS-specific quoting characters.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,471 @@
...
+ * possible to create an implementation outside of this project and to package the dialect (s) and
+ * service provider file in a JAR file, and to then include that JAR file inside the JDBC
+ * Connector's plugin directory within a Connect installation.
+ *
+ * <p>This discovery and registration process uses DEBUG messages to report the {@link
+ * DatabaseDialectProvider} classes that are found and registered. If you have difficulties getting
+ * the connector to find and register your dialect implementation classes, check that your JARs have
+ * the service provider file and your JAR is included in the JDBC connector's plugin directory.
+ */
+public interface DatabaseDialect extends ConnectionProvider {
```
rhauch: This interface defines methods for all interaction with a database via the JDBC driver. Unfortunately, there are a lot of methods. Some methods may seem strange because they could be implemented by directly using the JDBC driver, but there are lots of different driver implementations and unfortunately all of them have their own quirks.

When you review, please look at the signatures of the public methods, since these will become part of the public API.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,471 @@
...
+  /**
+   * Get a list of non-system table names in the database.
+   *
+   * @param conn database connection
+   * @return a list of tables; never null
+   * @throws SQLException if there is an error with the database connection
+   */
+  List<TableId> tableNames(
+      Connection conn
+  ) throws SQLException;
```
rhauch: One of the challenges dealing with many different JDBC drivers is that they use different combinations of "catalog", "schema", and "table" in their fully-qualified names. To properly deal with these, you can't simply deal with table names as we were before (there are quite a few issues related to simple names not uniquely identifying tables within a database), so I've introduced `TableId` and `ColumnId` to represent fully-qualified table names and column names, and they can deal with any combination of catalog, schema, and table names.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,471 @@
...
+
+  /**
+   * Get a new {@link ExpressionBuilder} that can be used to build expressions with quoted
+   * identifiers.
+   *
+   * @return the builder; never null
+   * @see #identifierRules()
+   * @see IdentifierRules#expressionBuilder()
+   */
+  ExpressionBuilder expressionBuilder();
```
rhauch: There is quite a bit of code that builds SQL query and update expressions, and all of those need to be able to deal with fully-qualified identifiers _and_ the unique quoting and identifier rules. The `ExpressionBuilder` makes it easier to build a SQL expression without having to worry about those details, since the expression builder delegates to an `IdentifierRules` for the details. However, a DBMS can customize the behavior of `ExpressionBuilder`, and so the dialect must be able to create instances.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,471 @@
...
+   * @param incrementingColumn the identifier of the incremented column; may be null if there is
+   *                           none
+   * @param timestampColumns   the identifiers of the timestamp column; may be null if there is
+   *                           none
+   * @return the {@link TimestampIncrementingCriteria} implementation; never null
+   */
+  TimestampIncrementingCriteria criteriaFor(
+      ColumnId incrementingColumn,
+      List<ColumnId> timestampColumns
+  );
```
rhauch: The timestamp and incrementing criteria is some of the most widely varied we've seen across different DBMSes, so this method allows a dialect to generate this criteria. Other similar methods do the same thing for other kinds of criteria and statements.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,471 @@
...
+   * @throws SQLException if there is a problem binding the value into the statement
+   * @see #statementBinder(PreparedStatement, PrimaryKeyMode, SchemaPair, FieldsMetadata,
+   * InsertMode)
+   */
+  void bindField(
+      PreparedStatement statement,
+      int index,
+      Schema schema,
+      Object value
+  ) throws SQLException;
```
rhauch: This is also something that should be straightforward with the JDBC API, but isn't. Any time a custom dialect overrides the default data type mapping, it may need to change how the values are bound to variables in a prepared statement. For example, `INT8` corresponds to SQL's `BYTE` type, except that SQL `BYTE` is unsigned while `INT8` and Java's `byte` are signed. Most JDBC drivers handle this correctly, but the Sybase driver does nothing special and thus some valid Java `byte` values cannot be inserted into a Sybase `BYTE` column. Therefore, the `SybaseDatabaseDialect` has to instead use `short` to properly handle the signed values, and this changes how `INT8` values are bound to prepared statements for insert/update.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+
+/**
+ * A {@link DatabaseDialect} implementation that provides functionality based upon JDBC and SQL.
+ *
+ * <p>This class is designed to be extended as required to customize the behavior for a specific
+ * DBMS. For example, override the {@link #createColumnConverter(ColumnMapping)} method to customize
+ * how a column value is converted to a field value for use in a {@link Struct}. To also change the
+ * field's type or schema, also override the {@link #addFieldToSchema} method.
+ */
+public class GenericDatabaseDialect implements DatabaseDialect {
```
rhauch: The `GenericDatabaseDialect` provides a complete implementation of all methods used by the JDBC source connector and nearly all the methods used by the JDBC sink connector. That does mean that in order to use the JDBC sink connector with a specific DBMS, there must be a custom dialect that _at least_ defines the mapping between Connect's Schema types and the DBMS data types.

This class has to provide a lot of methods and behaviors, and unfortunately that means that this class is quite large. Also, this class needs to be easily extended to override custom behavior, so this class defines a lot of protected methods so that subclasses can override the narrower protected method without having to duplicate a lot of logic in one of the public methods. 

Pulling all of these methods and behaviors into this class makes it very straightforward to implement a custom dialect for a DBMS.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+      super(GenericDatabaseDialect.class.getSimpleName(),
+            DatabaseDialectProvider.AVERAGE_MATCHING_SCORE
+      );
+    }
+
+    @Override
+    public DatabaseDialect create(AbstractConfig config) {
+      return new GenericDatabaseDialect(config);
+    }
+  }
```
rhauch: The `GenericDatabaseProvider` is one of the few of our dialect implementations that uses the `FixedScoreProvider` base class. The `GenericDatabaseProvider` is the _default_ provider, so it always returns the "average" score, which is lower than a "perfect" score (when the URL has a matching subprotocol supported by the dialect) but higher than a "no match" score (when the URL does not have a matching subprotocol).

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+    String username = config.getString(JdbcSourceConnectorConfig.CONNECTION_USER_CONFIG);
+    Password dbPassword = config.getPassword(JdbcSourceConnectorConfig.CONNECTION_PASSWORD_CONFIG);
+    Properties properties = new Properties();
+    if (username != null) {
+      properties.setProperty("user", username);
+    }
+    if (dbPassword != null) {
+      properties.setProperty("password", dbPassword.value());
+    }
+    properties = addConnectionProperties(properties);
```
rhauch: The `addConnectionProperties(Properties)` method is an example of a protected method that allows subclasses to override which properties are used without having to duplicate all of the non-trivial logic in this method. Method like this make it easier to subclass without duplicating a lot of logic, but it makes the class bigger. IMO, small price to pay for smaller subclasses.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+  @Override
+  public boolean isConnectionValid(
+      Connection connection,
+      int timeout
+  ) throws SQLException {
+    if (jdbcDriverInfo().jdbcMajorVersion() >= 4) {
+      return connection.isValid(timeout);
+    }
+    // issue a test query ...
+    String query = checkConnectionQuery();
```
rhauch: Here's an example of how we can't simply call a JDBC 4 method, since they won't work on drivers that are only at JDBC 2 or 3.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+    return new ColumnDefinition(id, jdbcType, typeName, classNameForType, nullability, mutability,
+                                precision, scale,
+                                signedNumbers != null ? signedNumbers.booleanValue() : false,
+                                displaySize != null ? displaySize.intValue() : 0,
+                                autoIncremented != null ? autoIncremented.booleanValue() : false,
+                                caseSensitive != null ? caseSensitive.booleanValue() : false,
+                                searchable != null ? searchable.booleanValue() : false,
+                                currency != null ? currency.booleanValue() : false,
+                                isPrimaryKey != null ? isPrimaryKey.booleanValue() : false
+    );
```
rhauch: This is another example of a method that can be easily overridden to customize the behavior, without having to override or copy the bulk of the logic for querying this information from the database.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+    switch (mapping.columnDefn().type()) {
+
+      case Types.BOOLEAN: {
+        return new ColumnConverter() {
+          @Override
+          public Object convert(ResultSet rs) throws SQLException {
+            return rs.getBoolean(col);
+          }
+        };
+      }
```
rhauch: Okay, I debated about going this route of returning a function `ColumnConverter` that can convert a column value into an object used in a source record, versus just having this method do the conversion and return the value directly. Dealing with functions in Java 7 is verbose, but in Java 8 nearly all of this boilerplate will go away. For example, the above block in Java 8 can be:

```
case Types.BOOLEAN: {
  return (rs)->rs.getBoolean(col);
}
```
Now, the benefit of returning a function is that any overhead of looking at the column definition or whether the driver supports JDBC 4 can be done once, and the resulting function can be called many times (once per record) without this overhead.

So, I decided that returning a function is (or rather _can be_) significantly more efficient in some cases, and it's worth the short-term verbosity for the next 6 months.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+        if (mapNumerics) {
+          int precision = defn.precision();
+          if (precision < 19 && defn.scale() == 0) { // integer
+            if (precision > 9) {
+              return new ColumnConverter() {
+                @Override
+                public Object convert(ResultSet rs) throws SQLException {
+                  return rs.getLong(col);
+                }
+              };
```
rhauch: While most of these cases don't have any custom logic, here's an example of where the overhead of looking at the precision and scale for a column becomes too expensive to do for every field in every row, and where returning a function is far superior.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+  ) {
+    ExpressionBuilder builder = expressionBuilder();
+    builder.append("INSERT INTO ");
+    builder.append(table);
+    builder.append("(");
+    builder.append(",", ExpressionBuilder.columnNames(), keyColumns, nonKeyColumns);
+    builder.append(") VALUES(");
+    builder.appendMultiple(",", "?", keyColumns.size() + nonKeyColumns.size());
+    builder.append(")");
+    return builder.toString();
```
rhauch: Here's an example of how the `ExpressionBuilder` is used. Here, `table` is a `TableId` and can be appended to the expression builder and automatically expanded into the fully-qualified quoted identifier. Line 1405 appends each key column and non-key column but delimited by a comma and properly quoted names (rather than fully-qualified identifiers).

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -0,0 +1,136 @@
...
+   * Create a new dialect instance with the given connector configuration.
+   *
+   * @param config the connector configuration; may not be null
+   */
+  public MySqlDatabaseDialect(AbstractConfig config) {
+    super(config, new IdentifierRules(".", "`", "`"));
+  }
+
+  @Override
+  protected String getSqlType(SinkRecordField field) {
```
rhauch: This is one of the few methods that _must_ be implemented in order to use this dialect in the sink connector. It maps from the record's schema type to the DBMS-specific data type.

But also note how this whole class is actually pretty short!

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -0,0 +1,210 @@
...
+    builder.append(")");
+    return Collections.singletonList(builder.toString());
+  }
+
+  @Override
+  public String buildUpsertQueryStatement(
+      final TableId table,
+      Collection<ColumnId> keyColumns,
+      Collection<ColumnId> nonKeyColumns
+  ) {
```
rhauch: Oracle UPSERT statements get nasty...

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -0,0 +1,261 @@
...
+          };
+        }
+        break;
+      }
+      default:
+        break;
+    }
+
+    // Delegate for the remaining logic
+    return super.createColumnConverter(mapping);
```
rhauch: Here's an example of a dialect that first handles DBMS-specific data types (e.g., `BIT` and `JSON`/`JSONB`), and then simply delegates to the `GenericDatabaseDialect` method for all of the standard types.

src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java
```diff
@@ -0,0 +1,290 @@
...
+    if (!autoCommit) {
+      connection.setAutoCommit(true);
+    }
+    try (Statement statement = connection.createStatement()) {
+      for (String ddlStatement : statements) {
+        statement.executeUpdate(ddlStatement);
+      }
+    } finally {
+      connection.setAutoCommit(autoCommit);
+    }
```
rhauch: Sybase is one of those DBMSes that does everything a bit differently. The JDBC sink connector turns off autocommit, so any DDL statements it executes is followed by an explicit commit. Of course Sybase doesn't like this, so here we have to temporarily turn on autocommit, apply the DDL statements, and then turn back off autocommit.

src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
```diff
@@ -1,320 +0,0 @@
...
-import javax.xml.bind.DatatypeConverter;
-
-import io.confluent.connect.jdbc.sink.metadata.SinkRecordField;
-import io.confluent.connect.jdbc.util.DateTimeUtils;
-
-import static io.confluent.connect.jdbc.sink.dialect.StringBuilderUtil.Transform;
-import static io.confluent.connect.jdbc.sink.dialect.StringBuilderUtil.joinToBuilder;
-import static io.confluent.connect.jdbc.sink.dialect.StringBuilderUtil.copiesToBuilder;
-
-public abstract class DbDialect {
```
rhauch: The JDBC sink connector already had a form of dialects. They didn't support much functionality, and still used the static `JdbcUtils` class to do most of the work. And, the sink connector used a static method to determine which dialect to use, which meant it couldn't be customized without subclassing the connector and then duplicating a lot of code. A search of GitHub and Google didn't turn up any subclasses or custom sink dialects.

Although the issue was around adding dialects for the _source connector_, I took this opportunity to refactor the existing dialects and have the new extensible `DatabaseDialect` framework support both sources and sinks. This helped to eliminate some duplicate code, and now a single custom `DatabaseDialect` implementation means that the source and sink connector will work with the DBMS.

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
```diff
@@ -66,18 +72,29 @@ protected ResultSet executeQuery() throws SQLException {
 
   @Override
   public SourceRecord extractRecord() throws SQLException {
-    Struct record = DataConverter.convertRecord(schema, resultSet, mapNumerics);
```
rhauch: Static methods like this are so difficult to customize.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -148,12 +149,21 @@
...
+  public static final String SCHEMA_PATTERN_DEFAULT = null;
+
+  public static final String CATALOG_PATTERN_CONFIG = "catalog.pattern";
+  private static final String CATALOG_PATTERN_DOC =
+      "Catalog pattern to fetch table metadata from the database:\n"
+      + "  * \"\" retrieves those without a catalog,"
+      + "  * null (default) means that the catalog name should not be used to narrow the search"
+      + " so that all table metadata would be fetched, regardless their catalog.";
+  private static final String CATALOG_PATTERN_DISPLAY = "Schema pattern";
+  public static final String CATALOG_PATTERN_DEFAULT = null;
```
rhauch: Some DBMSes like Sybase and SQL Server use the catalog for the database and the schema for the owner. That means that to support these DBMSes (and others like them) we have to allow users to be able to specify a catalog pattern.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -111,12 +121,10 @@ public void start(Map<String, String> properties) {
       offsets = context.offsetStorageReader().offsets(partitions);
     }
 
-    String schemaPattern
-        = config.getString(JdbcSourceTaskConfig.SCHEMA_PATTERN_CONFIG);
```
rhauch: The schema pattern and some of the other config properties are no longer needed by the task. However, they may be needed by the dialect, so if they're needed that's where they're used.

rhauch:
@kkonstantine and @wicknicks: this is ready for an in-depth review.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -26,61 +26,78 @@
...
 import io.confluent.connect.jdbc.sink.metadata.FieldsMetadata;
 import io.confluent.connect.jdbc.sink.metadata.SchemaPair;
+import io.confluent.connect.jdbc.util.ColumnId;
+import io.confluent.connect.jdbc.util.TableId;
 
 public class BufferedRecords {
   private static final Logger log = LoggerFactory.getLogger(BufferedRecords.class);
 
-  private final String tableName;
+  private final TableId tableId;
```
wicknicks: +1 for replacing `String` identifier to an object one.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+   *     should be returned; never null
+   */
+  protected Properties addConnectionProperties(Properties properties) {
+    return properties;
+  }
+
+  @Override
+  public PreparedStatement createPreparedStatement(
+      Connection db,
+      String query
```
wicknicks: should this method be here? It seems like it should be in a JDBC utils package/class than in a dialect implementation.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+    if (parts == null || parts.isEmpty()) {
+      throw new IllegalArgumentException("Invalid fully qualified name: '" + fqn + "'");
+    }
+    if (parts.size() == 1) {
+      return new TableId(null, null, parts.get(0));
+    }
+    if (parts.size() == 3) {
+      return new TableId(parts.get(0), parts.get(1), parts.get(2));
+    }
+    assert parts.size() >= 2;
```
wicknicks: this is always true

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: It does seem strange, but yes. This gives dialects the ability to set fetch direction, query timeouts, max rows for the result sets, etc.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: Correct, which is why I used an assertion. It's a great way to clarify the logic, but since asserts can be turned off there's no harm in stating them.

src/main/java/io/confluent/connect/jdbc/util/IdentifierRules.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public String trailingQuoteString() {
+    return trailingQuoteString;
+  }
+
+  public ExpressionBuilder expressionBuilder() {
+    return new ExpressionBuilder(this);
+  }
+
+  public List<String> parseQualifiedIdentifier(String fqn) {
```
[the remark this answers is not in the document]
rhauch: Not easily if at all. We have to be able to look for opening and closing quotes, since the valid characters are different inside quotes. Second, the variable nature of the actual char sequences we're looking for changes, and that'd make the regex pattern string more complicated to understand. Third, DBMSes are allowed to support multi-character quoting strings, which would complicate the regex even more. And finally, regexes aren't super performant and are useful when they would be dramatically easier to code; a really complicated regex loses all those benefits.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: Also, IMO it does not make sense to use a static JDBC utility method for things like this, since doing so prevents the ability customize the behavior -- that's what this PR is trying to move away from in the first place.

Now, one might argue that simply creating a prepared statement is something so basic that you shouldn't need anything but the JDBC API. And when you're working with a single DBMS that is definitely true. However, we're dealing with _lots_ of driver implementations and DBMSes, and they all have their quirks, limitations, and pain points. IMO, we _have_ to be prepared to customize almost all interaction with the JDBC driver.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+      String query
+  ) throws SQLException {
+    PreparedStatement stmt = db.prepareStatement(query);
+    stmt.setFetchDirection(ResultSet.FETCH_FORWARD);
+    return stmt;
+  }
+
+  @Override
+  public TableId parseTableIdentifier(String fqn) {
+    List<String> parts = identifierRules().parseQualifiedIdentifier(fqn);
```
kkonstantine: Knowing how diligent you are with javadocs, it's surprising that `parseQualifiedIdentifier` doesn't have one. Especially given that is returns this very particular list of strings. As a follow up, should this method return a class or an enum (depending on how extensible we expect it to be) instead of a list? I see the convenience behind using a list but it's hard to follow that's expected here. Not the end of the world either.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+   * @param cal  calendar
+   * @return the current time at the database
+   */
+  @Override
+  public Timestamp currentTimeOnDB(
+      Connection conn,
+      Calendar cal
+  ) throws SQLException, ConnectException {
+    String query = currentTimestampDatabaseQuery();
+    assert query != null;
```
kkonstantine: nit: Is `assert query != null && !query.isEmpty();` too condensed? I'm used to see those two checks together.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+    this.config = config;
+    this.defaultIdentifierRules = defaultIdentifierRules;
+    this.jdbcUrl = config.getString(JdbcSourceConnectorConfig.CONNECTION_URL_CONFIG);
+    if (!(config instanceof JdbcSinkConfig)) {
+      mapNumerics = config.getBoolean(JdbcSourceTaskConfig.NUMERIC_PRECISION_MAPPING_CONFIG);
+      catalogPattern = config.getString(JdbcSourceTaskConfig.CATALOG_PATTERN_CONFIG);
+      schemaPattern = config.getString(JdbcSourceTaskConfig.SCHEMA_PATTERN_CONFIG);
+      tableTypes = new HashSet<>(config.getList(JdbcSourceTaskConfig.TABLE_TYPE_CONFIG));
+    } else {
+      mapNumerics = JdbcSourceTaskConfig.NUMERIC_PRECISION_MAPPING_DEFAULT;
```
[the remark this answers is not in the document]
rhauch: The dialects are used for both sources and sinks, and the `JdbcSourceTaskConfig` properties are only used in methods used by the source task.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1753 @@
...
+   *                               to be determined from the database metadata
+   */
+  protected GenericDatabaseDialect(
+      AbstractConfig config,
+      IdentifierRules defaultIdentifierRules
+  ) {
+    this.config = config;
+    this.defaultIdentifierRules = defaultIdentifierRules;
+    this.jdbcUrl = config.getString(JdbcSourceConnectorConfig.CONNECTION_URL_CONFIG);
+    if (!(config instanceof JdbcSinkConfig)) {
```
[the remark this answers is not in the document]
rhauch: yeah, probably.

MrTrustworthy:
Reposting and expanding upon a comment on another thread on request of @rhauch 

I'm not exactly sure how it would be possible to run both a Debezium-MySQL dialect and a "normal" MySQL dialect on the same connect worker since they both have the same JDBC connection string. Maybe an additional config options "sql.dialect.hint" or something would be a useful tool that would allow dialects to break scoring ties in findBestFor()?

Ideally, it would not be necessary to implement an entire dialect to just provide a better debezium data type integration. I've played around with a "dialect extension" approach on the old dialect implementation but decided to scrap that approach due to too many roadblocks. Basically, in order to provide some extension mechanism for dialects that can deal with all eventualities in all database types, you'd need to provide so many hooks it would make the whole process harder than just to write a new dialect. 

But maybe your new implementation would make this feasible. For reference, in order to implement the debezium dialect, I needed to:
(1) have access to the schema from the schema registry whenever a PreparedStatement is assembled
(2) overwrite parts of the strings that are assembled into the queries. For CREATE TABLE statements, i needed to check for debezium datatypes in the schema and pick an appropriate MySQL datatype. For INSERT/UPDATE/UPSERT/... statements I relied on built-in MySQL functions to parse the data (but I'm not sure if this will work for all cases/databases). For example, in order to turn an integer back into a DATETIME for the MySQL, I would write a "FROM_UNIXTIME(?)" instead of a plain "?" in the string that will be used in a PreparedStatement.

Hope that helps
Felix

fcracker79:
Is there any news about merging this PR?

rhauch:
@fcracker79, this did not make the cut for the upcoming Confluent Platform release since it was too risky at too late a date. I'm hoping to merge this soon, though.

rhauch:
@MrTrustworthy, I'm not sure what you mean by "Debezium dialect". Debezium is an alternative Kafka Connect source connector that does not use JDBC in the same way that this connector does -- it only uses it to obtain the initial snapshot, and once that snapshot is complete it uses MySQL-specific APIs.

Now, you also talk about the Debezium data format. Debezium itself provides some SMTs, and one of those is able to unwrap the row-based information out of Debezium's envelope structure.

Despite that, I do think a dialect hint may be useful in some situations, so that's something we should probably incorporate. (Many drivers that support the same database use different subprotocols, so I'm not convinced that's a problem.)

MrTrustworthy:
@rhauch, to clarify what I mean regarding a debezium dialect:

Debezium supplies some additional information about data types in the SMT-unwraped data. For example, a timestamp in the source database might be serialized as INTEGER with the schemaName set to "io.debezium.time.MicroTimestamp". The default MySQL JDBC sink implementation can't deal with that and will not write the expected values into the sink database. The debezium dialect I've written in #367 is basically just a modified MySQL dialect that can deal with this additional meta information and correctly parse it to create both DDL and INSERT statements.

To go back to the dialect hint: Since it's basically just a specialized variation of the MySQL dialect, I needed some way to tell the connector it shouldn't use the standard MySQL dialect without having to change the JDBC connection string.

rhauch:
See also #404. The PostgreSQL dialect does not properly handle byte strings, since it uses `BLOB` (which exists and works, but isn't the same as MySQL BLOB, for example) and should instead use `BYTEA`.

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -169,8 +166,19 @@ public void stop() throws ConnectException {
...
+    } finally {
+      try {
+        cachedConnectionProvider.close();
+      } finally {
+        try {
+          if (dialect != null) {
+            dialect.close();
+          }
+        } finally {
+          dialect = null;
```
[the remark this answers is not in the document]
rhauch: This is more to handle the case if `close()` is called more than once.

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -169,8 +166,19 @@ public void stop() throws ConnectException {
       tableMonitorThread.join(MAX_TIMEOUT);
     } catch (InterruptedException e) {
       // Ignore, shouldn't be interrupted
+    } finally {
+      try {
+        cachedConnectionProvider.close();
+      } finally {
+        try {
+          if (dialect != null) {
+            dialect.close();
```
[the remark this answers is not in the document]
rhauch: Removed from the connection provider and kept in this class, since this is the class that instantiates the dialect.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,471 @@
...
+import io.confluent.connect.jdbc.util.ColumnId;
+import io.confluent.connect.jdbc.util.ConnectionProvider;
+import io.confluent.connect.jdbc.util.ExpressionBuilder;
+import io.confluent.connect.jdbc.util.IdentifierRules;
+import io.confluent.connect.jdbc.util.TableDefinition;
+import io.confluent.connect.jdbc.util.TableId;
+
+/**
+ * A component that is used to map datatypes and queries for JDBC sources. Every dialect should be
+ * threadsafe and all methods should be free of side effects.
```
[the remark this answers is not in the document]
rhauch: ack

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,471 @@
...
+  /**
+   * Return current time at the database
+   *
+   * @param conn database connection
+   * @param cal  calendar
+   * @return the current time at the database
+   * @throws SQLException if there is an error with the database connection
+   */
+  Timestamp currentTimeOnDB(
+      Connection conn,
```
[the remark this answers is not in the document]
rhauch: ack

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: If the assertion fails when they are combined, you don't know whether it is null or empty. By having separate assertions, when it does fail you know exactly why.

rhauch:
@kkonstantine, @wicknicks: I've rebased and hopefully addressed all of your comments, so this PR is ready for another review.

@MrTrustworthy, I've added a `dialect.name` configuration that when no provided or empty (the default), then the best dialect is used. However, when `dialect.name` is provided and matches the name of one of the available dialects, then that dialect will be used. Note that there is a recommender and validator, so the `dialect.name` configuration will be validated if given. Hopefully this satisfies your requirements by allowing you to explicitly specify which dialect you want to use.

rhauch:
retest this please

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+      case Types.DOUBLE: {
+        if (optional) {
+          builder.field(fieldName, Schema.OPTIONAL_FLOAT64_SCHEMA);
+        } else {
+          builder.field(fieldName, Schema.FLOAT64_SCHEMA);
+        }
+        break;
+      }
+
+      case Types.NUMERIC:
```
rhauch: @kkonstantine this is the first of two blocks which incorporate the substantiative portions of your recent changes from #396. Would love a sanity check here.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+      }
+
+      // FLOAT is, confusingly, double precision and effectively the same as DOUBLE. See REAL
+      // for single precision
+      case Types.FLOAT:
+      case Types.DOUBLE: {
+        return (rs) -> rs.getDouble(col);
+      }
+
+      case Types.NUMERIC:
```
rhauch: @kkonstantine this is the second of two blocks which incorporate the substantiative portions of your recent changes from #396. Would love a sanity check here.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+    }
+    connections.add(connection);
+    return connection;
+  }
+
+  @Override
+  public void close() {
+    for (Connection conn = connections.poll(); conn != null; conn = connections.poll()) {
+      try {
+        conn.close();
```
rayokota: Maybe use `org.apache.kafka.common.utils.Utils.closeQuietly`, which will catch `Throwable`?  It seems if `close()` throws a `Throwable`, the other connections will not get cleaned up.  This could cause issues on the server side.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+    if (jdbcDriverInfo().jdbcMajorVersion() >= 4) {
+      return connection.isValid(timeout);
+    }
+    // issue a test query ...
+    String query = checkConnectionQuery();
+    if (query != null) {
+      try (Statement statement = connection.createStatement()) {
+        boolean hasResultSet = statement.execute(query);
+        if (hasResultSet) {
+          statement.getResultSet().close();
```
rayokota: This `close()` should be in a finally block; otherwise a leak may occur

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -131,81 +144,102 @@ public void start(Map<String, String> properties) {
...
+    try {
+      running.set(false);
+      if (cachedConnectionProvider != null) {
+        cachedConnectionProvider.close();
+      }
+    } finally {
+      cachedConnectionProvider = null;
+      try {
+        if (dialect != null) {
+          dialect.close();
```
rayokota: Should we use `org.apache.kafka.common.utils.Utils.closeQuietly` ?

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+    return matchingTableTypes.toArray(new String[matchingTableTypes.size()]);
+  }
+
+  @Override
+  public IdentifierRules identifierRules() {
+    if (identifierRules.get() == null) {
+      if (defaultIdentifierRules != null && connections.peek() == null) {
+        // We've not yet established a connection to the database, so in this case
+        // just return the default rules ...
+        return defaultIdentifierRules;
```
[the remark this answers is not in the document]
rhauch: I was trying to avoid creating a connection just to be able to use the identifier rules in the generic case. But I think you're right -- that behavior is unexpected. All of the concrete types define specific identifier rules, so it's only the generic case that needs to talk to the database. Later in the method, If we can't establish a connection, we log a warning and use the defaults. So, I'll remove this if-block altogether.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -0,0 +1,461 @@
...
+   * given table or update an existing record in that table Variables for each key column should
+   * also appear in the WHERE clause of the statement.
+   *
+   * @param table         the identifier of the table; may not be null
+   * @param keyColumns    the identifiers of the columns in the primary/unique key; may not be null
+   *                      but may be empty
+   * @param nonKeyColumns the identifiers of the other columns in the table; may not be null but may
+   *                      be empty
+   * @return the upsert/merge statement; may not be null
+   */
```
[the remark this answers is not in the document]
rhauch: It's not caught, so using upsert with the general dialect will result in a failure, and this is the same behavior as before. Basically, you can't use upsert without a concrete dialect that overrides it.

I didn't add it to the JavaDoc because I thought it implied that it makes the implementation optional, when it really isn't optional if you want to use upsert.

However, I'll add some logic where this method is used to report a more meaningful exception.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: I just changed it to catch `Throwable` instead of `SQLException`.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
rhauch: I actually like the warning appearing on the same logger, so I just caught Throwable here and logged a warning.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: +1

src/main/java/io/confluent/connect/jdbc/util/ExpressionBuilder.java
```diff
@@ -0,0 +1,294 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+package io.confluent.connect.jdbc.util;
+
+import javax.xml.bind.DatatypeConverter;
+
+public class ExpressionBuilder {
```
C0urante: It'd be nice to have a little documentation on this class and some of its more commonly-used methods, since it's referenced pretty widely across the project.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: @rayokota, actually, we're not doing anything with the result set other than closing it. What would be in the try block?

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialectProvider.java
```diff
@@ -0,0 +1,208 @@
...
+  public static final int NO_MATCH_SCORE = 0;
+
+  /**
+   * The "average" score that should be returned by the {@link #score(JdbcUrlInfo)} method to signal
+   * that the DatabaseDialect instance can be used with a connection and .
+   */
+  public static final int AVERAGE_MATCHING_SCORE = 10;
+
+  /**
+   * The score for a dialect that is an excellent match.
```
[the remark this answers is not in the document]
rhauch: Yeah, there is currently no maximum. Do you suggest changing something?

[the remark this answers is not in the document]
C0urante: Either change the constant name to `EXCELLENT_MATCHING_SCORE` or change the value to `Integer.MAX_VALUE`. I think the first choice is better.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+    String query = checkConnectionQuery();
+    if (query != null) {
+      try (Statement statement = connection.createStatement()) {
+        boolean hasResultSet = statement.execute(query);
+        if (hasResultSet) {
+          statement.getResultSet().close();
+        }
+      }
+    }
+    return true;
```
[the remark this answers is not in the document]
rhauch: Yes, an exception will signify that the connection is not healthy. If the query succeeds, then we consider the connection health.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+    }
+  }
+
+  /**
+   * Get the query string to determine the current timestamp in the database.
+   *
+   * @return the query string; never null or empty
+   */
+  protected String currentTimestampDatabaseQuery() {
+    return "SELECT CURRENT_TIMESTAMP;";
```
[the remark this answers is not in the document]
rhauch: It doesn't matter, but I'll remove the semicolon here.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -169,31 +184,43 @@ public void close() throws SQLException {
...
+            tableId,
+            asColumns(fieldsMetadata.keyFieldNames),
+            asColumns(fieldsMetadata.nonKeyFieldNames)
         );
       default:
         throw new ConnectException("Invalid insert mode");
     }
   }
+
+  private Collection<ColumnId> asColumns(Collection<String> names) {
```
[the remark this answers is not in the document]
rhauch: Agree.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -0,0 +1,1727 @@
...
+    return new ColumnDefinition(
+        id,
+        jdbcType,
+        typeName,
+        classNameForType,
+        nullability,
+        mutability,
+        precision,
+        scale,
+        signedNumbers != null ? signedNumbers.booleanValue() : false,
```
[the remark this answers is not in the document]
rhauch: Correct, unboxing is not necessary, except `signedNumber` only makes sense for numbers and thus I'm defining `null` for non-numbers.

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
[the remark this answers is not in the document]
rhauch: I chose to not bring in the dependency and to simply catch the warning. I like having the log warning message be in the same logger, which will help when configuring logs for the JDBC connector.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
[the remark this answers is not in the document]
C0urante: Not sure I follow. The ternary check here to make sure a default value is used in the event that `signedNumbers` is null makes sense--but when it's non-null, why is an explicit call to `.booleanValue()` necessary? `signedNumbers` will be automatically unboxed from a `Boolean` Object to a `boolean` primitive without the call to `.booleanValue()`.

Issue #232, opened 2017-06-13
MYSQL - JDBC Source Connector issues with schema.pattern 


kvmkreddy:
JDBC Source connector is not working with 'schema.pattern' in MySQL. 

JDBC Source configurations:

> name=test_source_connector                                                                      
> connector.class=io.confluent.connect.jdbc.JdbcSourceConnector
> tasks.max=1
> connection.url=jdbc:mysql://<valid_hostname>:3306/
> connection.user=<valid_user>
> connection.password=<valid_password> 
> mode=timestamp+incrementing 
> incrementing.column.name=id 
> timestamp.column.name=created_on 
> topic.prefix=test_ 
> table.whitelist=some_table 
> schema.pattern=<valid_schema>

Exception stack trace is as follows:
`[2017-06-13 22:43:06,057] ERROR Stopping after connector error (org.apache.kafka.connect.cli.ConnectStandalone:99)
java.lang.IllegalArgumentException: Number of groups must be positive.
	at org.apache.kafka.connect.util.ConnectorUtils.groupPartitions(ConnectorUtils.java:42)
	at io.confluent.connect.jdbc.JdbcSourceConnector.taskConfigs(JdbcSourceConnector.java:127)
	at org.apache.kafka.connect.runtime.Worker.connectorTaskConfigs(Worker.java:230)
	at org.apache.kafka.connect.runtime.standalone.StandaloneHerder.recomputeTaskConfigs(StandaloneHerder.java:265)
	at org.apache.kafka.connect.runtime.standalone.StandaloneHerder.updateConnectorTasks(StandaloneHerder.java:295)
	at org.apache.kafka.connect.runtime.standalone.StandaloneHerder.putConnectorConfig(StandaloneHerder.java:182)
	at org.apache.kafka.connect.cli.ConnectStandalone.main(ConnectStandalone.java:93)`

It seems MySQL considering only schema name as catalog and not considering schema or schema pattern while fetching tables.

This returns all the tables in schema
>ResultSet rs = con.getMetaData().getTables("<valid_schema>", null, "%", null);

where as  
> ResultSet rs = con.getMetaData().getTables(null, "<valid_schema>", "%", null); 

returning none.

mayuanxiaonong:
pattern is used in Oracle db. In Mysql, use catalog.

rhauch:
[...]

PR #339, opened 2018-01-09
Add catalog configuration, and set catalog & schema after connection created.

mayuanxiaonong:
#232  Add `catalog` configuration, and set catalog & schema of connection after created. Used for MySQL-like database product etc.


mayuanxiaonong:
[clabot:check]

rhauch:
See #333 for a different approach that uses catalog pattern and schema pattern, without having to explicitly set the catalog, which differs in different DBMSes. #333 hasn't yet been fully tested, and is still in progress.

Issue #340, opened 2018-01-09
slow query on oracle when timestamp column's data type is date


mayuanxiaonong:
when I use oracle with a `date` type increasing column, the query become very slow. I found `stmt.setDate` is much better than `stmt.setTimestamp`.

ignaciofedz:
I have the same issue. It could be very useful to have this configuration option.

PR #341, opened 2018-01-09
Add timestamp.column.type configuration, to support various data types of timestamp column.

mayuanxiaonong:
#340   Add `timestamp.column.type` configuration. When set a column with `date` type as timestamp column, use `stmt.setDate` instead of `stmt.setTimestamp`, to speed up the query.


ignaciofedz:
+1 It would be interesting to integrate this modification to make the Oracle query more optimal and functional in an environment with many records.

PR #351, opened 2018-01-23
CC-1366: close ResultSets and Statements during shutdown

rayokota:
I went through the code to find possible places where a ResultSet might not be closed.  The only case seems to be in poll(), when either an exception occurs or the task is shutdown.    According to https://stackoverflow.com/questions/4507440/must-jdbc-resultsets-and-statements-be-closed-separately-although-the-connection, ResultSets should be closed even if the Connection is subsequently closed.  Again, this seems to be the only place where a ResultSet could go unclosed.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -243,14 +243,21 @@ public class JdbcSourceTask extends SourceTask {
 
         log.debug("Returning {} records for {}", results.size(), querier.toString());
         return results;
-      } catch (SQLException e) {
-        log.error("Failed to run query for table {}: {}", querier.toString(), e);
+      } catch (SQLException sqle) {
+        log.error("Failed to run query for table {}: {}", querier.toString(), sqle);
         resetAndRequeueHead(querier);
         return null;
+      } catch (Exception e) {
+        resetAndRequeueHead(querier);
+        throw e;
       }
     }
 
     // Only in case of shutdown
+    final TableQuerier querier = tableQueue.peek();
+    if (querier != null) {
+      resetAndRequeueHead(querier);
+    }
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -243,14 +243,21 @@ public void stop() throws ConnectException {
 
         log.debug("Returning {} records for {}", results.size(), querier.toString());
         return results;
-      } catch (SQLException e) {
-        log.error("Failed to run query for table {}: {}", querier.toString(), e);
+      } catch (SQLException sqle) {
+        log.error("Failed to run query for table {}: {}", querier.toString(), sqle);
         resetAndRequeueHead(querier);
         return null;
+      } catch (Exception e) {
```
rhauch: Maybe `Throwable`?

rayokota: Yes, probably safer to catch Throwable here.  I'll change it.  Thanks for the review!

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -243,14 +243,21 @@ public void stop() throws ConnectException {
 
         log.debug("Returning {} records for {}", results.size(), querier.toString());
         return results;
-      } catch (SQLException e) {
-        log.error("Failed to run query for table {}: {}", querier.toString(), e);
+      } catch (SQLException sqle) {
```
kkonstantine: nit: As I mentioned recently on another PR, we don't tend to name exception object with acronyms in the connectors. That's a stylistic comment to let you know what's the current trend. Up to you.

rayokota: Ok, thanks!  It looks like this would need to change across the board in this connector, as `sqle` is used pervasively.  I'll leave it like this for now.

PR #353, opened 2018-01-24
getObject() instead of getBytes()

Tang8330:
mysql-connector's implementation of getObject() - https://github.com/mysql/mysql-connector-j/blob/6.0.6/src/main/java/com/mysql/cj/jdbc/result/ResultSetImpl.java#L1222

Seems like we should be be using `getObject()` for this SQL data type.


Tang8330:
Re CLS: Soren Abildgaard has signed this on behalf of Zendesk (my organization) yesterday. Is that sufficient?

Tang8330:
[clabot:check]

PR #354, opened 2018-02-04
3.3.x fix postgres schema

edmondo1984:
This pull request covers the case where you need your connection to have a given schema but perform the inserts on a different one. It is a real use case taken from using Kafka Connect to integrate to Salesforce via the Heroku Connector.

The Heroku Connector is an extension that synchronize Salesforce bidirectionally with a Postgres on Heroku and such synchronization makes usage of triggers. One of those triggers references a function in the _default_ schema therefore:

- If you make your connection have  schema = salesforce, then the function won't be be found and the trigger will fail
- If you make your connection have schema = default, then the inserts of the connectors will go to the wrong tables.

As such, it is necessary to introduce a schema.override property in the configuration which allows to connect to schema A and performs the insert on schema B


edmondo1984:
Has this already been resolved in confluent 4 ?

rhauch:
@edmondo1984 , please see issue #303 and PR #333, which includes some DBMS-specific functionality and the ability to specify the catalogs and schemas. We could definitely add a Salesforce dialect, and I'd love to know whether the new dialect framework enables all of the customization you require.

PR #355, opened 2018-02-05
Ranged jdbc queries, qualified increment and timestamp column names, explicit start increments and timestamps

joncourt:
Added a few new features initially to support some large tables which the connector attempted to swallow whole (200mil+ rows) and a query that had a collision in the incrementing column name.

The following config settings have been added - which neatly describe the features I've added:

* `incrementing.span.max`: (default: 5,000,000) The maximum increment span to include in the query. Useful to limit query results size on very large tables
* `incrementing.start`: The starting point for an incrementer. This is useful where a max span has been set and the earliest increment is not 0.
* `incrementing.column.name.qualifier`: (default: "") The qualifier for the incrementing column if it's necessary. Use this to specify table alias for the column name where there is a conflict with other tables within the specified query.
* `require.non.zero.scale.incrementer`: (default: true) Usually an incrementing column has a scale of zero however many databases are tolerant of non-zero physical scale and assume a logical scale of zero anyway. Use this field to force the connector to ignore a non-zero physical scale.
* `timestamp.span.days.max`: (default: 100,000) The maximum number of days to include in a timestamp based offset span. Useful to limit query results size on very large tables.
* `timestamp.start`: (default: 1970-01-01 00:00:00.000) The initial timestamp to start rolling from. This is useful where a max timestamp span has been set and the earliest timestamp in the database is not 1970-01-01. Assume UTC and use format yyyy-MM-dd HH:mm:ss.SSS.
* `timestamp.column.name.qualifier`: (default: "") The qualifier for the timestamp column if it's necessary. Use this to specify table alias for the column name where there is a conflict with other tables within the specified query.

Additionally... I've switched the derby and sqlite in the tests to be memory based because I'm stuck on windows and all sorts of file lock problems crop up during the tests (that is the single most annoying thing about windows), using in memory based test databases neatly avoids this problem - but I'm unsure if there's a particular reason to use the file system based databases.


joncourt:
Brought the new tests up to date with the changes in the base class in master. Checks passing again.

sashutosh:
This is a very useful enhancement to Kafka connect . Something we were thinking to write ourselves, since we want to control the data to be loaded into Kafka.

joncourt:
Looks like the code base has moved on quite a bit since this code base... i'll close this PR for now and open it again if necessary later - updated.

Issue #361, opened 2018-02-14
Deadlock with big resultset in Ms Sql Server 


caiooliveiraeti:
When I try use a query mode with Ms Sql Server and this query return a large resultset a connector stop work.

```
java.lang.Thread.State: RUNNABLE
    at java.net.SocketInputStream.socketRead0(Native Method)
    at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
    at java.net.SocketInputStream.read(SocketInputStream.java:171)
    at java.net.SocketInputStream.read(SocketInputStream.java:141)
    at com.microsoft.sqlserver.jdbc.TDSChannel.read(IOBuffer.java:1981)
    at com.microsoft.sqlserver.jdbc.TDSReader.readPacket(IOBuffer.java:6310)
    - locked <0x00000000f07a3f80> (a com.microsoft.sqlserver.jdbc.TDSReader)
    at com.microsoft.sqlserver.jdbc.TDSCommand.detach(IOBuffer.java:7207)
    at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2468)
    - locked <0x00000000f1600090> (a java.lang.Object)
    at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:219)
    at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:199)
    at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeQueryInternal(SQLServerStatement.java:661)
    at com.microsoft.sqlserver.jdbc.SQLServerConnection.isValid(SQLServerConnection.java:4998)
    at io.confluent.connect.jdbc.util.CachedConnectionProvider.getValidConnection(CachedConnectionProvider.java:69)
    - locked <0x00000000f07a40f0> (a io.confluent.connect.jdbc.util.CachedConnectionProvider)
    at io.confluent.connect.jdbc.source.JdbcSourceTask.poll(JdbcSourceTask.java:230)
    at org.apache.kafka.connect.runtime.WorkerSourceTask.execute(WorkerSourceTask.java:179)
    at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:170)
    at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:214)
[...]

PR #362, opened 2018-02-14
Fix to call isValid connection only if necessary

caiooliveiraeti:
If call isValid in Ms Sql Server and has ResultSet is getting data in background jdbc driver has a problem. This solve a issue #361.

Call isValid only if need make a query.


caiooliveiraeti:
[clabot:check]

caiooliveiraeti:
[clabot:check]

Issue #363, opened 2018-02-15
Why use a fixed-size varchar in the MySQL dialect for sink connectors


MrTrustworthy:
Hey everyone,

I've just noticed a slightly confusing issue regarding the MySQL sink connector. If the schema specifies a "string" datatype, the MySQL dialect will create a column of type "VARCHAR(256)". 

This causes hard-to-find issues since messages that contain longer text passages are just silently truncated to fit the 256 characters limit (unless mysql is configured differently). This is a blocker for basically all tasks which deal with longer-than-256-character strings. 
Additionally, this is inconsistent with other database dialects like Postgres (see bottom of [this page](https://docs.confluent.io/current/connect/connect-jdbc/docs/sink_connector.html#features)) which use the TEXT datatype for creating columns from "string" input values.

Are there any reasons why the data type for "string" has been set to "VARCHAR(256)" and should remain that way?

Otherwise, the "fix" for this would be to switch to the TEXT data type in the MySQLDialect.java and adjust MySqlDialectTest.java accordingly. I can also submit a PR for that if you're interested.

Cheers
Felix

cricket007:
@MarkC0x Should this be closed?

MarkC0x:
Yeah, thanks for the reminder @cricket007

jekkos:
Changing the mapping causes some issues if you want to use a string field as primary key in MySQL. Currently I'm facing the following

```
java.sql.SQLSyntaxErrorException: BLOB/TEXT column 'siteId' used in key specification without a key length
```

PR #364, opened 2018-02-15
#363 Sink connector MySQL dialect - Using data type TEXT

MrTrustworthy:
Switched the column type of STRING from VARCHAR(256) to TEXT


rhauch:
This needs to be rebased since #333 has been merged, but it still should be a valid fix.

madnight:
@MrTrustworthy can you rebase this MR?

Issue #365, opened 2018-02-16
Have a configurable get current timestamp query


zevektor:
When in the JDBC Source Connector the mode is set to timestamp, the implementation tries to query the database to get the current time. This logic is implemented [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/util/JdbcUtils.java#L244-L251).

The query needed to get the current time could have been implemented as a string read from the config properties, instead I can see that the authors took for granted that most of the databases get the current timestamp with the query

`SELECT CURRENT_TIMESTAMP`

or a small subset of very specific variations. 

On the database I'm querying, I can get the current time using
`SELECT NOW()`
and the hardcoded query basically makes impossible to use this connector. 

Is there any way to easily replace this query? Otherwise I'm a bit stuck because I'd have to rely on the bulk mode, which I'd like to avoid.

zevektor:
Created #366 to add this feature

rhauch:
This should be addressed already as part of #333, without requiring the user to configure the timestamp yet still supporting the different ways each DBMS supports getting the current time. WDYT?

zevektor:
@rhauch your solution surely is cleaner and more elegant than mine, but for the most common use case it could be an overkill.

The current timestamp query is, most of the times, what really prevents the connector from working properly. When such problem occurs, a developer might only want to change this property instead of having to create an entire `DatabaseDialect` implementation for each of the most uncommon drivers.

PR #366, opened 2018-02-16
Added 'current time query' to customize the query to get current data…

zevektor:
…base time


zevektor:
#365

rhauch:
This should be addressed already as part of #333, without requiring the user to configure the timestamp yet still supporting the different ways each DBMS supports getting the current time. WDYT?

rhauch:
I'm going to close this since it should already be addressed, as mentioned above.

PR #367, opened 2018-02-20
Better Debezium data type integration

MrTrustworthy:
This is a very rough version of a better debezium data type integration to gauge if there is any interest in getting something like this into the mainline connector.

One of the main problems when using Debezium with the JDBC sink connector is that many data types in the source- and sink tables do not match at all. [Debezium reads the MySQL data types](http://debezium.io/docs/connectors/mysql/#data-types) and maps them into kafka data types, and the [JDBC sink connector re-maps those back into MySQL data types](https://docs.confluent.io/current/connect/connect-jdbc/docs/sink_connector.html#features).

Effectively, this causes issues such as TIMESTAMP columns in the source table turning into LONG tables in the sink table. 

The proposed solution contains two parts:
* Implementing a DebeziumMySQL dialect (heavily based on the basic MySQL dialect) that can correctly map the debezium-generated schema back into the corresponding MySQL data types and transform data if necessary. Currently, I have only implemented the "UPSERT" semantics of this dialect.
* Implementing an optional configuration flag that allows an user to override the dialect to be used. Users can set "sql.dialect=debezium" to use the debezium dialect. This is backwards compatible since it will default to reading the dialect from the JDBC string (the current behaviour). 

Again, this is only a rough first implementation. Please let me know if there is any interest in this feature and if the proposed approach (adding custom dialects) seems fine to you.

Cheers
Felix

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
@@ -184,7 +184,8 @@ public class BufferedRecords {
         return dbDialect.getUpsertQuery(
             tableName,
             fieldsMetadata.keyFieldNames,
-            fieldsMetadata.nonKeyFieldNames
+            fieldsMetadata.nonKeyFieldNames,
+            fieldsMetadata
         );
       case UPDATE:
         return  dbDialect.getUpdate(
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -39,6 +39,19 @@ public class JdbcSinkConfig extends AbstractConfig {
 
   }
 
+  public enum SqlDialects {
+    DEFAULT,
+    SQLITE,
+    ORACLE,
+    SAP,
+    HANA,
+    VERTICA,
+    SQLSERVER,
+    MARIADB,
+    MYSQL,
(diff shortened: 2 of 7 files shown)


rhauch:
@MrTrustworthy, please take a look at #333 for the proposed new DBMS-specific source and sink dialects. It should be much easier to customize the behavior, and one can even add custom dialects to an installation. Would be interested in feedback as well as thoughts on how this PR might be adapted to build on top of the new dialects.

MrTrustworthy:
Thanks for the comment @rhauch 

Your PR provides a much more sophisticated framework for custom SQL dialects. Is there any ETA about getting this into the mainline/next release? I'd definitely be happy to port the Debezium dialect into the new framework further down the line.

I'm not exactly sure how it would be possible to run both a Debezium-MySQL dialect and a "normal" MySQL dialect on the same connect worker since they both have the same JDBC connection string. Maybe an additional config options "sql.dialect.hint" or something would be a useful tool that would allow dialects to break scoring ties in findBestFor()?

rhauch:
@MrTrustworthy, excellent suggestion on the `sql.dialect.hint`. Would you mind adding that as a comment on #333?

BTW, a "Debezium-MSQL" dialect would really be a specialization of the MySQL dialect. But I wonder if there's a need for more general mechanism to override SQL types. Not sure what that'd look like, but it'd be a shame to have to create a dialect to provide that specialization.

MrTrustworthy:
@rhauch Responded in your PR so this PR can stay closed

PR #381, opened 2018-03-19
Fix to call isValid and Adds appender increment control

caiooliveiraeti:
1 - Fix to call isValid connection only if necessary
   If call isValid in Ms Sql Server and has ResultSet is getting data in background jdbc driver has a       problem. This solve a issue #361.
   Call isValid only if need make a query.
2 - Adds appender increment control


rhauch:
This likely needs to take into account the changes from #333.

PR #392, opened 2018-03-30
CC-918 Add Kafka Connect Maven plugin to build

C0urante:
**IMPORTANT**: This PR is targeted at `master` for the moment but should be redirected to `4.1.x` **after** the release occurs. DO NOT MERGE UNTIL THEN

Edit: We've now decided to redirect to `4.0.x`, and will be merging after the release is officially announced.

This PR adds the Kafka Connect Maven plugin to the project's build. Running `mvn package` should cause a plugin archive file to be generated in the `target/plugins/packages` directory, named `confluentinc-kafka-connect-jdbc-<version>-plugin.zip`.


pom.xml
```diff
@@ -142,6 +142,58 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>io.confluent</groupId>
+                <version>0.9.0-SNAPSHOT</version>
```
rhauch: Why are we using a snapshot release here?

pom.xml
```diff
@@ -142,6 +142,58 @@
...
+                <artifactId>kafka-connect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>kafka-connect</goal>
+                        </goals>
+                        <configuration>
+                            <title>Kafka Connect JDBC</title>
+                            <logo>logos/jdbc.png</logo>
```
rhauch: Can we legally use this logo?

pom.xml
```diff
@@ -142,6 +142,58 @@
...
+                <groupId>io.confluent</groupId>
+                <version>0.9.0-SNAPSHOT</version>
+                <artifactId>kafka-connect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>kafka-connect</goal>
+                        </goals>
+                        <configuration>
+                            <title>Kafka Connect JDBC</title>
```
rhauch: If we don't include a title, doesn't it take the title from the POM? Why duplicate this here if it's the same?

pom.xml
```diff
@@ -142,6 +142,58 @@
...
+                            <logo>logos/jdbc.png</logo>
+                            <sourceUrl>https://github.com/confluentinc/kafka-connect-jdbc</sourceUrl>
```
rhauch: Isn't this inherited?

pom.xml
```diff
@@ -142,6 +142,58 @@
...
+                            <pluginTypes>
+                                <pluginType>sink</pluginType>
+                                <pluginType>source</pluginType>
+                            </pluginTypes>
+
+                            <tags>
+                                <tag>jdbc</tag>
+                                <tag>java</tag>
+                                <tag>database</tag>
+                                <tag>schema evolution</tag>
```
rhauch: I'd also add `dbms`, `rdbms`, and `sql`, and remove `java`.

pom.xml
```diff
@@ -142,6 +142,58 @@
...
+                            </tags>
+
+                            <requirements>
+                                <requirement>Both: Confluent Platform installed and running</requirement>
```
rhauch: Isn't "Both: " here superfluous?

pom.xml
C0urante: Agh, forgot to change that. Important to note that we shouldn't actually use the `0.9.0` release when this is actually merged though, since it generates incorrectly-formatted plugin archives (see https://github.com/confluentinc/kafka-connect-packaging-plugins/pull/25).

pom.xml
rhauch: Also, shouldn't the user be able to install this connector from Confluent Hub and install into AK? How much value does this requirement add?

pom.xml
```diff
@@ -182,14 +181,12 @@
 
                             <tags>
                                 <tag>jdbc</tag>
-                                <tag>java</tag>
                                 <tag>database</tag>
                                 <tag>schema evolution</tag>
```
rhauch: I'd also remove `schema-evolution`. Sorry for missing that originally.

rhauch: Also, I wonder if we should include tags for at least some of the more popular DBMSes: `mysql`, `postgresql`, `oracle`, `sql server`, `db2`, `sybase`, `vertica`, `sap hana`, `derby`, `sqlite`.

C0urante:
Target branch redirected to `4.0.x`.

rhauch:
@C0urante, still WIP?

Issue #395, opened 2018-04-13
numeric error


luojianjava:
![image](https://user-images.githubusercontent.com/9797624/38728966-01198a5c-3f44-11e8-882b-e6642dd0885f.png)

![image](https://user-images.githubusercontent.com/9797624/38729124-7ed60e8e-3f44-11e8-975e-46d62fd7a2ec.png)

Why is sqlserver inconsistent with data stored in Kafka when data is numeric?

PR #396, opened 2018-04-17
CC-1843: Add config option for extended handling of numeric data type

kkonstantine:


src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -223,13 +229,47 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
             builder.field(fieldName, schema);
             break;
           }
+        } else if (mapNumerics == NumericMapping.COMPLETE) {
```
kkonstantine: Adding the implementation in a similar block here rather than consolidating potentially to a method. I'd prefer to keep the changes focused on this PR. There can be a follow up PR to make things more compact later.

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
```diff
@@ -30,14 +30,16 @@
...
+
 /**
  * BulkTableQuerier always returns the entire table.
  */
 public class BulkTableQuerier extends TableQuerier {
   private static final Logger log = LoggerFactory.getLogger(BulkTableQuerier.class);
 
   public BulkTableQuerier(QueryMode mode, String name, String schemaPattern,
-                          String topicPrefix, boolean mapNumerics) {
+                          String topicPrefix, NumericMapping mapNumerics) {
```
sansanichfb: Since you've updated parameter's type from `boolean` to `NumericMapping ` - maybe change parameter name as well, to something like `numbericMappingType`?

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -38,17 +38,23 @@
...
+import static io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig.NumericMapping;
+
 /**
  * DataConverter handles translating table schemas to Kafka Connect schemas and row data to Kafka
  * Connect records.
  */
 public class DataConverter {
   private static final Logger log = LoggerFactory.getLogger(JdbcSourceTask.class);
 
+  private static final int NUMERIC_TYPE_SCALE_LOW = -84;
```
sansanichfb: Would it be more readable to add small comment how do we get this number?

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
kkonstantine: Keeping the same name leaves some lines unchanged. Which is what I aim for with this PR. Doesn't sound too irrelevant anyways.

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -223,13 +229,47 @@ private static void addFieldSchema(ResultSetMetaData metadata, int col,
...
+            }
+          }
         }
         // fallthrough
 
       case Types.DECIMAL: {
         int scale = metadata.getScale(col);
-        if (scale == -127) { //NUMBER without precision defined for OracleDB
-          scale = 127;
+        if (scale == NUMERIC_TYPE_SCALE_UNSET) { //NUMBER without precision defined for OracleDB
```
sansanichfb: Nice refactoring 👍

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -587,6 +615,77 @@ public boolean visible(String name, Map<String, Object> config) {
...
+        validValues.add(e.toString().toLowerCase(Locale.ROOT));
+      }
+      return new EnumRecommender(canonicalValues, validValues);
+    }
+
+    @Override
+    public void ensureValid(String key, Object value) {
+      // calling toString on itself because IDE complains if the Object is passed.
+      if (!validValues.contains(value.toString())) {
+        throw new ConfigException(key, value, "Invalid enumerator");
```
sansanichfb: Maybe reframe a message to something like `Invalid enumeration value`

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -58,7 +64,7 @@ public static Schema convertSchema(
     return builder.build();
   }
 
-  public static Struct convertRecord(Schema schema, ResultSet resultSet, boolean mapNumerics)
+  public static Struct convertRecord(Schema schema, ResultSet resultSet, NumericMapping mapNumerics)
```
sansanichfb: Since this PR changes the logic of numeric conversions - maybe add a couple of unit-tests, exercising conversion of different numeric literals.

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -408,13 +450,43 @@ private static void convertFieldValue(ResultSet resultSet, int col, int colType,
...
+                colValue = resultSet.getShort(col);
+              } else {
+                colValue = resultSet.getByte(col);
+              }
+              break;
+            } else if (scale > 0) { // floating point
+              if (precision > 9) {
+                colValue = resultSet.getDouble(col);
+              } else {
+                log.info("float");
```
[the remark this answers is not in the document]
kkonstantine: leftover. fixed.

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
```diff
@@ -408,13 +450,43 @@ private static void convertFieldValue(ResultSet resultSet, int col, int colType,
...
+
         // fallthrough
       case Types.DECIMAL: {
         ResultSetMetaData metadata = resultSet.getMetaData();
+        int precision = metadata.getPrecision(col);
         int scale = metadata.getScale(col);
-        if (scale == -127) {
-          scale = 127;
+        log.trace("DECIMAL with precision: '{}' and scale: '{}'", precision, scale);
+        if (scale == NUMERIC_TYPE_SCALE_UNSET) {
```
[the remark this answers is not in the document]
kkonstantine: I only see one condition branch. (Also I'll err on keeping only the relative change here (which is the change from hardcoded value to the static)

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
kkonstantine: Keeping the old one to match what we have in `JdbcSinkConfig` (I added a comment to mention this comes from there).

src/main/java/io/confluent/connect/jdbc/source/DataConverter.java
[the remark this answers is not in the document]
rhauch: If I were writing this from scratch, I'd probably use ternary. However, @kkonstantine's intent on minimizing changes is the right choice here.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -89,6 +92,16 @@
   public static final boolean NUMERIC_PRECISION_MAPPING_DEFAULT = false;
   private static final String NUMERIC_PRECISION_MAPPING_DISPLAY = "Map Numeric Values By Precision";
 
+  public static final String NUMERIC_MAPPING_CONFIG = "numeric.mapping";
+  private static final String NUMERIC_MAPPING_DOC =
+          "Whether or not to attempt mapping NUMERIC values by precision to integral types";
```
rhauch: So it seems like this needs more detail, and it's no longer "Whether or not to attempt...".

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -89,6 +92,16 @@
   public static final boolean NUMERIC_PRECISION_MAPPING_DEFAULT = false;
```
rhauch: Might want to add how this is related to `numeric.mapping`, namely that `numeric.precision.mapping=true` is identical to `numeric.mapping=precision_only` but that `numeric.mapping=best_fit` is often a superior choice since it correctly maps numerics to integral and floating point numbers.

src/test/java/io/confluent/connect/jdbc/source/DataConverterTest.java
```diff
@@ -0,0 +1,186 @@
...
+    when(metadata.isNullable(1)).thenReturn(optional);
+    when(metadata.getPrecision(1)).thenReturn(precision);
+    when(metadata.getScale(1)).thenReturn(scale);
+
+    Schema schema = DataConverter.convertSchema("foo", metadata, numMapping);
+    Struct record = DataConverter.convertRecord(schema, resultSet, numMapping);
+
+    Object value = record.get("PrimitiveField1");
+    assertEquals(expectedValue, value);
+  }
```
rhauch: Wow, these are really nice and easy to read tests, @kkonstantine. Bravo!

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
wicknicks: I think we can remove this variable. It's only usage is in the ConfigDef below, which has been removed now.

docs/source_config_options.rst
```diff
@@ -71,6 +64,21 @@ Database
...
+  * Default: false
+  * Importance: low
+
+``numeric.mapping``
+  Map NUMERIC values by precision and optionally scale to integral or decimal types
+
+  * Type: string
+  * Default: null
+  * Valid Values: [none, precision_only, best_fit]
+  * Importance: low
```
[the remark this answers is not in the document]
rhauch: The following is my suggestion for a new subsection in `source_connector.rst` just after the `JDBC Drivers` subsection (and before `Examples`). Feel free to use this, adapt it, or start over.

```
Mapping Column Types
~~~~~~~~~~~~~~~~~~~~

The source connector has a few options for controlling how column types are mapped into
Connect field types. By default, the connector maps SQL/JDBC types to the most accurate representation,
which is straightforward for many SQL types but maybe a bit unexpected for some types.
For example, SQL's ``NUMERIC`` and ``DECIMAL`` types have very clear semantics controlled by
the precision and scale, and the most accurate representation is Connect's ``Decimal` logical type
that uses Java's ``BigDecimal`` representation. Unfortunately, Avro serializes ``Decimal`` types
in a binary that will often be difficult to work with.

The source connector's ``numeric.mapping`` configuration will control this mapping. The default
is ``none`` and has the behavior described above.

However, the ``best_fit`` option is probably what most users expect, as it attempts to map
``NUMERIC`` columns to Connect ``INT8``, ``INT16``, ``INT32``, ``INT64``, and ``FLOAT64``
based upon the column's precision and scale values:

============ ========== =================================
Precision    Scale      Connect "best fit" primitive type
============ ========== =================================
1 to 2       -84 to 0   INT8
3 to 4       -84 to 0   INT16
5 to 9       -84 to 0   INT32
10 to 18     -84 to 0   INT64
1 to 18      positive   FLOAT64

The ``precision_only`` option attempts to map ``NUMERIC`` columns to
Connect ``INT8``, ``INT16``, ``INT32``, and ``INT64`` based only upon the column's precision,
and where the scale is always 0.

============ ========== =================================
Precision    Scale      Connect "best fit" primitive type
============ ========== =================================
1 to 2       0          INT8
3 to 4       0          INT16
5 to 9       0          INT32
10 to 18     0          INT64

Anything other combination of precision and scale for ``NUMERIC`` columns will always map to
Connect's ``Decimal`` type.

.. note:: The ``numeric.precision.mapping`` property is older and is now deprecated. When 
enabled it is exactly equivalent to ``numeric.mapping=precision_only``, and when not enabled it 
is exactly equivalent to ``numeric.mapping=none``.

```

rhauch:
Fixes #101. Fixes #104. Fixes #395.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
kkonstantine: Leaving, since we still need to add something to the config definition. Once it's deprecated we can remove all the property specific variables.

kkonstantine:
Thanks for the great reviews everyone! Merging.

nik09541:
Hi @rhauch 
Please help in providing the latest jar for this snapshot as facing issues while building it. The repo for confluent has a restricted access. Please let us know when is release for this snapshot as it is needed urgently for handling numeric data types.

Thanks in advance

rmoff:
@nik09541 this code is part of Confluent Platform 4.1.1 onwards, you can download it from https://www.confluent.io/download/

PR #400, opened 2018-04-24
CC-1878: Allow for case differences while finding missing columns

wicknicks:
Signed-off-by: Arjun Satish <arjun@confluent.io>

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
@@ -180,8 +180,10 @@ public class DbStructure {
     }
 
     if (columnNamesLowerCase.size() != dbColumnNames.size()) {
-      log.warn("Table has column names which differ only with case sensitivity. Original "
-          + "columns={}", dbColumnNames);
+      log.warn(
+          "Table has column names that differ only by case. Original columns={}",
+          dbColumnNames
+      );
     }
 
     final Set<SinkRecordField> missingFieldsIgnoreCase = new HashSet<>();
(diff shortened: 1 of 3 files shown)


wicknicks:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -172,6 +172,33 @@ boolean amendIfNecessary(
...
+
+    // check if the missing fields can be located by ignoring case
+    Set<String> columnNamesLowerCase = new HashSet<>();
+    for (String columnName: dbColumnNames) {
+      columnNamesLowerCase.add(columnName.toLowerCase());
+    }
+
+    if (columnNamesLowerCase.size() != dbColumnNames.size()) {
+      log.warn("Table has column names which differ only with case sensitivity. Original "
+          + "columns={}", dbColumnNames);
```
rhauch: Perhaps different wording:

    Table has column names that differ from field names only by case. Original ...

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -172,6 +172,33 @@ boolean amendIfNecessary(
         missingFields.add(field);
       }
     }
-    return missingFields;
+
```
rhauch: Maybe return here if `missingFields.isEmpty()`?

wicknicks: This list is used to create ALTER table queries. So we need to return the set here.

rhauch: Right. If there are no missing fields at this point, why go through the work of checking case-insensitively? Why not return an empty list here?

PR #409, opened 2018-05-14
Add upsert mode for the Vertica dialect

sauliusvl:
When in `upsert` mode generate `MERGE` statements for the Vertica dialect of the form
```sql
MERGE INTO "table" AS target USING (
  SELECT ?::INT AS "id", ?::VARCHAR AS "name"
) AS incoming 
ON (target."id"=incoming."id")
WHEN MATCHED THEN UPDATE SET "name"=incoming."name","id"=incoming."id"
WHEN NOT MATCHED THEN INSERT ("name","id")
VALUES (incoming."name",incoming."id")
```
Unfortunately omitting the types from the `SELECT` statement and having non-varchar values makes Vertica error out with
```
com.vertica.support.exceptions.ErrorException: 
[Vertica][VJDBC](3376) ERROR: Failed to find conversion function from unknown to int
```
thus I had to somehow get the column type information passed to `getUpsertQuery()` - I'm open to suggestions as to how to do it otherwise (should I maybe rebase this on top of 
#333? or wait until it's merged?).

Casting to un-sized types since the destination table might be pre-created with differently sized types compared to the default ones, thus casting to explicit sizes might truncate values.

Note that the `WHEN MATCHED` branch includes key columns as advised by the Vertica [documentation](https://my.vertica.com/docs/9.1.x/HTML/index.htm#Authoring/AdministratorsGuide/Tables/MergeTables/OptimizedVsNonOptimizedMerge.htm).

### DIFF
--- a/docs/sink_connector.rst
+++ b/docs/sink_connector.rst
@@ -147,6 +147,7 @@ Oracle          `MERGE ..`
 PostgreSQL      `INSERT .. ON CONFLICT .. DO UPDATE SET ..`
 SQLite          `INSERT OR REPLACE ..`
 SQL Server      `MERGE ..`
+Vertica         `MERGE ..`
 Other           *not supported*
 ===========     ================================================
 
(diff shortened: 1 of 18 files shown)


rhauch:
@sauliusvl, thanks for submitting. Do you want to rebase this on top of the changes from #333 that were merged a few months back?

PR #416, opened 2018-05-23
CC-2043: Update pom.xml to target Java 8

kkonstantine:


pom.xml
```diff
@@ -142,6 +142,40 @@
...
+                    <execution>
+                        <id>test-compile</id>
+                        <phase>process-test-sources</phase>
+                        <goals>
+                            <goal>testCompile</goal>
+                        </goals>
+                        <configuration>
+                            <compilerArgs>
+                                <arg>-Xlint:all</arg>
+                                <arg>-Xlint:-unchecked</arg>
```
ijuma: Why are we disabling this?

kkonstantine: Was unable to use `@SafeVarargs` and thought initially it didn't worth adding individual suppressions. Turns out it's not that many and doesn't worth to keep different profiles between tests and main source code. Amended the commit.

PR #430, opened 2018-06-20
CC-2220 : Don't prefix schemaname & catalogname while building the `tables` config for JdbcSourceTaskConfig. This will be done by the expression builders.

mageshn:


src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -149,9 +149,9 @@ public void start(Map<String, String> properties) throws ConnectException {
       List<Map<String, String>> taskConfigs = new ArrayList<>(tablesGrouped.size());
       for (List<TableId> taskTables : tablesGrouped) {
         Map<String, String> taskProps = new HashMap<>(configProperties);
-        ExpressionBuilder builder = dialect.expressionBuilder();
-        builder.appendList().delimitedBy(",").of(taskTables);
-        taskProps.put(JdbcSourceTaskConfig.TABLES_CONFIG, builder.toString());
+        String tables = taskTables.stream().map(TableId::tableName)
+            .collect(Collectors.joining(","));
```
rhauch: Nit: this formatting could be better:

```
String tables = taskTables.stream()
                          .map(TableId::tableName)
                          .collect(Collectors.joining(","));

rhauch: I'm not convinced this is the correct thing to do, since `TableId.tableName` will lose all of the catalog and/or schemas, and `tables` will merely be a list of table names. If any table name is not unique in the user's scope, then the task will not understand which table to use.

mageshn: makes sense. Let me look at an alternative implementation.

PR #436, opened 2018-06-26
CC-2116: Fix offsets compatibility

mageshn:
With the dialect changes introduced in 5.0.x, the JDBCSourceTask always gets the table names as fully qualified table names. In the offset management, the connector was using the unqualified table name while writing the offsets and fully qualified table name for reading the offsets. Because of this when a connector restarts, it never finds the stored offsets and hence always starts from the beginning. This issue occurs only for TABLE mode.

The solution is to include a backward compatible offset management. We do the following

- Introduce a new offset protocol version. This will help evolve any changes to offsets in future.
- Offset writes will always follow new protocol version which is to include the FQN of the table
- During read, we include both the new protocol version offset key and the old one for all tables
- While selecting the offset, we always try the new protocol version key first followed by the old

This PR doesn't handle any changes to topic names. The topic names will continue to be unqualified table name to be backward compatible. If there are multiple tables, matching the intent is to error out and will be implemented in a separate PR.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
@@ -20,4 +20,5 @@ public class JdbcSourceConnectorConstants {
   public static final String TABLE_NAME_KEY = "table";
   public static final String QUERY_NAME_KEY = "query";
   public static final String QUERY_NAME_VALUE = "query";
+  public static final String OFFSET_PROTOCOL_VERSION = "version";
 }
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
@@ -20,6 +20,6 @@ public class JdbcSourceConnectorConstants {
   public static final String TABLE_NAME_KEY = "table";
   public static final String QUERY_NAME_KEY = "query";
   public static final String QUERY_NAME_VALUE = "query";
-  public static final String OFFSET_PROTOCOl_VERSION_KEY = "version";
+  public static final String OFFSET_PROTOCOL_VERSION_KEY = "version";
   public static final String PROTOCOL_VERSION_ONE = "1";
 }
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
@@ -21,4 +21,6 @@ public class JdbcSourceConnectorConstants {
   public static final String QUERY_NAME_KEY = "query";
   public static final String QUERY_NAME_VALUE = "query";
   public static final String OFFSET_PROTOCOL_VERSION = "version";
+  public static final String PROTOCOL_VERSION_ZERO = "0";
+  public static final String PROTOCOL_VERSION_ONE = "1";
 }
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
@@ -22,5 +22,4 @@ public class JdbcSourceConnectorConstants {
   public static final String QUERY_NAME_VALUE = "query";
   public static final String OFFSET_PROTOCOl_VERSION_KEY = "version";
   public static final String PROTOCOL_VERSION_ONE = "1";
-  public static final String TOPIC_NAME_KEY = "topic";
 }
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
@@ -20,6 +20,6 @@ public class JdbcSourceConnectorConstants {
   public static final String TABLE_NAME_KEY = "table";
   public static final String QUERY_NAME_KEY = "query";
   public static final String QUERY_NAME_VALUE = "query";
-  public static final String OFFSET_PROTOCOL_VERSION_KEY = "version";
+  public static final String OFFSET_PROTOCOL_VERSION_KEY = "protocol";
   public static final String PROTOCOL_VERSION_ONE = "1";
 }
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
@@ -20,7 +20,7 @@ public class JdbcSourceConnectorConstants {
   public static final String TABLE_NAME_KEY = "table";
   public static final String QUERY_NAME_KEY = "query";
   public static final String QUERY_NAME_VALUE = "query";
-  public static final String OFFSET_PROTOCOL_VERSION = "version";
-  public static final String PROTOCOL_VERSION_ZERO = "0";
+  public static final String OFFSET_PROTOCOl_VERSION_KEY = "version";
   public static final String PROTOCOL_VERSION_ONE = "1";
+  public static final String TOPIC_NAME_KEY = "topic";
 }
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -107,6 +107,7 @@ public class JdbcSourceTask extends SourceTask {
                                  ? Collections.singletonList(query) : tables;
 
     String mode = config.getString(JdbcSourceTaskConfig.MODE_CONFIG);
+    log.debug("Starting task with mode {} ",mode);
     Map<Map<String, String>, Map<String, Object>> offsets = null;
     if (mode.equals(JdbcSourceTaskConfig.MODE_INCREMENTING)
         || mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP)
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -21,6 +21,7 @@ import io.confluent.connect.jdbc.dialect.DatabaseDialects;
 import io.confluent.connect.jdbc.util.CachedConnectionProvider;
 import io.confluent.connect.jdbc.util.ColumnDefinition;
 import io.confluent.connect.jdbc.util.ColumnId;
+import io.confluent.connect.jdbc.util.ExpressionBuilder;
 import io.confluent.connect.jdbc.util.TableId;
 import io.confluent.connect.jdbc.util.Version;
 import org.apache.kafka.common.config.ConfigException;
@@ -156,10 +157,11 @@ public class JdbcSourceTask extends SourceTask {
               JdbcSourceConnectorConstants.TABLE_NAME_KEY,
               tableId.tableName()
           );
-          Map<String, String> partitionWithProtocolVer = new HashMap<>();
-          partitionWithProtocolVer.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, tableOrQuery);
-          partitionWithProtocolVer.put(JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION, "1");
-          partitions.add(partitionWithProtocolVer);
+          String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+          Map<String, String> partitionWithFqn = new HashMap<>();
+          partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+          partitionWithFqn.put(JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION, "1");
+          partitions.add(partitionWithFqn);
           partitions.add(partition);
           break;
         case QUERY:
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -240,8 +240,8 @@ public class JdbcSourceTask extends SourceTask {
     partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
     partitionWithFqn.put(
         JdbcSourceConnectorConstants.OFFSET_PROTOCOl_VERSION_KEY,
-        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE)
-    ;
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE
+    );
 
     Map<String, String> partition = Collections.singletonMap(
         JdbcSourceConnectorConstants.TABLE_NAME_KEY,
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -115,18 +115,7 @@ public class JdbcSourceTask extends SourceTask {
       switch (queryMode) {
         case TABLE:
           for (String table : tables) {
-            TableId tableId = dialect.parseTableIdentifier(table);
-            Map<String, String> partition = Collections.singletonMap(
-                JdbcSourceConnectorConstants.TABLE_NAME_KEY,
-                tableId.tableName()
-            );
-            partitions.add(partition);
-
-            String fqn = ExpressionBuilder.create().append(tableId, false).toString();
-            Map<String, String> partitionWithFqn = new HashMap<>();
-            partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
-            partitionWithFqn.put(JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION, "1");
-            partitions.add(partitionWithFqn);
+            partitionsForTable(partitions, table);
           }
           break;
         case QUERY:
@@ -137,7 +126,7 @@ public class JdbcSourceTask extends SourceTask {
           throw new ConnectException("Unknown query mode: " + queryMode);
       }
       offsets = context.offsetStorageReader().offsets(partitions);
-      log.debug("The partition offsets are {}", offsets);
+      log.trace("The partition offsets are {}", offsets);
     }
 
     String incrementingColumn
@@ -162,17 +151,7 @@ public class JdbcSourceTask extends SourceTask {
                 timestampColumns
             );
           }
-          TableId tableId = dialect.parseTableIdentifier(tableOrQuery);
-          partition = Collections.singletonMap(
-              JdbcSourceConnectorConstants.TABLE_NAME_KEY,
-              tableId.tableName()
-          );
-          String fqn = ExpressionBuilder.create().append(tableId, false).toString();
-          Map<String, String> partitionWithFqn = new HashMap<>();
-          partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
-          partitionWithFqn.put(JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION, "1");
-          partitions.add(partitionWithFqn);
-          partitions.add(partition);
+          partitionsForTable(partitions, tableOrQuery);
           break;
         case QUERY:
           partition = Collections.singletonMap(
@@ -191,9 +170,12 @@ public class JdbcSourceTask extends SourceTask {
         for (Map<String, String> toCheckPartition : partitions) {
           offset = offsets.get(toCheckPartition);
           if (offset != null) {
-            log.debug("Found non-null offset {} for partition {}", offsets, toCheckPartition);
-            offsetProtocolVersion = Integer.valueOf(toCheckPartition.getOrDefault(
-                JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION, "0"));
+            log.trace("Found non-null offset {} for partition {}", offsets, toCheckPartition);
+            offsetProtocolVersion = Integer.parseInt(
+                toCheckPartition.getOrDefault(
+                    JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION,
+                    JdbcSourceConnectorConstants.PROTOCOL_VERSION_ZERO
+                ));
             break;
           }
         }
@@ -251,6 +233,24 @@ public class JdbcSourceTask extends SourceTask {
     }
 
     running.set(true);
+  }
+
+  private void partitionsForTable(List<Map<String, String>> partitions, String table) {
+    TableId tableId = dialect.parseTableIdentifier(table);
+    Map<String, String> partition = Collections.singletonMap(
+        JdbcSourceConnectorConstants.TABLE_NAME_KEY,
+        tableId.tableName()
+    );
+    partitions.add(partition);
+
+    String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+    Map<String, String> partitionWithFqn = new HashMap<>();
+    partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+    partitionWithFqn.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION,
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE)
+    ;
+    partitions.add(partitionWithFqn);
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -36,6 +36,7 @@ import org.slf4j.LoggerFactory;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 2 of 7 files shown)


mageshn:
@rhauch I have not cleaned up the code yet and haven't added specific test yet. But wanted to get some initial feedback on the approach before spending time on those. Between, the test that was system test that was failing passes with this fix.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -149,21 +151,35 @@ public void start(Map<String, String> properties) {
...
+            log.trace("Found non-null offset {} for partition {}", offsets, toCheckPartition);
+            offsetProtocolVersion = Integer.parseInt(
+                toCheckPartition.getOrDefault(
+                    JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION,
+                    JdbcSourceConnectorConstants.PROTOCOL_VERSION_ZERO
+                ));
+            break;
+          }
+        }
+      }
```
rhauch: This logic is looking for a single protocol amongst all of the topics, yet it is breaking as soon as it finds the first. This might be easier to understand if we put that into a separate method, with JavaDoc that explains what it is doing and the assumptions it is making.

I'm not sure that the assumption about protocol versions might be valid. Imagine a connector runs with table A and B, and offsets are both stored. Then the connector is changed to use table A only and restarted. The offsets for table A and B both use protocol V0. Then the connector is upgraded to use protocol V1, and this updates the offsets for table A. Then the connector is changed to also consume from table B, but now the offsets for table A are V1 while the offsets for table B are still V1.

Is it not sufficient to always be able to read the offsets for each partition using that partition's protocol version? Obviously all future offsets would be written with the latest protocol version.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -149,21 +151,35 @@ public void start(Map<String, String> properties) {
...
+      if (offsets != null) {
+        for (Map<String, String> toCheckPartition : partitions) {
+          offset = offsets.get(toCheckPartition);
+          if (offset != null) {
+            log.trace("Found non-null offset {} for partition {}", offsets, toCheckPartition);
+            offsetProtocolVersion = Integer.parseInt(
+                toCheckPartition.getOrDefault(
+                    JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION,
+                    JdbcSourceConnectorConstants.PROTOCOL_VERSION_ZERO
+                ));
```
rhauch: Would be good to have the TRACE log message _after_ getting the protocol version, and then including the protocol version in the log message.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
rhauch: Oh, maybe this is simply trying to find the latest protocol version to use for writes? We definitely need to extract this to a separate method and have good JavaDoc. 

Why not always write using the most recent protocol version?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -216,6 +235,24 @@ public void start(Map<String, String> properties) {
...
+    partitions.add(partition);
+
+    String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+    Map<String, String> partitionWithFqn = new HashMap<>();
+    partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+    partitionWithFqn.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION,
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE)
+    ;
+    partitions.add(partitionWithFqn);
```
rhauch: Why are we adding both the single table name and FQN as separate partitions in the map?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -171,9 +174,17 @@ public SourceRecord extractRecord() throws SQLException {
...
+          String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+          partition = new HashMap<>();
+          partition.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+          partition.put(JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION, "1");
+          topic = topicPrefix + fqn;
+        } else {
+          String name = tableId.tableName(); // backward compatible
+          partition = Collections.singletonMap(JdbcSourceConnectorConstants.TABLE_NAME_KEY, name);
+          topic = topicPrefix + name;
+        }
```
rhauch: Seems like all of this could be extracted into a strategy function to compute the topic given the table ID (e.g., `topicName(TableId)`), and a second method to compute the partitions for a table ID (e.g., `sourcePartitions(TableId)`). We could have separate strategy implementations for each protocol version, and a particular strategy implementation can be passed into this class.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskTestBase.java
```diff
@@ -34,9 +34,16 @@
...
+  protected static Map<String, Object> SINGLE_TABLE_PARTITION_WITH_VERSION = new HashMap<>();
 
   static {
     SINGLE_TABLE_PARTITION.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, SINGLE_TABLE_NAME);
+    SINGLE_TABLE_PARTITION_WITH_VERSION.put(
+        JdbcSourceConnectorConstants.TABLE_NAME_KEY,
+        SINGLE_TABLE_NAME);
+    SINGLE_TABLE_PARTITION_WITH_VERSION.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION,
+        "1");
```
rhauch: This should all be moved into the PartitionProtocols class, which can have static instances of the strategy for each protocol.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -216,6 +235,24 @@ public void start(Map<String, String> properties) {
...
+    );
+    partitions.add(partition);
+
+    String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+    Map<String, String> partitionWithFqn = new HashMap<>();
+    partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+    partitionWithFqn.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION,
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE)
+    ;
```
wicknicks: nit: the `)` must be on the line with the semicolon.

wicknicks:
silly question: what is the problem we are solving in this PR? looks like it is related to CC-2116. Might be good to add a description above. thanks.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskTestBase.java
wicknicks: style nit: the `);` should be on the next line.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -116,7 +120,9 @@ public void testTimestampInvalidColumn() throws Exception {
 
   @Test
   public void testManualIncrementing() throws Exception {
-    expectInitializeNoOffsets(Arrays.asList(SINGLE_TABLE_PARTITION));
+    expectInitializeNoOffsets(Arrays.asList(
+        SINGLE_TABLE_PARTITION, 
+        SINGLE_TABLE_PARTITION_WITH_VERSION));
```
wicknicks: nit: the `);` should be on the next line. multiple occurrences in this file.

mageshn:
@rhauch I'm going to be adding a few more unit tests. But the latest changes gave me a green run for all the Mysql System tests
https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/944/console

mageshn:
@rhauch Do you want to take a pass at this now?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConstants.java
```diff
@@ -20,4 +20,6 @@
   public static final String TABLE_NAME_KEY = "table";
   public static final String QUERY_NAME_KEY = "query";
   public static final String QUERY_NAME_VALUE = "query";
+  public static final String OFFSET_PROTOCOL_VERSION_KEY = "version";
```
rhauch: I wonder if this should be "protocol" rather than "version", since the latter might be mistaken as the Connect version.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -105,6 +109,8 @@ public void start(Map<String, String> properties) {
                                  ? Collections.singletonList(query) : tables;
 
     String mode = config.getString(JdbcSourceTaskConfig.MODE_CONFIG);
+    //used only in table mode
+    Map<String, List<Map<String, String>>> tableToPartitions = new HashMap<>();
```
rhauch: I think `partitionsByTableFqn` would be a better name.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -113,9 +119,9 @@ public void start(Map<String, String> properties) {
       switch (queryMode) {
         case TABLE:
           for (String table : tables) {
-            Map<String, String> partition =
-                Collections.singletonMap(JdbcSourceConnectorConstants.TABLE_NAME_KEY, table);
-            partitions.add(partition);
+            List<Map<String, String>> tablePartitions = possibleTablePartitions(table);
```
rhauch: I think we need a comment here that describes what we're doing:

     // Find possible partition maps for different offset protocols ...

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -149,21 +157,32 @@ public void start(Map<String, String> properties) {
...
-          );
+          tablePartitionsToCheck = tableToPartitions.get(tableOrQuery);
           break;
         case QUERY:
           partition = Collections.singletonMap(
               JdbcSourceConnectorConstants.QUERY_NAME_KEY,
               JdbcSourceConnectorConstants.QUERY_NAME_VALUE
           );
+          tablePartitionsToCheck = new ArrayList<>();
+          tablePartitionsToCheck.add(partition);
```
rhauch: Suggestion to replace these two lines with:

    tablePartitionsToCheck = Collections.singleton(partition);

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -149,21 +157,32 @@ public void start(Map<String, String> properties) {
...
           break;
         default:
           throw new ConnectException("Unexpected query mode: " + queryMode);
       }
-      Map<String, Object> offset = offsets == null ? null : offsets.get(partition);
+
+      //select the first matching offset
+      // This helps in handling any backward compatible offset changes
```
rhauch: We need a better comment:
```
// The partition map varies by offset protocol. Since we don't know which protocol each table's offsets
// are keyed by, we need to use the different possible partitions (which are in newest protocol version first)
// to find the actual offsets for each table.

File: src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java [hunk 13]
```diff
@@ -216,6 +235,24 @@ public void start(Map<String, String> properties) {
     running.set(true);
   }
 
+  private List<Map<String, String>> possibleTablePartitions(String table) {
```
rhauch: Please add JavaDoc so that we better explain what this method is doing, especially around what the returned list means/contains.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -216,6 +235,24 @@ public void start(Map<String, String> properties) {
...
+    TableId tableId = dialect.parseTableIdentifier(table);
+
```
rhauch: How about adding a comment:

    // Define source partition map for offset protocol V1 that uses the table FQN

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -216,6 +235,24 @@ public void start(Map<String, String> properties) {
...
+    String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+    Map<String, String> partitionWithFqn = new HashMap<>();
+    partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+    partitionWithFqn.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION_KEY,
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE
+    );
+
```
rhauch: How about adding a comment:

    // Define source partition map for original offset protocol (V0) that uses the table name

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -171,9 +172,15 @@ public SourceRecord extractRecord() throws SQLException {
...
-        topic = topicPrefix + name;
+        String name = tableId.tableName();
+        topic = topicPrefix + name;// backward compatible
+        String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+        partition = new HashMap<>();
+        partition.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+        partition.put(
+            JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION_KEY,
+            JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE
+        );
```
rhauch: Didn't we talk about putting this logic into a helper method? Maybe an `OffsetProtocols` class with static methods:

    Map<String, String> sourcePartitionForProtocolV1(TableId table) { ... }
    Map<String, String> sourcePartitionForProtocolV0(TableId table) { ... }

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -216,6 +235,24 @@ public void start(Map<String, String> properties) {
...
+    Map<String, String> partition = Collections.singletonMap(
+        JdbcSourceConnectorConstants.TABLE_NAME_KEY,
+        tableId.tableName()
+    );
```
rhauch: This should call:

    Map<String, String> partition = OffsetProtocols.sourcePartitionForProtocolV0(tableId)`;

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -216,6 +235,24 @@ public void start(Map<String, String> properties) {
...
+  private List<Map<String, String>> possibleTablePartitions(String table) {
+    TableId tableId = dialect.parseTableIdentifier(table);
+
+    String fqn = ExpressionBuilder.create().append(tableId, false).toString();
+    Map<String, String> partitionWithFqn = new HashMap<>();
+    partitionWithFqn.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+    partitionWithFqn.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION_KEY,
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE
+    );
```
rhauch: This should call:

    Map<String, String> partition = OffsetProtocols.sourcePartitionForProtocolV1(tableId)`;

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskTestBase.java
```diff
@@ -34,9 +34,18 @@
...
   static {
     SINGLE_TABLE_PARTITION.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, SINGLE_TABLE_NAME);
+    SINGLE_TABLE_PARTITION_WITH_VERSION.put(
+        JdbcSourceConnectorConstants.TABLE_NAME_KEY,
+        SINGLE_TABLE_NAME
+    );
+    SINGLE_TABLE_PARTITION_WITH_VERSION.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION_KEY,
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE
+    );
```
rhauch: Could these could call `OffsetProtocols` methods?

mageshn:
@rhauch I have incorporated the comments. Would you like to take another pass at it.

src/main/java/io/confluent/connect/jdbc/source/OffsetProtocols.java
```diff
@@ -0,0 +1,45 @@
...
+package io.confluent.connect.jdbc.source;
+
+import io.confluent.connect.jdbc.util.ExpressionBuilder;
+import io.confluent.connect.jdbc.util.TableId;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OffsetProtocols {
```
rhauch: JavaDoc please?

src/main/java/io/confluent/connect/jdbc/source/OffsetProtocols.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public static Map<String, String> sourcePartitionForProtocolV1(TableId tableId) {
```
rhauch: JavaDoc please?

src/main/java/io/confluent/connect/jdbc/source/OffsetProtocols.java
```diff
@@ -0,0 +1,45 @@
...
+    Map<String, String> partitionForV1 = new HashMap<>();
+    partitionForV1.put(JdbcSourceConnectorConstants.TABLE_NAME_KEY, fqn);
+    partitionForV1.put(
+        JdbcSourceConnectorConstants.OFFSET_PROTOCOL_VERSION_KEY,
+        JdbcSourceConnectorConstants.PROTOCOL_VERSION_ONE
+    );
+    return partitionForV1;
+  }
+
+  public static Map<String, String> sourcePartitionForProtocolV0(TableId tableId) {
```
rhauch: JavaDoc please?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -171,9 +171,9 @@ public SourceRecord extractRecord() throws SQLException {
     final Map<String, String> partition;
     switch (mode) {
       case TABLE:
-        String name = tableId.tableName(); // backward compatible
-        partition = Collections.singletonMap(JdbcSourceConnectorConstants.TABLE_NAME_KEY, name);
-        topic = topicPrefix + name;
+        String name = tableId.tableName();
+        topic = topicPrefix + name;// backward compatible
+        partition = OffsetProtocols.sourcePartitionForProtocolV1(tableId);
```
rhauch: Isn't this constant for this querier instance? Can we initialize this once in the constructor and then simply reuse the same map over and over?

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -342,10 +423,38 @@ public void testAutoincrementRestoreOffset() throws Exception {
...
+  @Test
+  public void testTimestampRestoreVersionOneOffset() throws Exception {
+    TimestampIncrementingOffset offset = new TimestampIncrementingOffset(new Timestamp(10L), null);
+    testTimestampRestoreOffset(
+        Collections.singletonMap(SINGLE_TABLE_PARTITION_WITH_VERSION, offset.toMap())
+    );
+  }
+
+  @Test
+  public void testTimestampRestore() throws Exception {
```
rhauch: This is testing when we have offsets with both protocols. Is that worth having the method include that information? E.g., `testTimestampRestoreOffsetsWithMultipleProtocol()`

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -366,11 +475,38 @@ public void testTimestampRestoreOffset() throws Exception {
...
+  @Test
+  public void testTimestampAndIncrementingRestoreVersionOneOffset() throws Exception {
+    TimestampIncrementingOffset offset = new TimestampIncrementingOffset(new Timestamp(10L), 3L);
+    testTimestampAndIncrementingRestoreOffset(
+        Collections.singletonMap(SINGLE_TABLE_PARTITION_WITH_VERSION, offset.toMap())
+    );
+  }
+
   @Test
   public void testTimestampAndIncrementingRestoreOffset() throws Exception {
```
rhauch: This is testing when we have offsets with both protocols. Is that worth having the method include that information? E.g., `testTimestampAndIncrementingRestoreOffsetWithMultipleProtocol()`

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -316,11 +369,39 @@ public void testManualIncrementingRestoreOffset() throws Exception {
...
+  @Test
+  public void testAutoincrementRestoreVersionOneOffset() throws Exception {
+    TimestampIncrementingOffset offset = new TimestampIncrementingOffset(null, 1L);
+    testAutoincrementRestoreOffset(
+        Collections.singletonMap(SINGLE_TABLE_PARTITION_WITH_VERSION, offset.toMap())
+    );
+  }
+
   @Test
   public void testAutoincrementRestoreOffset() throws Exception {
```
rhauch: This is testing when we have offsets with both protocols. Is that worth having the method include that information? E.g., `testAutoincrementRestoreOffsetsWithMultipleProtocol()`

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -342,10 +423,38 @@ public void testAutoincrementRestoreOffset() throws Exception {
...
+        new Timestamp(8L),
+        null
+    );
     TimestampIncrementingOffset offset = new TimestampIncrementingOffset(new Timestamp(10L), null);
-    expectInitialize(Arrays.asList(SINGLE_TABLE_PARTITION),
-                     Collections.singletonMap(SINGLE_TABLE_PARTITION, offset.toMap()));
+    Map<Map<String, String>, Map<String, Object>> offsets = new HashMap<>();
+    offsets.put(SINGLE_TABLE_PARTITION_WITH_VERSION, offset.toMap());
+    offsets.put(SINGLE_TABLE_PARTITION, oldOffset.toMap());
+    testTimestampRestoreOffset(offsets);
```
rhauch: How about a comment here to say that we want to always use the offset with the latest protocol found?

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -366,11 +475,38 @@ public void testTimestampRestoreOffset() throws Exception {
...
   @Test
   public void testTimestampAndIncrementingRestoreOffset() throws Exception {
+    TimestampIncrementingOffset oldOffset = new TimestampIncrementingOffset(new Timestamp(10L), 2L);
     TimestampIncrementingOffset offset = new TimestampIncrementingOffset(new Timestamp(10L), 3L);
-    expectInitialize(Arrays.asList(SINGLE_TABLE_PARTITION),
-                     Collections.singletonMap(SINGLE_TABLE_PARTITION, offset.toMap()));
+    Map<Map<String, String>, Map<String, Object>> offsets = new HashMap<>();
+    offsets.put(SINGLE_TABLE_PARTITION_WITH_VERSION, offset.toMap());
+    offsets.put(SINGLE_TABLE_PARTITION, oldOffset.toMap());
+    testTimestampAndIncrementingRestoreOffset(offsets);
```
rhauch: How about a comment here to say that we want to always use the offset with the latest protocol found?

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -316,11 +369,39 @@ public void testManualIncrementingRestoreOffset() throws Exception {
...
   @Test
   public void testAutoincrementRestoreOffset() throws Exception {
+    TimestampIncrementingOffset oldOffset = new TimestampIncrementingOffset(null, 0L);
     TimestampIncrementingOffset offset = new TimestampIncrementingOffset(null, 1L);
-    expectInitialize(Arrays.asList(SINGLE_TABLE_PARTITION),
-                     Collections.singletonMap(SINGLE_TABLE_PARTITION, offset.toMap()));
+    Map<Map<String, String>, Map<String, Object>> offsets = new HashMap<>();
+    offsets.put(SINGLE_TABLE_PARTITION_WITH_VERSION, offset.toMap());
+    offsets.put(SINGLE_TABLE_PARTITION, oldOffset.toMap());
+    testAutoincrementRestoreOffset(offsets);
```
rhauch: How about a comment here to say that we want to always use the offset with the latest protocol found?

PR #438, opened 2018-06-29
Fail if the filtered tables list has duplicates based on unqualified table name

mageshn:
Fix for https://confluentinc.atlassian.net/browse/CC-2250

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
@@ -16,9 +16,6 @@
 
 package io.confluent.connect.jdbc.source;
 
-import io.confluent.connect.jdbc.dialect.DatabaseDialect;
-import io.confluent.connect.jdbc.util.ConnectionProvider;
-import io.confluent.connect.jdbc.util.TableId;
 import org.apache.kafka.connect.connector.ConnectorContext;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
@@ -33,6 +30,10 @@ import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+
+import io.confluent.connect.jdbc.dialect.DatabaseDialect;
+import io.confluent.connect.jdbc.util.ConnectionProvider;
+import io.confluent.connect.jdbc.util.TableId;
 
 /**
  * Thread that monitors the database for changes to the set of tables in the database that this
--- a/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
@@ -16,6 +16,9 @@
 
 package io.confluent.connect.jdbc.source;
 
+import io.confluent.connect.jdbc.dialect.DatabaseDialect;
+import io.confluent.connect.jdbc.util.ConnectionProvider;
+import io.confluent.connect.jdbc.util.TableId;
 import org.apache.kafka.connect.connector.ConnectorContext;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
@@ -25,13 +28,11 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-
-import io.confluent.connect.jdbc.dialect.DatabaseDialect;
-import io.confluent.connect.jdbc.util.ConnectionProvider;
-import io.confluent.connect.jdbc.util.TableId;
+import java.util.stream.Collectors;
 
 /**
  * Thread that monitors the database for changes to the set of tables in the database that this
@@ -48,6 +49,7 @@ public class TableMonitorThread extends Thread {
   private Set<String> whitelist;
   private Set<String> blacklist;
   private List<TableId> tables;
+  private final String configErrorMsg;
 
   public TableMonitorThread(DatabaseDialect dialect,
       ConnectionProvider connectionProvider,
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -150,6 +162,24 @@ private synchronized boolean updateTables() {
...
+    if (!duplicates.isEmpty()) {
+      log.info("The filtered tables are {} ", filteredTables);
+      log.error("Connector uses unqualified table name as the topic name and has detected "
+          + "duplicate unqualified table names. This could lead to mixed data type and "
+          + "subsequently errors during the processing. Hence to prevent processing errors, "
+          + "JDBC Source connector fails fast fr such configurations. To be able to start the "
+          + "connector, update connector's {} config to include exactly one table in each of the "
+          + "below listed tables.\n {} ", configErrorMsg, duplicates);
+      // this will force the connector to fail configuration as it doesn't have any table to process
+      filteredTables.clear();
```
rhauch: That makes sense, but the actual failure message will be "Number of groups must be positive", which is not going to be intuitive. How about if we set a flag here (perhaps the duplicate tableIDs), and then throw this error message in the `tables()` method. Both `updateTables()` and `tables()` are synchronized, so there shouldn't be any atomicity or visibility problems.

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -150,6 +162,24 @@ private synchronized boolean updateTables() {
...
+        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+    if (!duplicates.isEmpty()) {
+      log.info("The filtered tables are {} ", filteredTables);
+      log.error("Connector uses unqualified table name as the topic name and has detected "
+          + "duplicate unqualified table names. This could lead to mixed data type and "
+          + "subsequently errors during the processing. Hence to prevent processing errors, "
+          + "JDBC Source connector fails fast fr such configurations. To be able to start the "
+          + "connector, update connector's {} config to include exactly one table in each of the "
+          + "below listed tables.\n {} ", configErrorMsg, duplicates);
```
rhauch: The grammar in this error message needs to be corrected.

> "Duplicate table names found. This connector uses unqualified table names for topic names, and is currently configured with duplicate unqualified table names. Continuing would result in messages from these duplicate tables going to the same topic, and this could lead to mixed records, incompatible schemas, and potential problems in downstream consumers. This connector is stopping to avoid such problems. Change the connector's {} configuration to include exactly one table in each of the tables listed below:\n {}"

src/test/java/io/confluent/connect/jdbc/source/TableMonitorThreadTest.java
```diff
@@ -198,6 +278,19 @@ public void execute() {
...
+    return new Op() {
+      @Override
+      public void execute() {
+        List<TableId> expectedTableIds = new ArrayList<>();
+        for (TableId expectedTable: expectedTables) {
+          expectedTableIds.add(expectedTable);
+        }
+        assertEquals(expectedTableIds, tableMonitorThread.tables());
+      }
+    };
```
rhauch: Could use a lambda here:
```
protected Op checkTableIds(final TableId...expectedTables) {
   return ()-> {
        List<TableId> expectedTableIds = Arrays.asList(expectedTables);
        assertEquals(expectedTableIds, tableMonitorThread.tables());
   };
```
Also, this check also verifies the IDs are in the same order. Is that sufficient?

Finally, why does this return a lamba in the first place? Every place this method is called, you're immediately calling `execute()`. Why not just use a helper method that performs the assertion?

```
protected void checkTableIds(final TableId...expectedTables) {
    List<TableId> expectedTableIds = Arrays.asList(expectedTables);
    assertEquals(expectedTableIds, tableMonitorThread.tables());
}
```

mageshn:
@rhauch I have incorporated the comments. Would you like to take another pass at it?

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -105,6 +124,10 @@ public void run() {
     if (tables == null) {
       throw new ConnectException("Tables could not be updated quickly enough.");
     }
+    if (!duplicates.isEmpty()) {
+      log.info("The filtered tables are {} ", tables);
+      throw new ConnectException(configErrorMsg + duplicates);
```
rhauch: What do you think about printing the pairs of TableIds (e.g., `a.b.Foo / a.c.Foo`) rather than just dumping the map. I'm worried that people won't know what the map output means at the end of the long error message.

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -105,6 +124,10 @@ public void run() {
     if (tables == null) {
       throw new ConnectException("Tables could not be updated quickly enough.");
     }
+    if (!duplicates.isEmpty()) {
+      log.info("The filtered tables are {} ", tables);
```
rhauch: Should this go outside of the `if` block? Seems useful whether or not there are duplicates. Thoughts?

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
mageshn: could use duplicates.values()

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
mageshn: Its already logged in updateTables in debug. Maybe we should change that to info. Seems, pretty useful info to me.

rhauch: I agree its useful to have INFO, but shouldn't it happen every time? IOW, I thought you were going to remove this log message and change the one in `updateTables` to INFO?

rhauch: I agree its useful to have INFO, but shouldn't it happen every time? IOW, I thought you were going to remove this log message and change the one in updateTables to INFO?

mageshn:
it does appear here https://github.com/confluentinc/kafka-connect-jdbc/pull/438/commits/e772f155c7f1d7a045e04f0712f015212b0e8b5a

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -105,6 +124,9 @@ public void run() {
     if (tables == null) {
       throw new ConnectException("Tables could not be updated quickly enough.");
     }
+    if (!duplicates.isEmpty()) {
+      throw new ConnectException(configErrorMsg + duplicates.values());
```
rhauch: Can we not just compute `configErrorMsg` inside this block? Why do we want/need to precompute it in the constructor? We still need to hold onto the `duplicates` in a field, but this error would only be thrown once by the `run` method. I'd suggest moving that code into this block to eliminate the `configErrorMsg` field and instead use a local variable.

PR #429, opened 2018-06-14
Add function to query with initial timestamp.

daigorowhite:
In case of large table and not needed all record, or user just need recent record and realtime data from now, initial query is required with specific timestamp. but currently, initial query uses `Timestamp(0L)`.

In this PR , `kafka-connect-jdbc`'s `source` have new config which is `timestamp.initial`.
if you specify `timestamp.initial`, it works like this. 

- `0` : `1970-01-01 00:00:00.000` (Same as existed)
- `-1` : Current time will be used 
- `1528858322371` : `2018-06-13 11:52:02.371` ( UnixTime milliseconds )

and default is 0.

In our case, we need this function, could you have a look, please?

PR #441, opened 2018-07-06
set the offset manually

baifachuan:
Hi:
I have updated the `JdbcSourceConnectorConfig` and `JdbcSourceTask` class to parse the offset from parameter.

Now I can set the offset manually, I have an existing table about 20GB data, Using incrementing mode, but when I start the JDBC connector will copy the history data it takes a lot of time. 

So I have updated the code, I believe some people have the same problem.

thanks


baifachuan:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -185,6 +185,24 @@ public void start(Map<String, String> properties) {
...
+        offset.put("incrementing", incrementingOffset);
+      }
+
+      if (timestampOffset != null) {
+        if (offset == null) {
+          offset = new HashMap<>();
+        }
+        offset.put("timestamp", timestampOffset);
+      }
+
```
rhauch: Won't this always set the offset values to what is read from the config? So, if the connector task restarts (which will happen during a rebalance), if the `incrementing.offset` and/or `timestamp.offset` config properties are still set, then these lines will always use those values to update the offsets, and the connector task will essentially back up to this point. If it restarts again, it will again back up to this same exact point.

Also, the supplied `incrementing.offset` and `timestamp.offset` configuration values will be applied to all tables. Is that really what you want?

rhauch:
See also #429

PR #463, opened 2018-08-09
Fix values(CURRENT_TIMESTAMP) not working on DB2 because invalid sql-…

leubedane:
…statement. Replace with select-statement of CURRENT_TIMESTAMP

See more information on https://github.com/confluentinc/kafka-connect-jdbc/issues/211

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
@@ -62,7 +62,7 @@ public class Db2DatabaseDialect extends GenericDatabaseDialect {
 
   @Override
   protected String currentTimestampDatabaseQuery() {
-    return "values(CURRENT_TIMESTAMP)";
+    return "SELECT CURRENT_TIMESTAMP FROM SYSIBM.SYSDUMMY1;";
   }
 
   @Override


src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
```diff
@@ -62,7 +62,7 @@ public Db2DatabaseDialect(AbstractConfig config) {
 
   @Override
   protected String currentTimestampDatabaseQuery() {
-    return "values(CURRENT_TIMESTAMP)";
+    return "SELECT CURRENT_TIMESTAMP FROM SYSIBM.SYSDUMMY1;";
```
rhauch: IIUC, `CURRENT_TIMESTAMP` is equivalent to `CURRENT TIMESTAMP(6)`, which might be for example `2015-03-23-09.41.24.684842` ([ref](https://www.ibm.com/support/knowledgecenter/en/SS6NHC/com.ibm.swg.im.dashdb.sql.ref.doc/doc/r0005886.html)). Is that correct? Is this the precision we want to use?

rhauch:
Fixes #211

teebee:
This PR would fix CC-2455 for the master branch. I would appreciate if it will also be back ported to versions confluentinc:4.0.x (branch 4.0.x) and confluentinc:4.1.x (branch 4.1.x).

rhauch:
@leubedane, could you change the version on this PR to `5.0.x`, since that's the first version that used the new dialects?

@teebee, we'd need to open a different PR for branches earlier than `5.0.x`. In fact, we could go back as far as `3.0.x`: https://github.com/confluentinc/kafka-connect-jdbc/blob/3.0.x/src/main/java/io/confluent/connect/jdbc/JdbcUtils.java#L213; see the comment in the code at https://github.com/confluentinc/kafka-connect-jdbc/blob/3.0.x/src/main/java/io/confluent/connect/jdbc/JdbcUtils.java#L209

rhauch:
@leubedane wrote:

> Hi @rhauch do you think CURRENT_TIMESTAMP(3) --> 2015-03-23-09.41.24.684 is the right solution? The problem could be an offset between request and db server. So we are using millis instead of nanos.

Well, we need to test this, but using just millis if the timestamp column has greater precision might actually cause repeats. We've had a similar problem with other DBMSes due to casting and loss of precision between the client and server. It'd probably be better to have the connector retain the most precision possible.

leubedane:
@rhauch 
If i refer to your [ref](https://www.ibm.com/support/knowledgecenter/en/SS6NHC/com.ibm.swg.im.dashdb.sql.ref.doc/doc/r0005886.html), the ibm value works with nanos as maximum precision. So we can use CURRENT TIMESTAMP(12), but the values behind the nanos were alway zero if I understand the ibm site right. So If we use TIMESTAMP(6) we have the maximum available precision for ibm db2. Am I right with this?

teebee:
@leubedane The portion that denotes fractional seconds in timestamps can have up to 12 digits in Db2 10 for z/OS, which means that nanoseconds and even picoseconds can be represented: [New features in Db2 10 for z/OS](https://www.ibm.com/support/knowledgecenter/en/SS4LQ8_4.1.0/com.ibm.datatools.aqt.doc/gui/concepts/c_idaa_db2_v10_new_features.html).

The Java [`java.sql.Timestamp`](https://docs.oracle.com/javase/10/docs/api/java/sql/Timestamp.html) allows the specification of fractional seconds to a precision of nanoseconds.
However, in [`io.confluent.connect.jdbc.source.TimestampIncrementingTableQuerier`](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java#L196) the `java.sql.Timestamp` is converted into a `long`, such that the precision should get lost:

      @Override
      public Timestamp endTimetampValue()  throws SQLException {
        final long currentDbTime = dialect.currentTimeOnDB(
            stmt.getConnection(),
            DateTimeUtils.UTC_CALENDAR.get()
        ).getTime();
        return new Timestamp(currentDbTime - timestampDelay);
      }

Am I missing something here?! If not, millisecond precision is indeed sufficient unless `endTimetampValue()` - note the typo - will also be fixed.

leubedane:
I created a new pull request for 5.0.x version here https://github.com/confluentinc/kafka-connect-jdbc/pull/479

I used the CURRENT_TIMESTAMP(12) precision because it will fit all cases

rhauch:
Replaced by #479

hotcoder:
I am also facing the issue , i have posted question on this , how to resolve this issue ?

https://stackoverflow.com/questions/56218720/getting-exception-failed-to-get-current-time-from-db-using-query-valuescurrent

Issue #34, opened 2016-01-08
java.lang.OutOfMemoryError


bchazalet:
I running kafka-connect jdbc on one table, for the first time, and I am quickly running out of memory even though I have set:

```
batch.max.rows = 1
tasks.max = 1
```

It looks like the batch limit is not being effective in my case (I am running it against a postgres database). 

Does the batch limit apply the very first time kafka-connect runs?

ewencp:
`batch.max.rows` controls how much memory is used to buffer data by the connector itself during a call to its `poll()` method. What it does is generate and run a query, but if more than `batch.max.rows` are generated, it will stop processing after that many and return from `poll()`. This guarantees that if you have a large number of rows, it doesn't try to process them all at once.

However, this doesn't do anything like adding a `LIMIT` clause to the query. Perhaps we need an option to control that as well (or reuse `batch.max.rows`, although whether that can be done safely depends on the query mode you're using). However, I wouldn't have expected this to cause an out of memory error as the results should just stream in from the DB.

What database and JDBC driver are you using, and could you share anything about your connector configuration/use case that might shed any more light on how it is hitting the OOM issue?

bchazalet:
I am using postgresq 9.4 and the driver is the one from the confluent 2.0.0 package, i.e `postgresql-9.4-1206-jdbc41.jar`.

About the use case, I am using the jdbc connector against an existing table in the database, for the first time. The table has got around 10 millions rows, even though I rather care about the future updates rather than the data already in it.

I am using the `timestamp` mode.

Leaving the sensitive information out, here's my properties file:
```
name=test
connector.class=io.confluent.connect.jdbc.JdbcSourceConnector
tasks.max=10
batch.max.rows=100
mode=timestamp
timestamp.column.name=modified_timestamp
table.whitelist=mytable
connection.url=jdbc:postgresql://**********
```

bchazalet:
I think the problems lies with the streaming. 

We have internal projects where we are streaming the data (from postgres as well) and we had to have `autoCommit` set to false on the connection, as well as setting the `FetchDirection` to `FETCH_FORWARD` and `FetchSize` to some value on the prepared statement:

I don't know how portable that is across databases though, and if you're already doing something similar.

SleepyBrett:
Having the exact same issue against postgres (PostgreSQL) 9.5.0 using the driver in the confluent 2.0 platform.

kadrach:
[...]

PR #466, opened 2018-08-15
Fix OOM errors in PostgreSQL dialect

hagmonk:
Partial fix for #34, since that issue seems to cover more than just PostgreSQL.

Per multiple contributors to that issue, the changes required are:

- ensure autoCommit is set to false on the Connection
- ensure fetchSize is set to a non-zero value on the PreparedStatement

In this patch I re-use the "batch.max.rows" property from JdbcSourceConnectorConfig as the value for fetchSize. The rationale is that the documented effect of that field is almost identical to the
documented effect of calling setFetchSize, i.e. it limits the number of rows buffered internally. See
https://jdbc.postgresql.org/documentation/head/query.html#fetchsize-example.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -16,6 +16,7 @@
 
 package io.confluent.connect.jdbc.dialect;
 
+import io.confluent.connect.jdbc.source.JdbcSourceTaskConfig;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
@@ -72,6 +73,13 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
     super(config, new IdentifierRules(".", "\"", "\""));
   }
 
+  @Override
+  public Connection getConnection() throws SQLException {
+    Connection connection = super.getConnection();
+    connection.setAutoCommit(false);
+    return connection;
+  }
+
   /**
    * Perform any operations on a {@link PreparedStatement} before it is used. This is called from
    * the {@link #createPreparedStatement(Connection, String)} method after the statement is
@@ -86,8 +94,10 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
    */
   @Override
   protected void initializePreparedStatement(PreparedStatement stmt) throws SQLException {
+    int batchMaxRows = config.getInt(JdbcSourceTaskConfig.BATCH_MAX_ROWS_CONFIG);
     log.trace("Initializing PreparedStatement fetch direction to FETCH_FORWARD for '{}'", stmt);
     stmt.setFetchDirection(ResultSet.FETCH_FORWARD);
+    stmt.setFetchSize(batchMaxRows);
   }
 
 


hagmonk:
[clabot:check]

hagmonk:
An assumption I should have documented: from reading the code it appears that `io.confluent.connect.jdbc.sink.JdbcDbWriter` also sets the connection to autoCommit false. This is why I haven't conditionalized the autoCommit change in `PostgreSqlDatabaseDialect` to only apply in the case where it's a source. Hopefully that isn't a mistaken assumption.

mageshn:
@hagmonk thanks a lot for your fix. This LGTM.

hagmonk:
@cramja really good catch, I didn't notice this happening in my databases because I'm actually only copy two tables (albeit quite large ones).

I just spent a few minutes going through the code again. This is an ugly problem, because it appears auto commit can only be disabled at the connection level. `TableMonitorThread` grabs this connection and so will always be affected if it shares the same connection. Additionally, when `getMetaData` is called on the connection directly it goes off into JDBC land, so we have no hook on which to hang your proxy for those situations.

Additionally, the source and sink tasks all have a direct handle on the JDBC connection, which they can use for any purpose they wish. The dialect abstraction is really nice, but it doesn't offer an obvious way to encapsulate things such that this is neatly solvable at the `PostgreSqlDatabaseDialect` level.

@mageshn what do you think of this issue? I might be missing something here, since there are many layers involved. The goal is that we have auto commit *disabled* for the actual bulk reads from source tables. It should stay enabled for everything else, per @cramja's observation that we will otherwise cause pretty bad problems for PostgreSQL.

In terms of the impact of those bad problems, my (non-expert) understanding is:

*Long-held catalog read locks* - this will cause a hang for anything trying to query the catalog.

*Long idle in transaction* - this will prevent vacuum from cleaning up dead tuples, leading to table bloat amongst other problems.

On the other hand, without the ability to disable auto commit, kafka-connect-jdbc will remain totally useless for anyone attempting to use a larger-than-memory PostgreSQL table as a Kafka Connect source :) 

Fun times!

cramja:
@hagmonk I appreciate the in depth response. I can add a little more color to what is happening. 

Setting the `fetchSize` param and disabling autocommit causes the PgConnection to use cursors (see [getting results with cursor](https://jdbc.postgresql.org/documentation/head/query.html)). The original OOM error which you fixed was caused because the default for the connection is to buffer the entire result table in the ResultSet and commit the transaction as soon as the query is finished. The cursor is the correct way to get all the results in sane way. Potentially, you could use LIMIT but that seems to be more difficult than closing the cursor / committing. 

The reason I disabled the TableMonitor was for exactly the reason you point out - there wasn't a quick/clean way to call commit on the connection object in a way which was db-agnostic.

A consequence of having long read locks on the catalog is that you cannot do catalog changes (we first noticed this when trying to add a column).

I may look into a cleaner way to inform the dialect of the stages of the query so that it could handle cleanup/commit, but for now the Proxy and disabling the TableMonitor have worked as an interim solution.

hagmonk:
@cramja I'm trying to think of ways to avoid having to disable the TableMonitor. Two ideas so far:

1) disable autocommit, create the prepared statement, re-enable autocommit. Obviously there's a possible race there, and I haven't a strong grasp on how concurrency is handled inside `PGConnection`. It might also be the case that autocommit must be disabled at all times.

2) try finding a way to use [`execSQLQuery`](https://github.com/pgjdbc/pgjdbc/blob/08631ccdabdb8ba6d52f398e2b0b46a9cf0cafbf/pgjdbc/src/main/java/org/postgresql/core/BaseConnection.java#L37). I'm not sure this will work because I believe, per the pgjdbc docs, the fetchSize must also be set, and one sets that on a `PreparedStatement`.

Still, I'll do some reading and see if we are missing some tricks …

hagmonk:
Hmm, [`execSQLQuery`](https://github.com/pgjdbc/pgjdbc/blob/08631ccdabdb8ba6d52f398e2b0b46a9cf0cafbf/pgjdbc/src/main/java/org/postgresql/jdbc/PgConnection.java#L414):

```java
    boolean hasResultSet = stat.executeWithFlags(s, QueryExecutor.QUERY_SUPPRESS_BEGIN);
```

cramja:
> QUERY_SUPPRESS_BEGIN ...  override close to be a no-op

I'm not exactly sure what you would be trying to accomplish with this sequence. Wouldn't we need to start a transaction (BEGIN) to use a cursor?

> I'm still a bit hazy on the cleanup – we still need to have COMMIT sent to the connection once the ResultSet is exhausted, correct?

From what I've seen with just experimenting with the Postgres Jdbc library, this is the case (or closing the connection, but that's undesirable). 

_______

I think that the key here is to force the PreparedStatement to produce a ResultSet which is implemented with a cursor but also which commits / closes the cursor when either the ResultSet / PStatement is closed. That's why I used the proxy. 

The tricky part is that the Connection objects contain the autoCommit setting and are shared between the TableMonitor and Querier via the CachedConnectionProvider. So we can't just set autocommit to false without effecting the TableMonitor. This is why I disabled the table monitor. Though, we probably could just set autoCommit to "true" after the connection is committed to "reset" the connection for the TableMonitor, should it get the connection...

I looked over at the code again, and I'm not sure there's a good way to do this without isolating to connections which query the table's schema from the ones that do the bulk querying.

rhauch:
@hagmonk, the `SybaseDatabaseDialect` uses a different technique to [always disable autocommit for the metadata queries](https://github.com/confluentinc/kafka-connect-jdbc/blob/7e37a636f772b9eb90b14eeba614f2830548d485/src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java#L161-L195). Have you thought about doing something similar for the PostgreSQL dialect? It's true that you can only enable/disable autocommit for the entire connection, but Connect and connectors are single-threaded, and thus each connection is only used for one query at a time.

gharris1727:
This issue has been fixed in #758.

Thank you @cramja and @hagmonk for your Postgres knowledge, and exploration of this issue, it was very helpful when assembling the final fix.

PR #479, opened 2018-08-31
Fix values(CURRENT_TIMESTAMP) not working on DB2 because invalid sql-…

leubedane:
…statement. Replace with select-statement of CURRENT_TIMESTAMP(12)

See more on https://github.com/confluentinc/kafka-connect-jdbc/pull/463


leubedane:
@rhauch : how can we proceed with this issue? Are there any questions left so this solution doesnt work? I really want this feature for the next release if it is possible.

PR #350, opened 2018-01-22
Dev fork

math-g:
Hi,
This is a proposition relative to issues [#113](https://github.com/confluentinc/kafka-connect-jdbc/issues/113) and [#103](https://github.com/confluentinc/kafka-connect-jdbc/issues/103) that I faced on my project.
It gives good results on the client setup (large Oracle DB with a lot of views needed, no timezone in timestamps)
See Readme.md for a complete description.

I tried to adhere to existing basecode to add these features cleanly, even if some refactoring would now certainly be necessary (too much arguments in TimestampIncrementingTableQuerier for instance).
Also, a better solution would need to close previously [KAFKA-4524](https://issues.apache.org/jira/browse/KAFKA-4524)

As written in Readme, I added some unit tests
Best

Issue #174, opened 2016-12-05
JDBC Sink should specify UTC Calendar when binding date/time types


shikhar:
We should specify a UTC `Calendar` instance when binding `Date`/`Time`/`Timestamp` in https://github.com/confluentinc/kafka-connect-jdbc/blob/22c105653fb354b37f4094f0aff6f513f135e9c7/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java#L161-L181

Issue #103, opened 2016-07-27
Can timezone be specified for mode = timestamp?


rmoff:
My DB system clock is on BST (UTC +1). If I use `mode = timestamp` and insert data to a table (Oracle) as : 

```
SQL> create table kafka_test (ts1 timestamp not null,foo int);

Table created.

SQL> insert into kafka_test select sysdate-1,42 from dual;

1 row created.

SQL> commit;

Commit complete.

SQL> select * from kafka_test;

TS1                             FOO
-------------------------       ----------
27-JUL-16 05.39.45.000000 PM    42
```

Kafka Connect won't pick up that row until an hour from now, as it is using UTC in its timestamp, even though it returns the `CURRENT_TIMESTAMP` from the Oracle box. 

```
[2016-07-27 17:39:01,923] DEBUG Executing prepared statement with timestamp value = 1469554131000 (2016-07-26 17:28:51.000)  end time 2016-07-27 16:39:01.923 (io.confluent.connect.jdbc.TimestampIncrementingTableQuerier:171)
```

Is there a way to specify to kafka-connect-jdbc which timezone it should be using? 

thanks.

(connector config for reference: )

```
name=oracle-soe-incr-source
connector.class=io.confluent.connect.jdbc.JdbcSourceConnector
tasks.max=10


connection.url=jdbc:oracle:thin:soe/soe@db12-01.moffatt.me:1521/pdb
topic.prefix=ora-soe-
mode=timestamp
timestamp.column.name=TS1
table.whitelist=KAFKA_TEST
```

shikhar:
@rmoff does it work to specify a function to convert to UTC as the `timestamp.column.name`?

seems like for Oracle this is `sys_extract_utc`

jorgheymans:
confirming that sys_extract_utc works on Oracle, ran into this problem as well. 

I think it's worth mentioning in the documentation that the timestamp mode works with UTC timestamps.

Sarvesh-D:
@jorgheymans hi, i too am not able to use timestamp mode properly with oracle DB. can you tell me what you changed in the connector config that made it work for you. it would be great if you can paste the working connector config here. Thanks.

jorgheymans:
in the DDL we defined the timestamp like below:

```
EVENT_TIMESTAMP TIMESTAMP(6) default SYS_EXTRACT_UTC(CURRENT_TIMESTAMP) not null,
```

rrangine:
[...]

PR #505, opened 2018-10-15
Added db.timezone setting 

mariusneo:
Added db.timezone setting in order to be able to customize the timezone (to a different value than the default value UTC) used at the jdbc level.

When customizing this setting, the  date/time/timestamp values read from the source database will be instantiated with the customized time zone.

The timestamp  1474661402123L  corresponds to `2016-09-23 20:10:02.123` in UTC zone.
When the database timezone is not UTC, but Europe/Vienna, the value  read from the database should be `2016-09-23 22:10:02.123`.

This pull requests addresses exactly the problem mentioned in the example above, by allowing the user of kafka-jdbc connector to customize the timezone in which the date/time/timestamp values are read from the database.
This is being done by adding a new `timezone` field to io.confluent.connect.jdbc.dialect.GenericDatabaseDialect class which is to be used when reading/writing date/time/timestamp values.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <derby.version>10.11.1.1</derby.version>
         <commons-io.version>2.4</commons-io.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
-        <sqlite-jdbc.version>3.8.11.2</sqlite-jdbc.version>
+        <sqlite-jdbc.version>3.25.2</sqlite-jdbc.version>
         <postgresql.version>9.4-1206-jdbc41</postgresql.version>
         <jtds.driver.version>1.3.1</jtds.driver.version>
         <licenses.name>Apache License 2.0</licenses.name>
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <derby.version>10.11.1.1</derby.version>
         <commons-io.version>2.4</commons-io.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
-        <sqlite-jdbc.version>3.25.2</sqlite-jdbc.version>
+        <sqlite-jdbc.version>3.8.11.2</sqlite-jdbc.version>
         <postgresql.version>9.4-1206-jdbc41</postgresql.version>
         <jtds.driver.version>1.3.1</jtds.driver.version>
         <licenses.name>Apache License 2.0</licenses.name>
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -16,6 +16,11 @@
 
 package io.confluent.connect.jdbc.sink;
 
+import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
+import io.confluent.connect.jdbc.util.DatabaseDialectRecommender;
+import io.confluent.connect.jdbc.util.StringUtils;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -23,11 +28,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
-import io.confluent.connect.jdbc.util.DatabaseDialectRecommender;
-import io.confluent.connect.jdbc.util.StringUtils;
-
+import java.util.TimeZone;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
@@ -187,6 +188,13 @@ public class JdbcSinkConfig extends AbstractConfig {
       + "specific dialect. All properly-packaged dialects in the JDBC connector plugin "
       + "can be used.";
 
+  public static final String DB_TIMEZONE_CONFIG = "db.timezone";
+  public static final String DB_TIMEZONE_DEFAULT = "UTC";
+  private static final String DB_TIMEZONE_CONFIG_DOC =
+      "Alternative time zone of the database, to be used by JDBC driver instead of UTC (default)"
+          + "when instantiating PreparedStatements.";
+  private static final String DB_TIMEZONE_CONFIG_DISPLAY = "DB time zone";
+
   public static final ConfigDef CONFIG_DEF = new ConfigDef()
         // Connection
         .define(
@@ -303,6 +311,16 @@ public class JdbcSinkConfig extends AbstractConfig {
             4,
             ConfigDef.Width.LONG,
             FIELDS_WHITELIST_DISPLAY
+        ).define(
+          DB_TIMEZONE_CONFIG,
+          ConfigDef.Type.STRING,
+          DB_TIMEZONE_DEFAULT,
+          ConfigDef.Importance.HIGH,
+          DB_TIMEZONE_CONFIG_DOC,
+          DATAMAPPING_GROUP,
+          5,
+          ConfigDef.Width.MEDIUM,
+          DB_TIMEZONE_CONFIG_DISPLAY
         )
         // DDL
         .define(
@@ -365,6 +383,7 @@ public class JdbcSinkConfig extends AbstractConfig {
   public final List<String> pkFields;
   public final Set<String> fieldsWhitelist;
   public final String dialectName;
+  public final TimeZone timeZone;
 
   public JdbcSinkConfig(Map<?, ?> props) {
     super(CONFIG_DEF, props);
@@ -382,6 +401,9 @@ public class JdbcSinkConfig extends AbstractConfig {
     pkFields = getList(PK_FIELDS);
     dialectName = getString(DIALECT_NAME_CONFIG);
     fieldsWhitelist = new HashSet<>(getList(FIELDS_WHITELIST));
+    String dbTimeZone = getString(DB_TIMEZONE_CONFIG);
+    timeZone = dbTimeZone == null ? TimeZone.getTimeZone(ZoneOffset.UTC) :
+        TimeZone.getTimeZone(ZoneId.of(dbTimeZone));
   }
 
   private String getPasswordValue(String key) {
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -193,8 +193,8 @@ public class JdbcSinkConfig extends AbstractConfig {
   public static final String DB_TIMEZONE_CONFIG = "db.timezone";
   public static final String DB_TIMEZONE_DEFAULT = "UTC";
   private static final String DB_TIMEZONE_CONFIG_DOC =
-      "Name of the timezone that should be used in the connector when "
-      + "specifying time-based criteria. Defaults to UTC.";
+      "Name of the JDBC timezone that should be used in the connector when "
+      + "inserting time-based values. Defaults to UTC.";
   private static final String DB_TIMEZONE_CONFIG_DISPLAY = "DB Time Zone";
 
   public static final ConfigDef CONFIG_DEF = new ConfigDef()
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -16,11 +16,7 @@
 
 package io.confluent.connect.jdbc.sink;
 
-import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
-import io.confluent.connect.jdbc.util.DatabaseDialectRecommender;
-import io.confluent.connect.jdbc.util.StringUtils;
 import java.time.ZoneId;
-import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -29,6 +25,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
+
+import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
+import io.confluent.connect.jdbc.util.DatabaseDialectRecommender;
+import io.confluent.connect.jdbc.util.StringUtils;
+import io.confluent.connect.jdbc.util.TimeZoneValidator;
+
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
@@ -191,9 +193,9 @@ public class JdbcSinkConfig extends AbstractConfig {
   public static final String DB_TIMEZONE_CONFIG = "db.timezone";
   public static final String DB_TIMEZONE_DEFAULT = "UTC";
   private static final String DB_TIMEZONE_CONFIG_DOC =
-      "Alternative time zone of the database, to be used by JDBC driver instead of UTC (default)"
-          + "when instantiating PreparedStatements.";
-  private static final String DB_TIMEZONE_CONFIG_DISPLAY = "DB time zone";
+      "Name of the timezone that should be used in the connector when "
+      + "specifying time-based criteria. Defaults to UTC.";
+  private static final String DB_TIMEZONE_CONFIG_DISPLAY = "DB Time Zone";
 
   public static final ConfigDef CONFIG_DEF = new ConfigDef()
         // Connection
@@ -315,7 +317,8 @@ public class JdbcSinkConfig extends AbstractConfig {
           DB_TIMEZONE_CONFIG,
           ConfigDef.Type.STRING,
           DB_TIMEZONE_DEFAULT,
-          ConfigDef.Importance.HIGH,
+          TimeZoneValidator.INSTANCE,
+          ConfigDef.Importance.MEDIUM,
           DB_TIMEZONE_CONFIG_DOC,
           DATAMAPPING_GROUP,
           5,
@@ -402,8 +405,7 @@ public class JdbcSinkConfig extends AbstractConfig {
     dialectName = getString(DIALECT_NAME_CONFIG);
     fieldsWhitelist = new HashSet<>(getList(FIELDS_WHITELIST));
     String dbTimeZone = getString(DB_TIMEZONE_CONFIG);
-    timeZone = dbTimeZone == null ? TimeZone.getTimeZone(ZoneOffset.UTC) :
-        TimeZone.getTimeZone(ZoneId.of(dbTimeZone));
+    timeZone = TimeZone.getTimeZone(ZoneId.of(dbTimeZone));
   }
 
   private String getPasswordValue(String key) {
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -16,6 +16,9 @@
 
 package io.confluent.connect.jdbc.source;
 
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.util.TimeZone;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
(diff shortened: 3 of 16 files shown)


mariusneo:
This pull requests also relates to work perfomed on:
- https://github.com/confluentinc/kafka-connect-jdbc/pull/350 support for different timezones
- https://github.com/confluentinc/kafka-connect-jdbc/issues/174 adding UTC calendar when binding date/time/timestamp types
and also to the issues:

https://github.com/confluentinc/kafka-connect-jdbc/issues/103 Can timezone be specified for mode = timestamp?

rhauch:
@mariusneo I'll take a look at this over the next week or so

pom.xml
```diff
@@ -59,7 +59,7 @@
         <derby.version>10.11.1.1</derby.version>
         <commons-io.version>2.4</commons-io.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
-        <sqlite-jdbc.version>3.8.11.2</sqlite-jdbc.version>
+        <sqlite-jdbc.version>3.25.2</sqlite-jdbc.version>
```
rhauch: Why upgrade this dependency within this PR? Why not just issue a separate, much smaller/simpler PR?

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -16,18 +16,19 @@
...
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
-import io.confluent.connect.jdbc.util.DatabaseDialectRecommender;
-import io.confluent.connect.jdbc.util.StringUtils;
-
```
rhauch: It'd be better to not include these changes in your PR. Besides, our style is to include `java.*` imports first, so the original order was correct.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -187,6 +188,13 @@
       + "specific dialect. All properly-packaged dialects in the JDBC connector plugin "
       + "can be used.";
 
+  public static final String DB_TIMEZONE_CONFIG = "db.timezone";
+  public static final String DB_TIMEZONE_DEFAULT = "UTC";
+  private static final String DB_TIMEZONE_CONFIG_DOC =
+      "Alternative time zone of the database, to be used by JDBC driver instead of UTC (default)"
+          + "when instantiating PreparedStatements.";
+  private static final String DB_TIMEZONE_CONFIG_DISPLAY = "DB time zone";
```
rhauch:
```suggestion
  private static final String DB_TIMEZONE_CONFIG_DISPLAY = "DB Time Zone";
```

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -187,6 +188,13 @@
       + "specific dialect. All properly-packaged dialects in the JDBC connector plugin "
       + "can be used.";
 
+  public static final String DB_TIMEZONE_CONFIG = "db.timezone";
+  public static final String DB_TIMEZONE_DEFAULT = "UTC";
+  private static final String DB_TIMEZONE_CONFIG_DOC =
+      "Alternative time zone of the database, to be used by JDBC driver instead of UTC (default)"
+          + "when instantiating PreparedStatements.";
```
rhauch: Maybe something like:
```
Name of the timezone that should be used in the connector when specifying time-based criteria. Defaults to UTC.
```

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -303,6 +311,16 @@
             4,
             ConfigDef.Width.LONG,
             FIELDS_WHITELIST_DISPLAY
+        ).define(
+          DB_TIMEZONE_CONFIG,
+          ConfigDef.Type.STRING,
+          DB_TIMEZONE_DEFAULT,
+          ConfigDef.Importance.HIGH,
```
rhauch:
```suggestion
          ConfigDef.Importance.MEDIUM,
```

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -382,6 +401,9 @@ public JdbcSinkConfig(Map<?, ?> props) {
     pkFields = getList(PK_FIELDS);
     dialectName = getString(DIALECT_NAME_CONFIG);
     fieldsWhitelist = new HashSet<>(getList(FIELDS_WHITELIST));
+    String dbTimeZone = getString(DB_TIMEZONE_CONFIG);
+    timeZone = dbTimeZone == null ? TimeZone.getTimeZone(ZoneOffset.UTC) :
+        TimeZone.getTimeZone(ZoneId.of(dbTimeZone));
```
rhauch: The `getString(DB_TIMEZONE_CONFIG)` method will always return a non-null string, since the default value `is `UTC`. That means that lines 405-406 can be simplified to just:

    timeZone = TimeZone.getTimeZone(ZoneId.of(dbTimeZone))

Also, what happens if an invalid time zone is specified? We can prevent that by adding a Validator to the DB_TIMEZONE_CONFIG definition that checks whether the supplied value is a valid time zone name.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -303,6 +311,16 @@
...
+        ).define(
+          DB_TIMEZONE_CONFIG,
+          ConfigDef.Type.STRING,
+          DB_TIMEZONE_DEFAULT,
+          ConfigDef.Importance.HIGH,
+          DB_TIMEZONE_CONFIG_DOC,
+          DATAMAPPING_GROUP,
+          5,
+          ConfigDef.Width.MEDIUM,
+          DB_TIMEZONE_CONFIG_DISPLAY
```
rhauch: As mentioned below, we should provide a `ConfigDef.Validator` method that ensures that all values are indeed valid time zone names. This will help when configuring connectors, since an error message like `'{}' is not a valid name of a time zone` would appear in the context of `db.timezone`.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -537,7 +548,17 @@ private static final void addConnectorOptions(ConfigDef config) {
...
+    ).define(
+        DB_TIMEZONE_CONFIG,
+        Type.STRING,
+        DB_TIMEZONE_DEFAULT,
+        Importance.HIGH,
+        DB_TIMEZONE_CONFIG_DOC,
+        CONNECTOR_GROUP,
+        ++orderInGroup,
+        Width.MEDIUM,
+        DB_TIMEZONE_CONFIG_DISPLAY);
```
rhauch: Need a validator hear, like mentioned above for JdbcSinkConnectorConfig.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -760,6 +781,12 @@ public NumericMapping numericMapping() {
     return NumericMapping.get(this);
   }
 
+  public TimeZone timeZone() {
+    String dbTimeZone = getString(JdbcSourceTaskConfig.DB_TIMEZONE_CONFIG);
+    return dbTimeZone == null ? TimeZone.getTimeZone(ZoneOffset.UTC) :
+        TimeZone.getTimeZone(ZoneId.of(dbTimeZone));
```
rhauch: Same comment as above: the default plus a validator means that we should always have a valid `dbTimeZone` string that will always correspond to a valid time zone, and the code could be simplified to:

    String dbTimeZone = getString(JdbcSourceTaskConfig.DB_TIMEZONE_CONFIG);
    return TimeZone.getTimeZone(ZoneId.of(dbTimeZone));

src/test/java/io/confluent/connect/jdbc/dialect/BaseDialectTest.java
```diff
@@ -313,11 +314,16 @@ public void bindFieldPrimitiveValues() throws SQLException {
...
     verifyBindField(++index, Schema.BYTES_SCHEMA, ByteBuffer.wrap(new byte[]{42})).setBytes(index, new byte[]{42});
     verifyBindField(++index, Schema.STRING_SCHEMA, "yep").setString(index, "yep");
-    verifyBindField(++index, Decimal.schema(0), new BigDecimal("1.5").setScale(0, BigDecimal.ROUND_HALF_EVEN)).setBigDecimal(index, new BigDecimal(2));
-    verifyBindField(++index, Date.SCHEMA, new java.util.Date(0)).setDate(index, new java.sql.Date
-        (0), DateTimeUtils.UTC_CALENDAR.get());
-    verifyBindField(++index, Time.SCHEMA, new java.util.Date(1000)).setTime(index, new java.sql.Time(1000), DateTimeUtils.UTC_CALENDAR.get());
-    verifyBindField(++index, Timestamp.SCHEMA, new java.util.Date(100)).setTimestamp(index, new java.sql.Timestamp(100), DateTimeUtils.UTC_CALENDAR.get());
+    verifyBindField(++index, Decimal.schema(0),
+            new BigDecimal("1.5").setScale(0, BigDecimal.ROUND_HALF_EVEN))
+            .setBigDecimal(index, new BigDecimal(2));
```
rhauch: Our style is a bit strange to those not familiar with it, but these lines would be formatted:
```
    verifyBindField(
      ++index, 
      Decimal.schema(0),
      new BigDecimal("1.5").setScale(0, BigDecimal.ROUND_HALF_EVEN)
    ).setBigDecimal(index, new BigDecimal(2));
```

src/test/java/io/confluent/connect/jdbc/dialect/BaseDialectTest.java
```diff
@@ -313,11 +314,16 @@ public void bindFieldPrimitiveValues() throws SQLException {
...
+    Calendar utcCalendar = DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC));
+    verifyBindField(++index, Date.SCHEMA, new java.util.Date(0)).setDate(index,
+        new java.sql.Date(0), utcCalendar);
+    verifyBindField(++index, Time.SCHEMA, new java.util.Date(1000))
+        .setTime(index, new java.sql.Time(1000), utcCalendar);
+    verifyBindField(++index, Timestamp.SCHEMA, new java.util.Date(100))
+        .setTimestamp(index, new java.sql.Timestamp(100), utcCalendar);
```
rhauch: Again, the formatting of these lines should either be on line line or:
```
    verifyBindField(
      ++index, 
      Timestamp.SCHEMA, 
      new java.util.Date(100)
    ).setTimestamp(index, new java.sql.Timestamp(100), utcCalendar);
```

src/test/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialectTest.java
```diff
@@ -237,9 +240,12 @@ public void bindFieldPrimitiveValues() throws SQLException {
...
-        (0), DateTimeUtils.UTC_CALENDAR.get());
-    verifyBindField(++index, Time.SCHEMA, new java.util.Date(1000)).setTime(index, new java.sql.Time(1000), DateTimeUtils.UTC_CALENDAR.get());
-    verifyBindField(++index, Timestamp.SCHEMA, new java.util.Date(100)).setTimestamp(index, new java.sql.Timestamp(100), DateTimeUtils.UTC_CALENDAR.get());
+    Calendar utcCalendar = DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC));
+    verifyBindField(++index, Date.SCHEMA, new java.util.Date(0))
+        .setDate(index, new java.sql.Date(0), utcCalendar);
+    verifyBindField(++index, Time.SCHEMA, new java.util.Date(1000))
+        .setTime(index, new java.sql.Time(1000), utcCalendar);
+    verifyBindField(++index, Timestamp.SCHEMA, new java.util.Date(100))
+        .setTimestamp(index, new java.sql.Timestamp(100), utcCalendar);
```
rhauch: Formatting, as mentioned above.

pom.xml
mariusneo: @rhauch you are definitely right here.
I've created a different pull request to address this change.

https://github.com/confluentinc/kafka-connect-jdbc/pull/508

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
mariusneo: :+1:

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
mariusneo: :+1:

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
mariusneo: :+1:

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
mariusneo: you're right for simplifying the retrieval of the timezone. thanks for the hint.

regarding an invalid timezone i'll include a validator in the configuration of the timezone and add a corresponding unit test.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
mariusneo: :+1:

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
mariusneo: :+1:

src/test/java/io/confluent/connect/jdbc/dialect/BaseDialectTest.java
mariusneo: :+1:

src/test/java/io/confluent/connect/jdbc/dialect/BaseDialectTest.java
mariusneo: modified.

rhauch:
@mariusneo thanks for the quick turnaround! Things look pretty good so far, but I'm going to do another pass, hopefully tomorrow.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -240,6 +243,14 @@
   public static final long TIMESTAMP_DELAY_INTERVAL_MS_DEFAULT = 0;
   private static final String TIMESTAMP_DELAY_INTERVAL_MS_DISPLAY = "Delay Interval (ms)";
 
+  public static final String DB_TIMEZONE_CONFIG = "db.timezone";
+  public static final String DB_TIMEZONE_DEFAULT = "UTC";
+  private static final String DB_TIMEZONE_CONFIG_DOC =
+      "Alternative TimeZone of the database, to be used by JDBC driver instead of UTC (default)"
+          + "when instantiating PreparedStatements.";
```
rhauch: How about changing this to be similar to what's used in the sink?

    Name of the timezone that should be used in the connector when
    querying with time-based criteria. Defaults to UTC.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -187,6 +190,13 @@
       + "specific dialect. All properly-packaged dialects in the JDBC connector plugin "
       + "can be used.";
 
+  public static final String DB_TIMEZONE_CONFIG = "db.timezone";
+  public static final String DB_TIMEZONE_DEFAULT = "UTC";
+  private static final String DB_TIMEZONE_CONFIG_DOC =
+      "Name of the timezone that should be used in the connector when "
+      + "specifying time-based criteria. Defaults to UTC.";
```
rhauch: This is for the sink, so I'm amending my earlier suggestion:

```suggestion
      + "inserting time-based values. Defaults to UTC.";
```

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -20,56 +20,52 @@
...
 import java.util.GregorianCalendar;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.TimeZone;
 
 public class DateTimeUtils {
 
-  public static final TimeZone UTC = TimeZone.getTimeZone("UTC");
+  private static final ThreadLocal<Map<TimeZone, Calendar>> TIMEZONE_CALENDARS =
+      ThreadLocal.withInitial(HashMap::new);
```
rhauch: HashMap's `computeIfAbsent()` method used below is not threadsafe, and although currently the methods in this utility are only called from the task's thread, it might be worth using `ConcurrentHashMap::new` here so that this class is able to be called concurrently from multiple threads.

src/test/java/io/confluent/connect/jdbc/sink/PreparedStatementBinderTest.java
```diff
@@ -137,9 +140,16 @@ public void bindRecordInsert() throws SQLException, ParseException {
...
+    Calendar utcCalendar = DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC));
+    verify(statement, times(1))
+        .setDate(index++, new java.sql.Date(((java.util.Date) valueStruct.get("date")).getTime()),
+            utcCalendar);
+    verify(statement, times(1))
+        .setTime(index++, new java.sql.Time(((java.util.Date) valueStruct.get("time")).getTime()),
+            utcCalendar);
+    verify(statement, times(1)).setTimestamp(index++,
+        new java.sql.Timestamp(((java.util.Date) valueStruct.get("timestamp")).getTime()),
+        utcCalendar);
```
rhauch: Nit: would be good to clean up the formatting of these methods to match our multi-line method call style, and check the changes in the other tests, too.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
mariusneo: In a first iteration i was thinking to do as you suggested, but then I thought that this is not necessary anymore because the map is being used as a threadlocal variable . In the way the DateTimeUtils class is currently built (based on ThreadLocal variables) there's  no way of falling into concurrency issues.

Take into account that if the class wouldn't be based on ThreadLocal holders, the java.util.Calendar obtained from the method io.confluent.connect.jdbc.util.DateTimeUtils#getTimeZoneCalendar could be used sinmultaneously   in multiple threads, which would cause problems (because java.util.Calendar is not thread-safe).

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
mariusneo: I used as inspiration : 

https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java

```
/**
	 * Default JDBC TimeZone. Unless specified, the JVM default TimeZone is going to be used by the underlying JDBC Driver.
	 *
	 * @since 5.2.3
	 */
	String JDBC_TIME_ZONE = "hibernate.jdbc.time_zone";
```

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
mariusneo: :+1

mariusneo:
@rhauch sorry for still having multi-line formatting issues. I've reviewed my changes and hopefully now the formatting looks in a better state. thank you for your patience.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
rhauch: Ah, that's a great point. The calendars are not threadsafe, so the thread local approach doesn't need to be concurrent.

mariusneo:
@rhauch could you please have a look on the latest changes from this PR? Thanks in advance, Marius.

PR #507, opened 2018-10-18
CC-2900: Update logs in JDBC Sink and Source

cyrusv:
Upgraded a couple of trace logs to debug.

Added a bunch of new logs to give us more insight into how JDBC runs.

https://confluentinc.atlassian.net/browse/CC-2900

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
@@ -128,6 +128,7 @@ public class BufferedRecords {
 
   public List<SinkRecord> flush() throws SQLException {
     if (records.isEmpty()) {
+      log.debug("Records is empty");
       return new ArrayList<>();
     }
     for (SinkRecord record : records) {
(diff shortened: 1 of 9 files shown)


rhauch:
@cyrusv should this PR should be based on the `5.0.x` branch rather than `master`?

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -130,6 +130,7 @@ public void run() {
   }
 
   public void shutdown() {
+    log.info("Shutting down TableMonitorThread");
```
rhauch: Since this is INFO level, how about:
```suggestion
    log.info("Shutting down thread monitoring tables.");
```

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -177,6 +178,7 @@ public BufferedRecords(
   }
 
   public void close() throws SQLException {
+    log.info("Closing BufferedRecords with preparedStatement: {}", preparedStatement);
```
jeqo: Just found this PR when reporting this issue https://github.com/confluentinc/kafka-connect-jdbc/issues/746

Consider changing `info` level to `debug` or `trace` in this case.

PR #508, opened 2018-10-22
MINOR: bump up version of sqlite-jdbc library to the latest version available

mariusneo:
Bump up version of the sqlite-jdbc library used for testing as embedded database in order to benefit of making use of jdbc4 classes.


mariusneo:
@rhauch it's done. thanks for the hint.

PR #509, opened 2018-10-22
CC-2910: Changed source task to close connections when finished

rhauch:
The JDBC source task was attempting to close the CachedConnectionProvider and its connections when the task was stopped, which happens asynchronously while the `poll()` method may be still working in the task’s primary thread. If the connection provider is closed but is needed again, it will reopen connections as needed but these will never be closed.

With this change, the connection provider is closed at the end of `poll()` after the task has been stopped (or after an exception has been thrown) and after the connections are no longer being used.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -77,6 +77,7 @@ public class JdbcSourceTask extends SourceTask {
 
   @Override
   public void start(Map<String, String> properties) {
+    log.info("Starting JDBC source task");
     try {
       config = new JdbcSourceTaskConfig(properties);
     } catch (ConfigException e) {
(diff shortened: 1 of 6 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -316,6 +328,8 @@ public void stop() throws ConnectException {
         return null;
       } catch (Throwable t) {
         resetAndRequeueHead(querier);
+        // This task has failed, so close any resources (may be reopened if needed) before throwing
+        closeResources();
```
rayokota: Maybe change this to `running.set(false)`, and then have a finally block around the contents of this method that calls `closeResources()`?  Then `closeResources()` would check that `running` is false before doing anything.

rhauch: Well `running` is more about the start-stopped state of the connector, not whether it is actually doing anything. The provider and dialect will both reopen if necessary, so that's not really an issue (tho they might not be closed again, which is kind of the state we might be in prior to this fix).

rhauch:
Rebased on 5.0.x to not conflict with recent changes.

rayokota:
LGTM

PR #510, opened 2018-10-22
CC-2455: Corrected timestamp query for DB2

rhauch:
The new query works on all versions of DB2.

See also #479 (already merged) for a fix on 5.0.x and later.


rhauch:
This will have to be merged manually onto 5.0.x, since the code has been refactored and already fixed (see #479).

jcustenborder:
LGTM

PR #518, opened 2018-11-01
CC-2961: Improve the log messages around connections to include sanitized DB information

rhauch:


src/main/java/io/confluent/connect/jdbc/util/ConnectionProvider.java
```diff
@@ -51,4 +51,14 @@ boolean isConnectionValid(
...
+
+  /**
+   * Get the publicly viewable identifier for this connection provider and / or the database.
+   * The resulting value should not contain any secrets or passwords.
+   *
+   * @return the identifier; never null
+   */
+  default String identifier() {
+    return toString();
+  }
```
rhauch: Step 1: In order to log useful information about the connection providers, we add a new `identifier` method that returns a string, and default to the `toString()` method.

src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java
```diff
@@ -122,6 +122,11 @@ public synchronized void close() {
     }
   }
 
+  @Override
+  public String identifier() {
+    return provider.identifier();
+  }
```
rhauch: Step 2: The CachedConnectionProvider just delegates to the real provider.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1691,6 +1691,22 @@ protected String getSqlType(SinkRecordField f) {
...
+   */
+  protected String sanitizedUrl(String url) {
+    // Only replace standard URL-type properties ...
+    return url.replaceAll("(?i)([?&]password=)[^&]*", "$1****");
+  }
+
+  @Override
+  public String identifier() {
+    return name() + " database " + sanitizedUrl(jdbcUrl);
+  }
```
rhauch: Step 3: The `DatabaseDialect` interface extends `ConnectionProvider`, so in each dialect we override the `identifier()` method to include the name and sanitized JDBC URL. By default, we just sanitize the standard JDBC URL property `password`, but since different JDBC drivers have different URL formats, some custom dialects override `sanitizedUrl` to also sanitize other components of the URL.

src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
```diff
@@ -161,4 +161,11 @@ public String buildUpsertQueryStatement(
     builder.append(")");
     return builder.toString();
   }
+
+  @Override
+  protected String sanitizedUrl(String url) {
+    // DB2 has semicolon delimited property name-value pairs
+    return super.sanitizedUrl(url)
+                .replaceAll("(?i)(;password=)[^;]*", "$1****");
```
rayokota: Maybe `"(?i)([:;]password=)[^;]*"` in case they put the password first (right after the colon) like `"jdbc:db2://sysmvs1.stl.ibm.com:5021/STLEC1:password=dbadm;user=dbadm;"`

PR #527, opened 2018-11-15
MINOR Add log statements to debug different flush paths

cyrusv:
This will illuminate why we flush, either because of mismatched
schemas, batch size limits, or processed entire message batch.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
@@ -112,6 +112,7 @@ public class BufferedRecords {
       // Continue with current batch state
       records.add(record);
       if (records.size() >= config.batchSize) {
+        log.debug("Flushing buffered records after exceeding configured batch size.");
         flushed = flush();
       } else {
         flushed = Collections.emptyList();
@@ -119,6 +120,8 @@ public class BufferedRecords {
     } else {
       // Each batch needs to have the same SchemaPair, so get the buffered records out, reset
       // state and re-attempt the add
+      log.debug("Flushing buffered records after due to unequal schema pairs: "
+          + "Current Schemas: {}, Next Schemas: {}", currentSchemaPair, schemaPair);
       flushed = flush();
       currentSchemaPair = null;
       flushed.addAll(add(record));
@@ -131,6 +134,7 @@ public class BufferedRecords {
       log.debug("Records is empty");
       return new ArrayList<>();
     }
+    log.debug("Flushing {} Buffered Records", records.size());
     for (SinkRecord record : records) {
       preparedStatementBinder.bindRecord(record);
     }
--- a/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
@@ -121,7 +121,7 @@ public class BufferedRecords {
       // Each batch needs to have the same SchemaPair, so get the buffered records out, reset
       // state and re-attempt the add
       log.debug("Flushing buffered records after due to unequal schema pairs: "
-          + "Current Schemas: {}, Next Schemas: {}", currentSchemaPair, schemaPair);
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -112,13 +112,16 @@ public BufferedRecords(
       // Continue with current batch state
       records.add(record);
       if (records.size() >= config.batchSize) {
+        log.debug("Flushing buffered records after exceeding configured batch size.");
```
rhauch: How about including `records.size()` and maybe even `config.batchSize` values in the message? Maybe:
```suggestion
        log.debug(
            "Flushing buffered records after number of records {} exceeds configured batch size {}.", 
            records.size(), 
            config.batch.size
        );
```

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -131,6 +134,7 @@ public BufferedRecords(
       log.debug("Records is empty");
       return new ArrayList<>();
     }
+    log.debug("Flushing {} Buffered Records", records.size());
```
rhauch: Nit: 
```suggestion
    log.debug("Flushing {} buffered records", records.size());
```

src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
```diff
@@ -66,7 +66,9 @@ void write(final Collection<SinkRecord> records) throws SQLException {
       }
       buffer.add(record);
     }
-    for (BufferedRecords buffer : bufferByTable.values()) {
+    for (TableId tableId : bufferByTable.keySet()) {
+      BufferedRecords buffer = bufferByTable.get(tableId);
```
rhauch: Minor: If we're going to use the keys and values, it might be better to iterate over the entries:
```
for (Map.Entry<TableId, BufferRecords> entry : bufferByTable.entrySet()) {
  TableId tableId = entry.getKey();
  BufferedRecords buffer = entry.getValue();
```
Thoughts?

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -112,13 +112,16 @@ public BufferedRecords(
...
+        log.debug("Flushing buffered records after exceeding configured batch size.");
         flushed = flush();
       } else {
         flushed = Collections.emptyList();
       }
     } else {
       // Each batch needs to have the same SchemaPair, so get the buffered records out, reset
       // state and re-attempt the add
+      log.debug("Flushing buffered records after due to unequal schema pairs: "
+          + "Current Schemas: {}, Next Schemas: {}", currentSchemaPair, schemaPair);
```
rhauch: Nit:
```suggestion
          + "current schemas: {}, next schemas: {}", currentSchemaPair, schemaPair);
```

src/main/java/io/confluent/connect/jdbc/sink/metadata/SchemaPair.java
```diff
@@ -46,4 +46,8 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hash(keySchema, valueSchema);
   }
+
+  public String toString() {
+    return String.format("<SchemaPair: %s, %s>", keySchema.toString(), valueSchema.toString());
```
rhauch: Since the varargs parameter of `String.format(...)` is of type `Object` for which `toString()` will be called if non-null, it's a bit safer to do the following in case the key schema or value schema are null:
```suggestion
    return String.format("<SchemaPair: %s, %s>", keySchema, valueSchema);
```

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
cyrusv: I decided to put the flushed size log into the `flush()` method itself so I can be sure to catch it in every code path within the flush itself. I will add the batch size

PR #545, opened 2018-12-13
CC-3525: Reduced log message for creating query

rhauch:
The INFO-level log message about creating query was introduced recently, but INFO level results in the message being logged too frequently (about every 5 seconds). It has been reduced to DEBUG-level.


cyrusv:
This looks good, but I was expecting a change to `GenericDatabaseDialect.addFieldToSchema` per Yeva's comment

cyrusv:
https://github.com/confluentinc/kafka-connect-jdbc/pull/546

PR #556, opened 2018-12-26
Fix tests to pass on windows as well

satm12:
Some tests did not close task resources which caused an exception to be
thrown when deleting the test db directory.
Update tests to expect sql statements with system new line character.


satm12:
[clabot:check]

Issue #219, opened 2017-05-15
unparsable date format


sunnn:
ERROR Failed to get current time from DB using query select CURRENT_TIMESTAMP; on database SQLite (io.confluent.connect.jdbc.util.JdbcUtils:205) 
Unparseable date: "2017-05-15 10:25:12" does not match (\p{Nd}++)\Q-\E(\p{Nd}++)\Q-\E(\p{Nd}++)\Q \E(\p{Nd}++)\Q:\E(\p{Nd}++)\Q:\E(\p{Nd}++)\Q.\E(\p{Nd}++) 
@willyhoang

aplamada:
I experienced the same issue.
Is this a limitation of sqlite? Can one use a timestamp / timestamp+incrementing [mode ](https://docs.confluent.io/current/connect/connect-jdbc/docs/source_config_options.html#mode) sqlite?

wicknicks:
@sunnn, @aplamada can you paste the entire exception here? also: could you try using the latest version of the connector. that might resolve db specific issues. thanks!

aplamada:
@wicknicks unfortunately I am no longer working on the project. As I mentioned previously I tried to use `mode=timestamp+incrementing` with a sqlite [example](https://docs.confluent.io/current/connect/kafka-connect-jdbc/source-connector/index.html#examples).  

*Just A Thought*: Keeping in mind that the [quick start](https://docs.confluent.io/current/connect/kafka-connect-jdbc/source-connector/index.html#quick-start) is done with a sqlite database, it might be helpful to continue with a sqlite database also in the [example](https://docs.confluent.io/current/connect/kafka-connect-jdbc/source-connector/index.html#examples) section.

a-titorenko:
same issue: 
`[2018-12-27 12:13:36,137] ERROR Failed to run query for table TimestampIncrementingTableQuerier{table="order_events", query='null', topicPrefix='raw-', incrementingColumn='seq_num', timestampColumns=[updated]}: {} (io.confluent.connect.jdbc.source.JdbcSourceTask:326)
java.sql.SQLException: Error parsing time stamp
        at org.sqlite.jdbc3.JDBC3ResultSet.getTimestamp(JDBC3ResultSet.java:576)
        at io.confluent.connect.jdbc.dialect.GenericDatabaseDialect.currentTimeOnDB(GenericDatabaseDialect.java:463)
[...]

PR #596, opened 2019-02-12
CC-3812: Fix for SQLite timestamp parse error

Prathibha-m:
Fix for issue in https://github.com/confluentinc/kafka-connect-jdbc/issues/219


src/main/java/io/confluent/connect/jdbc/dialect/SqliteDatabaseDialect.java
```diff
@@ -129,4 +129,9 @@ public String buildUpsertQueryStatement(
     builder.append(")");
     return builder.toString();
   }
+
+  @Override
+  protected String currentTimestampDatabaseQuery() {
+    return "SELECT strftime('%Y-%m-%d %H:%M:%S.%f','now')";
```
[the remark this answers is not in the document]
Prathibha-m: Thanks @rhauch for reviewing! 

Based on this suggestion: https://github.com/confluentinc/kafka-connect-jdbc/issues/219#issuecomment-450123435 
The error seemed to be because the sqlite JDBC driver expected a timestamp along with the milliseconds value. 

Retrieving the timestamp as follows fixed this
https://stackoverflow.com/questions/17574784/sqlite-current-timestamp-with-milliseconds

src/test/java/io/confluent/connect/jdbc/dialect/SqliteDatabaseDialectTest.java
```diff
@@ -276,4 +280,10 @@ public void testDescribeTable() throws SQLException {
     assertEquals(false, columnDefn.isPrimaryKey());
     assertEquals(true, columnDefn.isOptional());
   }
+
+  @Test
+  public void useCurrentTimestampValue() throws SQLException {
+    Calendar cal = DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone("PST"));
+    assertNotNull(dialect.currentTimeOnDB(sqliteHelper.connection, cal));
```
rhauch: Do we want to verify that the returned string matches the format? Just checking that the method returns non-null doesn't seem too useful.

It's actually going to be difficult to verify the timestamp in the string, because the timestamp that the DB returns will be off by a fraction of a second of one that the test method gets, but that might be the difference between one minute and the next, or one hour and the next, or even one day and the next. So, about the best we could do is check that the format is correct and each component is not off by more than one (or seconds off by more than 5) from the test method's timestamp.

Prathibha-m: That makes sense! I modified the unit test to support that verification now.

PR #597, opened 2019-02-12
Flatten + Demux in write

HarisGodil:


HarisGodil:
whoops, accidental creation

PR #605, opened 2019-02-28
Fix PostgreSqlDatabaseDialect to generate proper UPSERT queries when …

### LINKED ISSUE #401: Postgresql sink constructs invalid upsert sql when pk.fields include all fields
io.confluent.connect.jdbc.sink.dialect.PostgreSqlDialect.getUpsertQuery()

change

    builder.append(") ON CONFLICT (");
    joinToBuilder(builder, ",", keyCols, escaper());
    builder.append(") DO UPDATE SET ");
    joinToBuilder(
        builder,
        ",",
        cols,
        new StringBuilderUtil.Transform<String>() {
          @Override
          public void apply(StringBuilder builder, String col) {
            builder.append(escaped(col)).append("=EXCLUDED.").append(escaped(col));
          }
        }
    );

to

    builder.append(") ON CONFLICT (");
    joinToBuilder(builder, ",", keyCols, escaper());
    if (cols.size() > 0) {
      builder.append(") DO UPDATE SET ");
      joinToBuilder(
          builder,
          ",",
          cols,
          new StringBuilderUtil.Transform<String>() {
            @Override
            public void apply(StringBuilder builder, String col) {
              builder.append(escaped(col)).append("=EXCLUDED.").append(escaped(col));
            }
          });
    } else {
      builder.append(") DO NOTHING");
    }

hashhar:
…no non-key columns are present

This commit fixes #401.

- Currently the PostgreSqlDatabaseDialect generates UPSERT queries of
  the form `ON CONFLICT (<keyColums>) DO UPDATE SET (<nonKeyColumns>)`
  unconditionally. This causes malformed queries to be generated when
  `nonKeyColumns` is empty. ie. all columns are part of the primary key.
- To fix this scenario, we generate `DO NOTHING` in the case when
  `nonKeyColumns` is empty. To test for this, two additional asserts are
  added to `PostgreSqlDatabaseDialectTest#upsert()`.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -243,11 +243,15 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
            .delimitedBy(",")
            .transformedBy(ExpressionBuilder.columnNames())
            .of(keyColumns);
-    builder.append(") DO UPDATE SET ");
-    builder.appendList()
-           .delimitedBy(",")
-           .transformedBy(transform)
-           .of(nonKeyColumns);
+    if (nonKeyColumns.isEmpty()) {
+      builder.append(") DO NOTHING");
+    } else {
+      builder.append(") DO UPDATE SET ");
+      builder.appendList()
+              .delimitedBy(",")
+              .transformedBy(transform)
+              .of(nonKeyColumns);
+    }
     return builder.toString();
   }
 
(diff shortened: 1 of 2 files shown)


hashhar:
[clabot:check]

hashhar:
@rhauch I've added the tests for the change within the existing `upsert()` test. Should I move them to separate tests?

hashhar:
@rhauch Just wondering if you had gotten around to giving this a view yet. Do let me know if I need to improve something.

hashhar:
Thanks for merging this upstream. And thanks for all the work on the Kafka Connect ecosystem.

AlexVPopov:
Sorry, in which version is this available?

reneveyj:
Any workaround before this comes to production?

PR #618, opened 2019-03-18
Non optional columns should be NOT NULL

sasa-tomic:
According to the SQL Specification, a primary key can not contain NULL.
Some SQL engines accept a column spec for primary key without "NOT NULL",
but some do not - for instance, in DB2 one cannot define a "PRIMARY KEY"
column without "NOT NULL".

References: https://www.ibm.com/support/knowledgecenter/en/SSEPGG_11.1.0/com.ibm.db2.luw.sql.ref.doc/doc/r0000927.html
> PRIMARY KEY (column-name,...)
> Defines a primary key that is composed of the identified columns. The clause must not be specified more than once, and the identified columns must be defined as NOT NULL.

https://docs.microsoft.com/en-us/sql/relational-databases/tables/create-primary-keys?view=sql-server-2017
> All columns defined within a PRIMARY KEY constraint must be defined as NOT NULL.

(and is essentially the same for all other DBMS dialects that I checked)

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1633,6 +1633,9 @@ public class GenericDatabaseDialect implements DatabaseDialect {
           f.schemaType(),
           f.defaultValue()
       );
+      if (!isColumnOptional(f)) {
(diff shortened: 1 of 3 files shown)


sasa-tomic:
[clabot:check]

sasa-tomic:
Master is broken due to 0c2926ae099aae80cadb059c3ce2fdc20db7085e

PR #619, opened 2019-03-18
Fixed typo in tests quoteIdentfiiers -> quoteIdentifiers

sasa-tomic:


sasa-tomic:
Master is broken due to 0c2926a so the CI check fails

PR #622, opened 2019-03-20
Backport db.timzeone config to 4.1.x

cyrusv:
Date/time/timestamp values read from the source DB will be
instantiated into the specified timezone.

Backports https://github.com/confluentinc/kafka-connect-jdbc/pull/505/files

4.1.x must support Java 7, explaining some of the deviations from the original.

The strategy is to minimize changing public interfaces of the 4.1.x branch, even if it requires deviation from the 5.x feature and changes to the tests files in 4.1.x. Notably, I decided that `DbDialect.withTimeZone(TimeZone)` is the best way to provide timezones to the Column Formatters, even though it opens the potential for NPE if the DbDialect object is never given a timezone, since it is impossible for the user to misconfigure it is an acceptable tradeoff.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
@@ -29,6 +29,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
 
 import javax.xml.bind.DatatypeConverter;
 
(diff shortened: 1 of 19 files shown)


src/main/java/io/confluent/connect/jdbc/sink/dialect/DbDialect.java
```diff
@@ -43,12 +44,18 @@
 
   private final String escapeStart;
   private final String escapeEnd;
+  private TimeZone timeZone;
```
rhauch: I understand that with these changes the `withTimeZone(...)` method will always be called as the `db.timezone` configuration defaults to UTC. While that should always be the case, do you think it's safer to initialize this to UTC?

cyrusv: I thought about initializing UTC as a default, and I decided it's better to get an error than to silently get a value that contradicts what the user configured the connector to do. I will add this comment to the docs for withTimeZone if you agree with my reasoning!

rhauch: Yeah, that's good reasoning.

PR #628, opened 2019-03-28
Use Java 7 to valiate timezones

cyrusv:
Fixes build error since 4.1.x should support Java 7.
The behavior is slightly more limited than in the 5.x version, as demonstrated in the tests. The docs will reflect that in the 4.1.x version the valid timezones are listed by `TimeZone.getAvailableIDs()`

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/util/TimeZoneValidator.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/TimeZoneValidator.java
@@ -16,8 +16,9 @@
 
 package io.confluent.connect.jdbc.util;
 
-import java.time.DateTimeException;
-import java.time.ZoneId;
+import java.util.Arrays;
+import java.util.TimeZone;
+
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -226,7 +226,8 @@
   public static final String DB_TIMEZONE_DEFAULT = "UTC";
   private static final String DB_TIMEZONE_CONFIG_DOC =
       "Name of the JDBC timezone that should be used in the connector when "
-          + "querying with time-based criteria. Defaults to UTC.";
+          + "querying with time-based criteria.  Valid timezones are those given by "
+          + "TimeZone.getAvailableIDs().Defaults to UTC.";
```
rhauch: Is there value in having this code in the docs, or can we just say any valid JDK time zones are allowed?

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -181,7 +181,8 @@
   public static final String DB_TIMEZONE_DEFAULT = "UTC";
   private static final String DB_TIMEZONE_CONFIG_DOC =
       "Name of the JDBC timezone that should be used in the connector when "
-          + "inserting time-based values. Defaults to UTC.";
+          + "inserting time-based values. Valid timezones are those given by "
+          + "TimeZone.getAvailableIDs(). Defaults to UTC.";
```
rhauch: Same comment.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
cyrusv: I think you're right. My original idea was to specify since JDK 7 and JDK 8 have slightly different support, but your suggestion of "any valid JDK time zone" leaves it to the user to use the correct input for their own JDK version

PR #641, opened 2019-04-18
Fixes #127. Added delete support for sink.

### LINKED ISSUE #127: Delete Support for sink
Hello,

In kafka and especially kafka streams a delete is expressed as a null value being published for the specific key that needs to be deleted. I do understand that the current source can't catch deletes as the rows are then gone and wont come up as query results. But the Sink side should probably be able to handle deletes correctly. Especially if you want to store the results of a streams application in a relational database.

Best

### LINKED ISSUE #607: delete support for sink connector
jdbc sink connector deletes are disabled，what are the methods set deletes enabked。
i have tried use deleted.enabled=true，but it does not work。

andybryant:
A record with `null` value is considered a tombstone event and result in deleting the corresponding row in the destination table.
Deletes are disabled by default. Set config `delete.enabled` to `true` to enable. Also requires `pk.mode = record_key` otherwise it can't find the record to delete.

Tested against Postgres

Fixes #607 and fixes #127

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -361,6 +361,21 @@ public interface DatabaseDialect extends ConnectionProvider {
   );
 
   /**
+   * Build the DELETE prepared statement expression for the given table and its columns. Variables
+   * for each key column should also appear in the WHERE clause of the statement.
+   *
+   * @param table         the identifier of the table; may not be null
+   * @param keyColumns    the identifiers of the columns in the primary/unique key; may not be null
+   *                      but may be empty
+   * @return the delete statement; may not be null
+   * @throws UnsupportedOperationException if the dialect does not support upserts
+   */
+  String buildDeleteStatement(
+      TableId table,
+      Collection<ColumnId> keyColumns
+  );
+
+  /**
    * Build the DROP TABLE statement expression for the given table.
    *
    * @param table   the identifier of the table; may not be null
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -368,12 +368,14 @@ public interface DatabaseDialect extends ConnectionProvider {
    * @param keyColumns    the identifiers of the columns in the primary/unique key; may not be null
    *                      but may be empty
    * @return the delete statement; may not be null
-   * @throws UnsupportedOperationException if the dialect does not support upserts
+   * @throws UnsupportedOperationException if the dialect does not support deletes
    */
-  String buildDeleteStatement(
+  default String buildDeleteStatement(
       TableId table,
       Collection<ColumnId> keyColumns
-  );
+  ) {
+    throw new UnsupportedOperationException();
+  }
 
   /**
    * Build the DROP TABLE statement expression for the given table.
--- a/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.jdbc.sink;
 
+import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.slf4j.Logger;
@@ -26,8 +27,9 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import io.confluent.connect.jdbc.dialect.DatabaseDialect;
@@ -36,6 +38,10 @@ import io.confluent.connect.jdbc.sink.metadata.FieldsMetadata;
 import io.confluent.connect.jdbc.sink.metadata.SchemaPair;
 import io.confluent.connect.jdbc.util.ColumnId;
 import io.confluent.connect.jdbc.util.TableId;
+
+import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.InsertMode.INSERT;
+import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
 
 public class BufferedRecords {
   private static final Logger log = LoggerFactory.getLogger(BufferedRecords.class);
@@ -47,10 +53,14 @@ public class BufferedRecords {
   private final Connection connection;
 
   private List<SinkRecord> records = new ArrayList<>();
-  private SchemaPair currentSchemaPair;
+  private Schema keySchema;
+  private Schema valueSchema;
   private FieldsMetadata fieldsMetadata;
-  private PreparedStatement preparedStatement;
-  private StatementBinder preparedStatementBinder;
+  private PreparedStatement updatePreparedStatement;
+  private PreparedStatement deletePreparedStatement;
+  private StatementBinder updateStatementBinder;
+  private StatementBinder deleteStatementBinder;
+  private boolean deletesInBatch = false;
 
   public BufferedRecords(
       JdbcSinkConfig config,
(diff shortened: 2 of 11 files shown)


rhauch:
@andybryant thanks for updating the changes and creating this new PR. Very much looking forward to having this feature, but currently the build results show a number of test failures.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -102,6 +102,13 @@
       + " table, when possible.";
   private static final String BATCH_SIZE_DISPLAY = "Batch Size";
 
+  public static final String DELETE_ENABLED = "delete.enabled";
+  private static final String DELETE_ENABLED_DEFAULT = "false";
+  private static final String DELETE_ENABLED_DOC =
+      "Whether to treat ``null`` record values as deletes. Requires ``pk.mode`` "
+      + "to be ``record_key``.";
```
rhauch: Because `Validator` only has the context of an individual configuration value, it's not straightforward to validate  one config value (e.g., `pk.mode` must be `record_key`). One technique that works, however, is to add a check in the `JdbcSinkConfig` constructor to check these two config values against this constraint and throw `ConfigException` if they are not compatible. This will help ensure that invalid configurations are caught immediately before the connector is started and without having to figure out why the task failing at a later time.

Also, to help C3 and other GUI tools present valid options to begin with, how about adding
* a `Recommender` implementation for `delete.enabled` that hides it if `pk.mode` is not `record_key`; and
* a `Recommender` for `pk.mode` that recommends only `record_key` if `delete.enabled=true`.

It also may be

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -360,6 +360,21 @@ String buildUpsertQueryStatement(
...
+   * @param table         the identifier of the table; may not be null
+   * @param keyColumns    the identifiers of the columns in the primary/unique key; may not be null
+   *                      but may be empty
+   * @return the delete statement; may not be null
+   * @throws UnsupportedOperationException if the dialect does not support upserts
+   */
+  String buildDeleteStatement(
+      TableId table,
+      Collection<ColumnId> keyColumns
+  );
```
rhauch: Should this be a default method that throws UnsupportedOperationException? It's possible that other projects depend on this connector and add their own dialect, and adding a non-default method would break those projects. Yeah, it's likely they extend `GenericDatabaseDialect`, but we can't rule it out. Plus, I think this makes sense, too, because by default dialects (that don't extend `GenericDatabaseDialect`) probably won't support deletes anyway.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -229,6 +293,26 @@ private String getInsertSql() {
...
 
+  private String getDeleteSql() {
+    String sql = null;
+    if (config.deleteEnabled) {
+      switch (config.pkMode) {
+        case RECORD_KEY:
+          if (fieldsMetadata.keyFieldNames.isEmpty()) {
+            throw new ConnectException("Require primary keys to support delete");
+          }
+          sql = dbDialect.buildDeleteStatement(
```
rhauch: Should we catch `UnsupportedOperationException` and throw a `ConnectException` with a more useful message?

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -229,6 +293,26 @@ private String getInsertSql() {
...
+              tableId,
+              asColumns(fieldsMetadata.keyFieldNames)
+          );
+          break;
+        default:
+          throw new ConnectException("Deletes are only supported for pk.mode record_key");
```
rhauch: Strictly speaking, this case wouldn't be necessary if Config validation ensures that `config.pkMode == RecordKey` when `config.deleteEnabled`. However, I still think it's good to have, just in case.

src/test/java/io/confluent/connect/jdbc/sink/BufferedRecordsTest.java
```diff
@@ -103,6 +103,123 @@ public void correctBatching() throws SQLException {
...
+
+    assertEquals(Arrays.asList(recordA, recordA, recordADelete), buffer.add(recordB));
+
+    assertEquals(Collections.singletonList(recordB), buffer.add(recordA));
+
+    assertEquals(Collections.singletonList(recordA), buffer.flush());
+  }
+
+  @Test
+  public void insertThenDeleteThenInsertInBatchFlush() throws SQLException {
```
rhauch: Aren't here a few other cases that we should consider testing? It'd be great to know how much coverage of  `add(record)` the unit tests have, and whether to add more tests if there are more critical lines not hit by the current tests.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -67,64 +77,87 @@ public BufferedRecords(
...
+          insertSql,
+          deleteSql,
+          fieldsMetadata
       );
       close();
-      preparedStatement = connection.prepareStatement(sql);
-      preparedStatementBinder = dbDialect.statementBinder(
-          preparedStatement,
+      // should this be using dbDialect.createPreparedStatement ?
+      updatePreparedStatement = connection.prepareStatement(insertSql);
```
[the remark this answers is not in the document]
andybryant: done

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
andybryant: added recommenders for both and tested locally in C3

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
andybryant: good idea - fixed

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
andybryant: done

src/test/java/io/confluent/connect/jdbc/sink/BufferedRecordsTest.java
andybryant: added a few extra tests in latest commit. Let me know if you can think of any other missing scenarios

andybryant:
@rhauch - thanks for reviewing. I've pushing a new commit aimed at addressing your review suggestions. Looks like Jenkins is happy. Let me know if anything else is outstanding. Cheers Andy

nghyjohn:
I have tried to delete record on mysql,
while doing sink, I have used TimestampConverter for my last_update field.
Error shows seems it is because delete action will have null in after,
any suggestions?

[2019-04-30 08:48:45,271] ERROR WorkerSinkTask{id=mysql-jdbc-sink-0} Task threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask:177)
org.apache.kafka.connect.errors.ConnectException: Tolerance exceeded in error handler
        at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execAndHandleError(RetryWithToleranceOperator.java:178)
        at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execute(RetryWithToleranceOperator.java:104)
        at org.apache.kafka.connect.runtime.TransformationChain.apply(TransformationChain.java:50)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.convertAndTransformRecord(WorkerSinkTask.java:506)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.convertMessages(WorkerSinkTask.java:464)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:320)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:224)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:192)
        at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:175)
        at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:219)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: org.apache.kafka.connect.errors.DataException: Only Struct objects supported for [converting timestamp formats], found: null
        at org.apache.kafka.connect.transforms.util.Requirements.requireStruct(Requirements.java:52)
        at org.apache.kafka.connect.transforms.TimestampConverter.applyWithSchema(TimestampConverter.java:336)
        at org.apache.kafka.connect.transforms.TimestampConverter.apply(TimestampConverter.java:275)
        at org.apache.kafka.connect.runtime.TransformationChain.lambda$apply$0(TransformationChain.java:50)
        at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execAndRetry(RetryWithToleranceOperator.java:128)
        at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execAndHandleError(RetryWithToleranceOperator.java:162)
        ... 14 more
[2019-04-30 08:48:45,276] ERROR WorkerSinkTask{id=mysql-jdbc-sink-0} Task is being killed and will not recover until manually restarted (org.apache.kafka.connect.runtime.WorkerTask:178)

andybryant:
@rhauch - any chance you'll get some time to review the latest update? I'd hate for this PR to go stale too.
Thx Andy

andybryant:
@nghyjohn - the transforms are performed before the connector is called (you can see it's not in the stacktrace at all). So for this setup to work, you would need all your transforms to not fail on null message values. It would be nice if there was an elegant way to do this but I'm not aware of one. In your case, you could write a wrapper transform that forwards messages with null values, otherwise it delegates to the standard TimestampConverter.

devshawn:
@andybryant @rhauch - any updates on this PR? We need this feature for an upcoming project and I'm trying to avoid forking and building this.

bigkraig:
@rhauch friendly ping

rhauch:
Merging to the `5.3.x` branch for inclusion in the upcoming CP 5.3.0. Also forward merged to `master`.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -67,64 +77,86 @@ public BufferedRecords(
...
-        record.valueSchema()
-    );
+    final List<SinkRecord> flushed = new ArrayList<>();
+
+    boolean schemaChanged = false;
+    if (!Objects.equals(keySchema, record.keySchema())) {
+      keySchema = record.keySchema();
+      schemaChanged = true;
+    }
+    if (isNull(record.valueSchema())) {
```
anthavio: Oddly even when record.value() returns null, record.valueSchema() returns NOT null for me

limadelrey:
> Merging to the `5.3.x` branch for inclusion in the upcoming CP 5.3.0. Also forward merged to `master`.

When is CP 5.3.0 expected to launch?

helpermethod:
Is this already part of the 5.3.0 beta-releases?

helpermethod:
For anyone who is struggling with using the `5.3.0` release:

If you are using JSON both the `payload` and the `schema` attributes of the `value` need to be set to `null`.

PR #642, opened 2019-04-18
CC-4423 Remove semicolon from Db2 dialect timestamp query

cyrusv:
This fix was included in 4.1.x
https://github.com/confluentinc/kafka-connect-jdbc/blob/4.1.x/src/main/java/io/confluent/connect/jdbc/util/JdbcUtils.java#L249

but lost when the code was refactored from 5.0.x to 5.1.x

One big shortcoming here is that there seems to be no testing against this, but the only way to really know would be to have an integration test that spins up a DB2 instance. Any other tests against this are kind of over-specific, and what we really want to know is that this works against DB2. Sadly this is pressing issue for users so no time to write integration test.


cyrusv:
@kkonstantine, I added some unit tests

PR #649, opened 2019-05-15
Fix NPE when auto.create and incorrect access privileges

MarkC0x:
Made a simple change in `TableDefinitions.java` (against the 5.0.x branch) to handle the situation where `dbTable` can be null, and cause a NPE from the `TableDefinitions.refresh(...)` method (as per the suggestion from rhauch in the issue comments).

Verified that the _original exception_ from the attempt to create the table with insufficient permissions gets re-thrown (and not the NPE).


MarkC0x:
[clabot:check]

gharris1727:
This was fixed in #706, which duplicates this PR.

PR #653, opened 2019-05-21
Update PostgreSqlDatabaseDialect.java

kourosko:
Add VIEW in table search. This will allow to insert in a view and from there in postgres with triggers we can push it any where we want.
My original proble was casting text to uuid.


kourosko:
clabot:check

kourosko:
clabot:check

PR #655, opened 2019-06-06
CC-4946: changing log msgs on filtering and avoiding frequently retrying tasks when no tables

shreyakannan19:
1) make log messages clearer after filtering tables
2) when no readable tables are available, don't run tasks and inform user via log

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -117,15 +117,14 @@ public class JdbcSourceConnector extends SourceConnector {
     String query = config.getString(JdbcSourceConnectorConfig.QUERY_CONFIG);
     List<Map<String, String>> taskConfigs;
     if (!query.isEmpty()) {
-      taskConfigs = new ArrayList<>(1);
       Map<String, String> taskProps = new HashMap<>(configProperties);
       taskProps.put(JdbcSourceTaskConfig.TABLES_CONFIG, "");
-      taskConfigs.add(taskProps);
+      taskConfigs = Collections.singletonList(taskProps);
     } else {
       List<String> currentTables = tableMonitorThread.tables();
       if (currentTables.isEmpty()) {
-        taskConfigs = new ArrayList<>();
-        log.warn("No tasks are running because no tables were found");
+        taskConfigs = Collections.emptyList();
+        log.warn("No tasks will be run because no tables were found");
       } else {
         int numGroups = Math.min(currentTables.size(), maxTasks);
         List<List<String>> tablesGrouped = ConnectorUtils.groupPartitions(currentTables, numGroups);
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -115,25 +115,30 @@ public class JdbcSourceConnector extends SourceConnector {
   @Override
   public List<Map<String, String>> taskConfigs(int maxTasks) {
     String query = config.getString(JdbcSourceConnectorConfig.QUERY_CONFIG);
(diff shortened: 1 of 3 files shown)


shreyakannan19:
[clabot:check]

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -115,25 +115,30 @@ public void start(Map<String, String> properties) throws ConnectException {
   @Override
   public List<Map<String, String>> taskConfigs(int maxTasks) {
     String query = config.getString(JdbcSourceConnectorConfig.QUERY_CONFIG);
+    List<Map<String, String>> taskConfigs;
     if (!query.isEmpty()) {
-      List<Map<String, String>> taskConfigs = new ArrayList<>(1);
+      taskConfigs = new ArrayList<>(1);
```
C0urante: Nit: you can replace this line and the `taskConfigs.add(taskProps)` line with a single call to the `Collections.singletonList(...)`

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -115,25 +115,30 @@ public void start(Map<String, String> properties) throws ConnectException {
...
-      List<List<String>> tablesGrouped = ConnectorUtils.groupPartitions(currentTables, numGroups);
-      List<Map<String, String>> taskConfigs = new ArrayList<>(tablesGrouped.size());
-      for (List<String> taskTables : tablesGrouped) {
-        Map<String, String> taskProps = new HashMap<>(configProperties);
-        taskProps.put(JdbcSourceTaskConfig.TABLES_CONFIG,
-                      StringUtils.join(taskTables, ","));
-        taskConfigs.add(taskProps);
+      if (currentTables.isEmpty()) {
+        taskConfigs = new ArrayList<>();
+        log.warn("No tasks are running because no tables were found");
```
C0urante: I'd be careful about the tense of this log message. This method returns configs that are used to generate tasks, so I think future tense might be more appropriate. Something like "No tasks will be run..." instead of "No tasks are running..." but feel free to tweak if you'd like.

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -115,25 +115,30 @@ public void start(Map<String, String> properties) throws ConnectException {
...
-      int numGroups = Math.min(currentTables.size(), maxTasks);
-      List<List<String>> tablesGrouped = ConnectorUtils.groupPartitions(currentTables, numGroups);
-      List<Map<String, String>> taskConfigs = new ArrayList<>(tablesGrouped.size());
-      for (List<String> taskTables : tablesGrouped) {
-        Map<String, String> taskProps = new HashMap<>(configProperties);
-        taskProps.put(JdbcSourceTaskConfig.TABLES_CONFIG,
-                      StringUtils.join(taskTables, ","));
-        taskConfigs.add(taskProps);
+      if (currentTables.isEmpty()) {
+        taskConfigs = new ArrayList<>();
```
C0urante: Nit: can replace with `Collections.emptyList()` to avoid creating a new `ArrayList` every time.

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -139,7 +139,12 @@ private synchronized boolean updateTables() {
     }
 
     if (!filteredTables.equals(this.tables)) {
-      log.debug("After filtering we got tables: " + Arrays.toString(filteredTables.toArray()));
+      if (filteredTables.isEmpty()) {
+        log.debug("Based on the supplied filtering rules, there are no matching tables to read from: ");
```
C0urante: Nit: don't need the `: ` at the end of this log message

PR #665, opened 2019-06-19
Add basic snowflake support

GeitV:
Made snowflake work, used same type of upsert logic that is used in OracleDatabaseDialect


GeitV:
[clabot:check]

gerbrand:
@GeitV Internally we're using Snowflake too. For bulk loading and storing data into and from Snowflake has a separate mechanism (https://docs.snowflake.com/en/user-guide-data-load.html and https://docs.snowflake.com/en/user-guide-data-unload.html) so using KafkaConnectJdbc for bulk storing and loading data would probably not a good idea.

Issue #379, opened 2018-03-12
com.microsoft.sqlserver.jdbc.SQLServerException: Column in table is of a type that is invalid for use as a key column in an index.


gmyrianthous:
I am trying to run a sink in order to send data from a kafka topic to a Microsoft SQL Server however I get the following error: 

```
org.apache.kafka.connect.errors.RetriableException: com.microsoft.sqlserver.jdbc.SQLServerException: Column 'AC_NO' in table 'CUSTOMERS' is of a type that is invalid for use as a key column in an index.
        at io.confluent.connect.jdbc.sink.JdbcSinkTask.put(JdbcSinkTask.java:84)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:495)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:288)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:198)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:166)
        at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:170)
        at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:214)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Column 'AC_NO' in table 'CUSTOMERS' is of a type that is invalid for use as a key column in an index.
        at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:258)
        at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1535)
        at com.microsoft.sqlserver.jdbc.SQLServerStatement.doExecuteStatement(SQLServerStatement.java:845)
[...]

PR #673, opened 2019-07-02
SqlServerDialect uses varchar(900) for primary key as string type

HungUnicorn:
When sinking data to SqlServer with table auto-creation is enabled and primary key is string type, it was using varchar(max) for primary key, which leads to the exception:
```
com.microsoft.sqlserver.jdbc.SQLServerException: Column in table is of a type that is invalid for use as a key column in an index
```
https://github.com/confluentinc/kafka-connect-jdbc/issues/379

SqlServer 2012(and before) allows `varchar(900)` to be primary key. More than that is not acceptable.
As other RDBMS related sinks identify string type primary key and set it particularly, it makes sens to adapt it for MSSQL sink.


HungUnicorn:
[clabot:check]

PR #679, opened 2019-07-11
Add the possibility to specify a custom TopicNameProvider to convert table name to topic name on the fly

RossierFl:
This PR aim to allow the user to specify a custom TopicNameProvider that can generate the topic name from the table name on the fly. 

To give you an example, my team and I were unable to use the `table.whitelist` & `table.blacklist` feature of this connector because of the different naming convention between Kafka & DBMSs.

As we run a multitenant Kafka cluster managed by an API, the naming convention play a huge role, and with the actual version of the kafka-connect-jdbc, it's for example impossible to lowercase the table name, replacing '_' by '-', etc. 

The proposed solution allow the user to fully customised the mapping between table name & topic name based on his own naming convention.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.jdbc.source;
 
+import io.confluent.connect.jdbc.util.TopicNameProvider;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.source.SourceRecord;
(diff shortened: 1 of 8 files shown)


RossierFl:
@rhauch @C0urante Could one of you have a look and provide some feedback if possible ?

C0urante:
@RossierFl this is a fascinating proposal and the use case is interesting; however, I'm wondering if an SMT wouldn't be more appropriate here. Adding a new JDBC connector-specific API would increase configuration complexity and it seems unlikely to be used by most of our users, whereas SMTs are already supported by the Connect framework and can be used with any type of connector, not just the JDBC connector.

I guess this boils down to a simple question: is there anything this new API would provide that an SMT wouldn't be able to with equal or less configuration complexity?

RossierFl:
@C0urante You are right, it's way easier with an SMT. I guess we can close this request as it's useless ! 
Thank you for your input !

PR #696, opened 2019-08-13
changes to set for schema

venkatesh1753:


gharris1727:
@venkatesh1753 `connect-transforms` is a proprietary repository, and cannot be included as a dependency in this open source project.
SMTs should be configured in the connector configuration, and not linked into any connectors.

PR #697, opened 2019-08-19
[confluent] implemented snowflake dialect and temp table merge flow

Davidtedwards2017:


Davidtedwards2017:
[clabot:check]

cyrusv:
@Davidtedwards2017, thank you for the hard work and contribution!

A challenge we have with merging new dialects is that in order to fully support them, we need integration tests that validate end-to-end functionality. I understand this is especially challenging with connectors to cloud services, which require shared accounts, and typically cost money to operate, even for testing.

For Snowflake, specifically, we highly recommend using the connector developed by Snowflake: https://docs.snowflake.net/manuals/user-guide/kafka-connector.html. This connector will soon also be available on [Confluent Hub](http://hub.confluent.io)!

jfinzel:
I am very interested in this functionality.  I understand the problem with integration testing, however, the connector developed by Snowflake only loads the raw message data, which many problems unsolved.  The functionality of this PR goes much further than that.  Is there any path forward for this?  @Davidtedwards2017 do you have any ideas as to how to workaround this integration testing limitation?

sfc-gh-pmaric:
@jfinzel  what are the problems that are unsolved with the Kafka connector?  Many Snowflake customers use it for a schema on read approach, keeping the raw data decoupled from processed data, which ensures the connector does not error when Kafka topic structure changes.  Some customers use the Snowflake JDBC driver to load topics to Snowflake in a normalized manner, so schema on write approach.

WilsonSunBritten:
I love this solution and was finding myself writing something quite similar to address the exact same problem. I would love to see this actually release so it could be used. It's so valuable I would consider using this branch over the mainline even though it's significantly behind and will not receive updates. 

As the primary issue to merging(post a conflict resolution) is the ability to integration test end to end this snowflake connection, what if this solution was broken up into parts. The advancements to batched records and for dialects to be able to extend them could come into this main project along with all their tests. The SnowflakeDialect could be left unmerged but instead be available in a separate standalone repository. 

See https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialectProvider.java, as Dialect providers are brought in using Java's Service Provider Api, any individual who wishes to make use of the SnowflakeDialect could bring it in as a separate jar and place it on the kafka-connect jdbc classpath.

getrdone93:
@Davidtedwards2017 

Is there any movement on this issue?

PR #706, opened 2019-09-13
MINOR: Avoid NPE when columns cannot be described

gharris1727:
This prevents shadowing table create permission exceptions

Signed-off-by: Greg Harris <gregh@confluent.io>

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/util/TableDefinitions.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/TableDefinitions.java
@@ -86,6 +86,8 @@ public class TableDefinitions {
     if (dbTable != null) {
       log.info("Refreshing metadata for table {} to {}", tableId, dbTable);
       cache.put(dbTable.id(), dbTable);
+    } else {
+      log.warn("Failed to refresh metadata for table {}", tableId);
     }
     return dbTable;
   }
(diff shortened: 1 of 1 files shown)


gharris1727:
This duplicates #649 and only one of these two should be merged.

src/main/java/io/confluent/connect/jdbc/util/TableDefinitions.java
```diff
@@ -83,8 +83,10 @@ public TableDefinition refresh(
       TableId tableId
   ) throws SQLException {
     TableDefinition dbTable = dialect.describeTable(connection, tableId);
-    log.info("Refreshing metadata for table {} to {}", tableId, dbTable);
-    cache.put(dbTable.id(), dbTable);
+    if (dbTable != null) {
+      log.info("Refreshing metadata for table {} to {}", tableId, dbTable);
+      cache.put(dbTable.id(), dbTable);
+    }
```
wicknicks: should we log a warn in an else block here saying that the task failed to describe the table? It's hard to follow "absence of logs" when these connectors are typically running against a large number of tables.

gharris1727: It looks like some of the call-sites log the warning based on the return value, but some don't.
I'll add the log message.
```suggestion
    } else {
      log.warn("Failed to refresh metadata for table {}", tableId);
    }
```

PR #713, opened 2019-09-26
Update readme to indicate integration tests are required

cyrusv:
This fix will warn customers to consider integration tests before contributing a dialect.


gharris1727:
Should we include this after an integration test framework is in place?
@rhauch

cyrusv:
@gharris1727 I'm not sure it matters if the framework is in place or not, since we still wouldn't accept them. This is just a fair warning for potential contributors, and hopefully in the near future the path to writing such a test will be smoother.

PR #727, opened 2019-10-21
CC-6773: Abort source task if all columns were not converted correctly.

wicknicks:
Signed-off-by: Arjun Satish <arjun@confluent.io>


wicknicks:
@levzem good point, we probably do want to use that sort of config but that would need a much bigger change and across all possible area of error. In this PR, I wanted to fix one such bug. For this specific instance, I think we want to discuss if it is ok to silently skip bad records, or throw an exception that kills the task (as we are doing in this PR). IMHO, we should start by throwing exception, and later add the error handling behavior you mention as a feature request.

levzem:
it is probably safer to fail in that case and add it as configurable later. in either case, the log statements need to be updated and then LGTM from me

wicknicks:
@levzem fixed the log messages

levzem:
LGTM

PR #730, opened 2019-10-24
suffix param

maqroll:
One of our DB2 source task extraction requires hints ([optguidelines](https://www.ibm.com/support/knowledgecenter/en/SSEPGG_11.1.0/com.ibm.db2.luw.admin.perf.doc/doc/c0070117.html)). 

DB2 requires optguidelines to be at the end of the statement so we added a suffix query config.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
@@ -54,25 +54,23 @@ public class BulkTableQuerier extends TableQuerier {
     switch (mode) {
       case TABLE:
         builder.append("SELECT * FROM ").append(tableId);
-        addSuffixIfPresent(builder);
 
-        String queryStr = builder.toString();
-        
-        recordQuery(queryStr);
-        log.debug("{} prepared SQL query: {}", this, queryStr);
-        stmt = dialect.createPreparedStatement(db, queryStr);
         break;
       case QUERY:
-        addSuffixIfPresent(builder);
-        queryStr = query + builder.toString();
-
-        recordQuery(queryStr);
-        log.debug("{} prepared SQL query: {}", this, queryStr);
-        stmt = dialect.createPreparedStatement(db, queryStr);
+        builder.append(query);  
+        
         break;
       default:
         throw new ConnectException("Unknown mode: " + mode);
     }
+
+    addSuffixIfPresent(builder);
+    
+    String queryStr = builder.toString();
+
+    recordQuery(queryStr);
+    log.debug("{} prepared SQL query: {}", this, queryStr);
+    stmt = dialect.createPreparedStatement(db, queryStr);
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
@@ -50,11 +50,10 @@ public class BulkTableQuerier extends TableQuerier {
 
   @Override
   protected void createPreparedStatement(Connection db) throws SQLException {
+    ExpressionBuilder builder = dialect.expressionBuilder();  
     switch (mode) {
       case TABLE:
-        ExpressionBuilder builder = dialect.expressionBuilder();
-
-        builder.append("SELECT * FROM").append(tableId);
+        builder.append("SELECT * FROM ").append(tableId);
         addSuffixIfPresent(builder);
 
         String queryStr = builder.toString();
@@ -64,9 +63,12 @@ public class BulkTableQuerier extends TableQuerier {
         stmt = dialect.createPreparedStatement(db, queryStr);
         break;
       case QUERY:
-        recordQuery(query);
-        log.debug("{} prepared SQL query: {}", this, query);
-        stmt = dialect.createPreparedStatement(db, query);
+        addSuffixIfPresent(builder);
+        queryStr = query + builder.toString();
+
+        recordQuery(queryStr);
+        log.debug("{} prepared SQL query: {}", this, queryStr);
+        stmt = dialect.createPreparedStatement(db, queryStr);
         break;
       default:
         throw new ConnectException("Unknown mode: " + mode);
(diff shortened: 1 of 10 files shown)


maqroll:
@gharris1727 How do you manage to get your PRs reviewed?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -255,6 +255,12 @@
       + "For backward compatibility, the default is 'always'.";
   public static final String QUOTE_SQL_IDENTIFIERS_DISPLAY = "Quote Identifiers";
 
+  public static final String QUERY_SUFFIX_CONFIG = "suffix.query";
+  public static final String QUERY_SUFFIX_DEFAULT = "";
+  public static final String QUERY_SUFFIX_DOC = 
+      "Suffix to append at the end of the generated query.";
```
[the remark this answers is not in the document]
maqroll: Even bulk query mode in mode query? In that case the provided query gets executed as such, so if you need a suffix you can append it directly to the query. Is the only case I'll consider leaving out. What do you think?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -133,7 +135,7 @@ protected void createPreparedStatement(Connection db) throws SQLException {
     }
 
     // Append the criteria using the columns ...
-    criteria = dialect.criteriaFor(incrementingColumn, timestampColumns);
+    criteria = dialect.criteriaFor(incrementingColumn, timestampColumns, suffix);
     criteria.whereClause(builder);
```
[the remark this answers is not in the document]
maqroll: As a side effect TimestampIncrementingCriteriaTest didn't test this change? Should we add BulkTableQuerierTest and TimestampIncrementingTableQuerierTest to do so?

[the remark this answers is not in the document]
gharris1727: Yes, I think those are places that need to test this behavior.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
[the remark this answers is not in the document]
maqroll: @gharris1727 What about this comment?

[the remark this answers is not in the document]
gharris1727: I agree that the suffix is not necessary in bulk mode, and the same behavior can be accomplished by editing the query itself.
However, if someone changes modes, I think the different modes should behave the same for this suffix. If they didn't, you'd have to add documentation here like "except in bulk query mode", which I think is a worse user experience.

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
```diff
@@ -41,25 +42,33 @@ public BulkTableQuerier(
...
         recordQuery(queryStr);
         log.debug("{} prepared SQL query: {}", this, queryStr);
         stmt = dialect.createPreparedStatement(db, queryStr);
         break;
       case QUERY:
-        recordQuery(query);
-        log.debug("{} prepared SQL query: {}", this, query);
-        stmt = dialect.createPreparedStatement(db, query);
+        addSuffixIfPresent(builder);
+        queryStr = query + builder.toString();
```
gharris1727: Can you call `builder.append(query);` rather than using string concatenation?

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
```diff
@@ -41,25 +42,33 @@ public BulkTableQuerier(
...
+
+        recordQuery(queryStr);
+        log.debug("{} prepared SQL query: {}", this, queryStr);
+        stmt = dialect.createPreparedStatement(db, queryStr);
```
gharris1727: I think it's time to pull this common block outside the switch case.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -194,7 +195,12 @@ public void start(Map<String, String> properties) {
 
       if (mode.equals(JdbcSourceTaskConfig.MODE_BULK)) {
         tableQueue.add(
-            new BulkTableQuerier(dialect, queryMode, tableOrQuery, topicPrefix)
+            new BulkTableQuerier(
+                dialect, 
+                queryMode, 
+                tableOrQuery, 
+                topicPrefix, 
+                suffix)
```
gharris1727: tiny nit:
```suggestion
                suffix
            )
```

src/main/java/io/confluent/connect/jdbc/source/TableQuerier.java
```diff
@@ -134,6 +138,12 @@ private void closeResultSetQuietly() {
     resultSet = null;
   }
 
+  protected void addSuffixIfPresent(ExpressionBuilder builder) {
+    if (!this.suffix.trim().isEmpty()) {
```
gharris1727: Maybe we can trim the suffix on the way in, when reading the configuration.

src/test/java/io/confluent/connect/jdbc/source/TableQuerierTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import org.junit.Test;
+import org.mockito.Matchers;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class TableQuerierTest {
```
gharris1727: Can you add a test for when the suffix is omitted entirely?

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
maqroll: Sure. I did it because previous version (in that case) didn't push query through builder.

maqroll:
@gharris1727 Is there anything pending on my side to complete this PR?

gharris1727:
@maqroll No, you've done everything correctly. I'm waiting on a code freeze to expire, and this feature should make it to 5.4.1 after that.

gharris1727:
I was incorrect before, the earliest release that will contain this feature are `5.5.0` and `6.0.0`.

PR #738, opened 2019-11-05
Implement the ability to validate config provider values

klebervirgilio:
Prior to this commit, the
https://docs.confluent.io/current/connect/references/restapi.html#put--connector-plugins-(string-name)-config-validate endpoint
wasn't capable to properly validate the `connection.url` property using vault or env config providers.

This should solve this issue: #738


klebervirgilio:
[clabot:check]

gharris1727:
I believe that this PR is not necessary, and that we were just experiencing a configuration issue.

PR #748, opened 2019-11-21
* Handling deletes with schema !null

maqroll:
Try to fix bug described in #171 

@gharris1727 pls have a look.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
@@ -85,7 +85,7 @@ public class BufferedRecords {
       schemaChanged = true;
     }
     if (isNull(record.valueSchema())) {
-      // For deletes, both the value and value schema come in as null.
+      // For deletes, value and optionally value schema come in as null.
       // We don't want to treat this as a schema change if key schemas is the same
       // otherwise we flush unnecessarily.
       if (config.deleteEnabled) {
@@ -101,7 +101,6 @@ public class BufferedRecords {
       valueSchema = record.valueSchema();
       schemaChanged = true;
     }
-
     if (schemaChanged) {
       // Each batch needs to have the same schemas, so get the buffered records out
       flushed.addAll(flush());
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -203,6 +208,12 @@ public BufferedRecords(
     return flushedRecords;
   }
 
+  private void checkDeleteInValue(SinkRecord record) {
```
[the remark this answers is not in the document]
maqroll: Yeah, but in that case that method (add) go over complexity limits...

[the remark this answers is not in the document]
gharris1727: Yeah, you can add a checkstyle suppression.
I think theres already quite a few, and refactoring this in a bugfix PR is not a great idea.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -153,6 +152,12 @@ public BufferedRecords(
         );
       }
     }
+    
+    // set deletesInBatch if schema value is not null
```
rhauch: "schema value" or "record value"?

gharris1727:
```suggestion
    // set deletesInBatch if the record value is null, but schema value was non-null
```

gharris1727: Actually, re-reading this it makes more sense, it just leaves out the record-value because that's already in the code itself.

rhauch: Maybe this does make sense, given that the null-schema case is handled above this line, and that when we get here we know the value schema is not null.

gharris1727:
@maqroll would you please rebase your changes on the new 5.3.x and fix any merge conflicts?

Thanks!

maqroll:
ok?

gharris1727:
This fix will appear in 5.3.3 and 5.4.1 whenever they are released.

PR #758, opened 2019-12-05
Fix Postgres OOM by limiting fetch size

### LINKED ISSUE #34: java.lang.OutOfMemoryError

gharris1727:
This PR adds an integration test with an embedded postgres instance.
There are three tests:
1. Verify that the test setup produces OOM exceptions with large batch sizes.
2. Verify that the test setup does _not_ produce OOM exceptions with small batch sizes.
3. Verify that tables are not left locked when autoCommit is disabled. 

Test 3 addresses a side-effect of autoCommit that was called out in #466, where read-only queries with autoCommit disabled would leave open locks on the database. This test verifies that the `JdbcSourceTask` does not leave open locks from any operations.

#466 specifically observed that the TableMonitorThread would leave open locks if auto-commit was disabled on that connection. This PR circumvents that issue by leaving the TableMonitorThread connection as-is, with autoCommit enabled. There are no tests covering the Connector or TableMonitorThread, because the behavior of those classes is unaffected by these changes. A connector test would also have some extra complication as the TableMonitorThread is asynchronous.

Fixes #34 .

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -16,6 +16,7 @@
 
 package io.confluent.connect.jdbc.dialect;
 
+import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
@@ -81,6 +82,10 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
    * to {@link ResultSet#FETCH_FORWARD forward} as an optimization for the driver to allow it to
    * scroll more efficiently through the result set and prevent out of memory errors.
    *
+   * <p>This method also sets the {@link PreparedStatement#setFetchSize(int) fetch size} to
+   * the {@link JdbcSourceConnectorConfig#BATCH_MAX_ROWS_CONFIG batch size} of the connector.
+   * This will bound the memory usage of the connector for large tables.
+   *
    * @param stmt the prepared statement; never null
    * @throws SQLException the error that might result from initialization
    */
@@ -88,6 +93,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
   protected void initializePreparedStatement(PreparedStatement stmt) throws SQLException {
     log.trace("Initializing PreparedStatement fetch direction to FETCH_FORWARD for '{}'", stmt);
     stmt.setFetchDirection(ResultSet.FETCH_FORWARD);
+    stmt.setFetchSize(config.getInt(JdbcSourceConnectorConfig.BATCH_MAX_ROWS_CONFIG));
   }
 
 
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -242,13 +242,18 @@ public class JdbcSourceTask extends SourceTask {
       }
     }
 
-    disableAutoCommit(cachedConnectionProvider.getConnection());
     running.set(true);
     log.info("Started JDBC source task");
   }
 
   protected CachedConnectionProvider connectionProvider(int maxConnAttempts, long retryBackoff) {
-    return new CachedConnectionProvider(dialect, maxConnAttempts, retryBackoff);
+    return new CachedConnectionProvider(dialect, maxConnAttempts, retryBackoff) {
+      @Override
+      protected void onConnect(final Connection connection) throws SQLException {
+        super.onConnect(connection);
+        connection.setAutoCommit(false);
+      }
+    };
   }
 
   //This method returns a list of possible partition maps for different offset protocols
@@ -356,14 +361,6 @@ public class JdbcSourceTask extends SourceTask {
     return null;
   }
 
-  private void disableAutoCommit(Connection connection) {
-    try {
-      connection.setAutoCommit(false);
-    } catch (SQLException e) {
-      throw new ConnectException("Failed trying to disable autocommit for table queries");
-    }
-  }
-
   private void resetAndRequeueHead(TableQuerier expectedHead) {
     log.debug("Resetting querier {}", expectedHead.toString());
     TableQuerier removedQuerier = tableQueue.poll();
@@ -387,16 +384,22 @@ public class JdbcSourceTask extends SourceTask {
       boolean incrementingOptional = false;
       boolean atLeastOneTimestampNotOptional = false;
       final Connection conn = cachedConnectionProvider.getConnection();
-      Map<ColumnId, ColumnDefinition> defnsById = dialect.describeColumns(conn, table, null);
-      for (ColumnDefinition defn : defnsById.values()) {
-        String columnName = defn.id().name();
-        if (columnName.equalsIgnoreCase(incrementingColumn)) {
-          incrementingOptional = defn.isOptional();
-        } else if (lowercaseTsColumns.contains(columnName.toLowerCase(Locale.getDefault()))) {
-          if (!defn.isOptional()) {
-            atLeastOneTimestampNotOptional = true;
+      boolean autoCommit = conn.getAutoCommit();
+      try {
+        conn.setAutoCommit(true);
+        Map<ColumnId, ColumnDefinition> defnsById = dialect.describeColumns(conn, table, null);
+        for (ColumnDefinition defn : defnsById.values()) {
+          String columnName = defn.id().name();
+          if (columnName.equalsIgnoreCase(incrementingColumn)) {
+            incrementingOptional = defn.isOptional();
+          } else if (lowercaseTsColumns.contains(columnName.toLowerCase(Locale.getDefault()))) {
+            if (!defn.isOptional()) {
+              atLeastOneTimestampNotOptional = true;
+            }
           }
         }
+      } finally {
+        conn.setAutoCommit(autoCommit);
       }
 
       // Validate that requested columns for offsets are NOT NULL. Currently this is only performed
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -99,6 +99,8 @@ public class JdbcSourceTask extends SourceTask {
 
     cachedConnectionProvider = connectionProvider(maxConnAttempts, retryBackoff);
 
+    disableAutoCommit(cachedConnectionProvider.getConnection());
+
     List<String> tables = config.getList(JdbcSourceTaskConfig.TABLES_CONFIG);
     String query = config.getString(JdbcSourceTaskConfig.QUERY_CONFIG);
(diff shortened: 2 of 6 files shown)


src/main/java/io/confluent/connect/jdbc/source/TableQuerier.java
```diff
@@ -107,12 +109,24 @@ public boolean next() throws SQLException {
...
     schemaMapping = null;
     lastUpdate = now;
   }
 
+  private void releaseLocksQuietly() {
+    if (db != null) {
+      try {
+        db.commit();
+      } catch (SQLException e) {
+        // intentionally ignored
```
levzem: maybe `log.warn()`?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -99,6 +99,8 @@ public void start(Map<String, String> properties) {
 
     cachedConnectionProvider = connectionProvider(maxConnAttempts, retryBackoff);
 
+    disableAutoCommit(cachedConnectionProvider.getConnection());
```
levzem: should we not renable autocommit after every statement - isn't there times when we do need autocommit?

gharris1727: In a single-threaded scenario, we could swap autocommit on and off. I initially thought it would be easier to follow if it was always disabled, and then do explicit commits everywhere.
But seeing as the only other query made in the task (other than the main query) is in `validateNonNullable`, which is only called during the start() method. I could move the disableAutoCommit after the validateNonNullable, and remove the explicit commit there.

I think after the task startup, it's simpler to leave autoCommit disabled, and then call commit explicitly at the conclusion of each query.

src/main/java/io/confluent/connect/jdbc/source/TableQuerier.java
gharris1727: `// intentionally ignored` was the style at the time

gharris1727:
Note that `io.zonky.test:embedded-postgres` is Apache 2.0 licensed.

src/main/java/io/confluent/connect/jdbc/source/TableQuerier.java
```diff
@@ -107,12 +109,24 @@ public boolean next() throws SQLException {
   public void reset(long now) {
     closeResultSetQuietly();
     closeStatementQuietly();
+    releaseLocksQuietly();
```
wicknicks: should this be in a finally block, in case any of the above fail?

src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
```diff
@@ -0,0 +1,163 @@
...
+              row.append(rs.getObject(i));
+              row.append("\t");
+            }
+            log.debug(row.toString());
+          }
+        }
+      }
+    }
+    assertEquals("Found idle locks left open", 0, count);
+    log.info("No locks found");
```
wicknicks: we usually don't have log statements in tests because they are mostly run in jenkins (no one is watching), and if there are no asserts, then it is assumed that the scenario they are testing did not error.

src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
```diff
@@ -0,0 +1,163 @@
...
+    props.put(JdbcSourceConnectorConfig.CONNECTION_USER_CONFIG, "postgres");
+    props.put(JdbcSourceConnectorConfig.MODE_CONFIG, JdbcSourceConnectorConfig.MODE_BULK);
+    props.put(JdbcSourceTaskConfig.TOPIC_PREFIX_CONFIG, "topic_");
+  }
+
+  public void startTask() {
+
+    task = new JdbcSourceTask();
+    task.start(props);
+  }
```
wicknicks: do we need to test what happens when the table monitor thread is running in the same JVM as these tasks that are setting auto commit to false?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -242,6 +242,7 @@ public void start(Map<String, String> properties) {
       }
     }
 
+    disableAutoCommit(cachedConnectionProvider.getConnection());
```
javabrett: `CachedConnectionProvider#getConnection())` performs a validity-test on the connection instance it caches before returning it.  This may result in a dead connection being discarded and replaced with a new instance:

https://github.com/confluentinc/kafka-connect-jdbc/blob/66307690724f6fe54d0a2bb28bbfe8b30fc6a447/src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java#L63-L66

Due to the `start()`/`poll()` lifecycle, is it possible that a new connection without auto-commit disabled might become active?  Calls might need to be moved, or the option pushed-down into the connection-provider framework.

src/main/java/io/confluent/connect/jdbc/source/TableQuerier.java
gharris1727: All of the other `Quietly` functions here swallow the errors, so they won't fail in this block.

src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
gharris1727: This was a debugging statement for me, the assert above it captures what I want the function to verify.

src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
gharris1727: The table monitor thread is in essentially the same relationship to this task as other tasks from the JDBC connector, and we don't see any issues from this running in production.
This is already effectively tested by running an existing source and sink connector in the same JVM, since sink connectors disable autocommit.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -81,13 +82,18 @@ public PostgreSqlDatabaseDialect(AbstractConfig config) {
...
+   * This will bound the memory usage of the connector for large tables.
+   *
    * @param stmt the prepared statement; never null
    * @throws SQLException the error that might result from initialization
    */
   @Override
   protected void initializePreparedStatement(PreparedStatement stmt) throws SQLException {
     log.trace("Initializing PreparedStatement fetch direction to FETCH_FORWARD for '{}'", stmt);
     stmt.setFetchDirection(ResultSet.FETCH_FORWARD);
+    stmt.setFetchSize(config.getInt(JdbcSourceConnectorConfig.BATCH_MAX_ROWS_CONFIG));
```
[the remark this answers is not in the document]
gharris1727: 1. As far as I understand, MySQL has a similar behavior, except that their url workaround is effective. I'm sure that theres at least one other database that we don't know about that has this same issue, and would require the same fix.

    However, without the abillity to integration test this functionality, I don't feel comfortable making changes to those dialects. Even for Postgres, I needed to find a way to integration test this change before I felt comfortable putting this up for PR. The cost to adding integration tests at this time is also higher than we can afford right now, and beyond the scope of this fix.

2. If someone included this URL option without knowing, then I don't think they would know/care that it was being ignored. Therefore, I'll assume that someone using this URL option put it there explicitly.

    If the URL option is higher than the batch size (perhaps by someone trying to optimize throughput) then it would be effectively the same as if it was equal the batch size except with higher memory usage. This is because the connector will still only read at most `batch.size` of data from the result set regardless of the fetch size.

    If the URL option was lower than the batch size (perhaps by someone trying to optimize memory) then they would notice that it didn't reduce memory usage. This would be non-obvious, but if they're optimizing memory usage, why would they change this URL parameter before reducing the batch size? Tuning the batch size performs effectively the same memory tuning behavior that they want, but it's actually supported by the connector itself, and documented behavior, rather than depending on this pass-through driver tuning parameter.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
gharris1727: Thank you for the call-out, this was a serious issue with that implementation. I've added a callback that disables autocommit for all new connections from that cached connection provider.

gharris1727:
There are a few bug fixes/features which have added configuration options:

- `delete.enabled` https://github.com/confluentinc/kafka-connect-jdbc/pull/641
    This is arguable, since perhaps someone was relying on the error-on-null-value behavior, or wants to ensure that data is not deleted from their target system under any circumstances
- `quote.sql.identifiers` https://github.com/confluentinc/kafka-connect-jdbc/pull/572
    This could have been implemented without a config, but was included in order to maintain prior behavior for existing databases where data has already been transferred.
- `db.timezone` https://github.com/confluentinc/kafka-connect-jdbc/pull/505
    This definitely needs a configuration to give information about the database
- `connection.attempts`, `connection.backoff.ms`, https://github.com/confluentinc/kafka-connect-jdbc/pull/203
    This is a retry "feature", and it's reasonable to configure these to be more responsive or forgiving to network conditions
- `numeric.precision.mapping` https://github.com/confluentinc/kafka-connect-jdbc/pull/177 `numeric.mapping` https://github.com/confluentinc/kafka-connect-jdbc/pull/396
    This could have been implemented without a config, but was included in order to maintain prior behavior for existing databases where data has already been transferred.
- `schema.pattern` https://github.com/confluentinc/kafka-connect-jdbc/pull/109
    This definitely needs a configuration to give information about the database
- `validate.non.null` https://github.com/confluentinc/kafka-connect-jdbc/pull/51
    I think this shouldn't have been a configuration, and we should have just demoted this from an error to a warning. This is an example of a customer-specific configuration that isn't healthy long-term.

These are just the bug fixes that resulted in a configuration option, but there are many more that were fixed without a configuration option, many more than are listed here.

Overall, I think that the only justification for a configuration added as a bug fix is preservation of backwards compatibility, and I don't think that justification works here. Nobody is relying on the OOM exception as part of their workflow, and it has no impact on the actual data that's being transferred.

This is just the first dialect to have this fix applied, and applying it to each of them is simple. The only reason we're not adding them at this time is because of the cost of testing.
The last thing that I'd want is to add a configuration that is essentially a "throw OOM exception enable/disable" option, only to have it become obsolete after we make an investment and figure out which dialects can support it.

Let me know your thoughts @javabrett, and thanks for your review!

yohei1126:
Hi, when new Docker images includes this fix is published to Docker hub?
https://hub.docker.com/r/confluentinc/cp-kafka-connect/tags
It looks like `latest` and `5.4.0` tags are already updated in a day ago but I do not know these tags include this fix.

gharris1727:
This will not be released until the next patch release, so `5.3.3`, `5.4.1`, `5.5.0`, or `6.0.0`.
It will appear in docker images once those versions are released.
Until then you will need to build the connector from source in order to use this fix.

kgrimsby:
Is this included in the latest confluent 5.4 deb packages?

gharris1727:
@kgrimsby No, this will not be included in the 5.4 deb packages. It will be included in the deb packages of the versions I called out in my previous message.

kgrimsby:
My bad @gharris1727 ! Sorry for wasting your time..

Issue #755, opened 2019-12-04
Postgres ALTER TABLE fails when column type is BOOLEAN


jmakuc:
Hi there. I'm trying to write a record to a postgres table but it fails when tries to modify the table definition:

`[2019-12-04 20:25:30,590] WARN Amend failed, re-attempting (io.confluent.connect.jdbc.sink.DbStructure:170)
org.postgresql.util.PSQLException: ERROR: column "active" is of type boolean but default expression is of type integer
  Hint: You will need to rewrite or cast the expression.`

The error is triggered by this modification:

`ADD "active" BOOLEAN DEFAULT 1,`

I've tried on psql 10 and 11 directly in the command line and in fact you cannot define a default value for boolean column using the number 1. 

I've tracked the issue down to this line:

https://github.com/confluentinc/kafka-connect-jdbc/blob/237e4dcb88ace630fc5670520a4727216d3e568b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java#L1718

It says that '1' and '0' has better portability, so I didn't wanted to just overwrite the value because it could break some other dialect. On the other hand, copying the entire method to the Psql dialiect just to change that, seams too much duplication.

Ideas?

Thanks!

liukrimhrim:
> Hi there. I'm trying to write a record to a postgres table but it fails when tries to modify the table definition:
> 
> `[2019-12-04 20:25:30,590] WARN Amend failed, re-attempting (io.confluent.connect.jdbc.sink.DbStructure:170) org.postgresql.util.PSQLException: ERROR: column "active" is of type boolean but default expression is of type integer Hint: You will need to rewrite or cast the expression.`
> 
> The error is triggered by this modification:
> 
> `ADD "active" BOOLEAN DEFAULT 1,`
> 
> I've tried on psql 10 and 11 directly in the command line and in fact you cannot define a default value for boolean column using the number 1.
> 
> I've tracked the issue down to this line:
> 
> https://github.com/confluentinc/kafka-connect-jdbc/blob/237e4dcb88ace630fc5670520a4727216d3e568b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java#L1718
> 
> It says that '1' and '0' has better portability, so I didn't wanted to just overwrite the value because it could break some other dialect. On the other hand, copying the entire method to the Psql dialiect just to change that, seams too much duplication.
> 
> Ideas?
> 
> Thanks!

[...]

PR #759, opened 2019-12-10
CC-7382: use true/false as default values for boolean type

liukrimhrim:
Addressing ticket CC-7382: https://confluentinc.atlassian.net/browse/CC-7382, which comes from triaging this issue https://github.com/confluentinc/kafka-connect-jdbc/issues/755. 

The issue is reporting a bug of postgres dialect of jdbc connector. When there is boolean column with default value, there will be an exception. The bug is caused by us using 0/1 as default values for boolean type, which is not recognized by postgres. 

My fix is to overwrite `formatColumnValue()` to do special handling on boolean default for postgres. And I extended the unit test to test the changes. 

My code passed unit tests, and I have manually tested on postgres instance. It works as intended.

### DIFF
--- a/src/test/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialectTest.java
+++ b/src/test/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialectTest.java
@@ -97,7 +97,8 @@ public class MySqlDatabaseDialectTest extends BaseDialectTest<MySqlDatabaseDiale
         "`c3` VARCHAR(256) NOT NULL,\n" + "`c4` VARCHAR(256) NULL,\n" +
         "`c5` DATE DEFAULT '2001-03-15',\n" + "`c6` TIME(3) DEFAULT '00:00:00.000',\n" +
         "`c7` DATETIME(3) DEFAULT '2001-03-15 00:00:00.000',\n" + "`c8` DECIMAL(65,4) NULL,\n" +
-        "`c9` TINYINT DEFAULT 1,\n" + "PRIMARY KEY(`c1`))";
+        "`c9` TINYINT DEFAULT 1,\n" +
+        "PRIMARY KEY(`c1`))";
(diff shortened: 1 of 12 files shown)


src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -31,6 +35,16 @@
 
 public class PostgreSqlDatabaseDialectTest extends BaseDialectTest<PostgreSqlDatabaseDialect> {
 
+  @Override
+  @Before
+  public void setup() throws Exception {
+    super.setup();
```
gharris1727: I think you should add this case to the base test, even though it makes the PR much larger

liukrimhrim: I'm convinced.

src/test/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialectTest.java
```diff
@@ -97,7 +97,7 @@ public void shouldBuildCreateQueryStatement() {
         "`c3` VARCHAR(256) NOT NULL,\n" + "`c4` VARCHAR(256) NULL,\n" +
         "`c5` DATE DEFAULT '2001-03-15',\n" + "`c6` TIME(3) DEFAULT '00:00:00.000',\n" +
         "`c7` DATETIME(3) DEFAULT '2001-03-15 00:00:00.000',\n" + "`c8` DECIMAL(65,4) NULL,\n" +
-        "PRIMARY KEY(`c1`))";
+        "`c9` TINYINT DEFAULT 1,\n" + "PRIMARY KEY(`c1`))";
```
gharris1727:
```suggestion
        "`c9` TINYINT DEFAULT 1,\n" +
        "PRIMARY KEY(`c1`))";
```

PR #763, opened 2019-12-16
Make sink metadata column ordering more consistent.

norganna:
Currently the use of a HashMap to store the columns loses the ordering of the columns from the schema.
When the create table is generated, the order of the columns is unpredictable.
This change makes the hash a LinkedHashMap which mostly preserves the original column order (with the exception that key columns are placed before non-key columns).

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
@@ -19,9 +19,9 @@ import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
 
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


norganna:
[clabot:check]

gharris1727:
@norganna Thanks for your contribution!

Can you share how you noticed this issue? I'd be curious to know what database and driver was sensitive to column ordering, and how we might go about writing a test to prevent regressions in the future.

norganna:
We were doing MySQL to SQL server, and the BI guys complained that the new columns were in a jumbled up order (not the same as they were used to in the source schema). It's not necessarily an issue other than cosmetically

gharris1727:
I see. I went through the upstream ConnectSchema implementation and saw that it's using a LinkedHashMap to maintain order as well. That makes me think that this was just a small oversight in the implementation that's been here since the file was introduced in #100 .

I think that this also affects the only other call site of `allFields`, the amend table operation. It appears that all of the insert/upsert/update statements and binders use the separated key fields and non-key fields instead of the allFields variable, so they are unaffected.

@norganna Can you rebase your change on 5.0.x, and then update this PR to target that branch? This should be safe to backport that far. I would also appreciate any time you can put into writing a small unit test in `FieldsMetadataTest`. I think having two value schemas with the same fields but in a different order and verifying that the iteration order of each is correct is more than enough to catch a regression in the future.

norganna:
Closing this as #787 replaces it.

PR #765, opened 2019-12-17
Add option for dropping an invalid batch of messages

louisburton:
`drop.invalid.batch`
Add option for dropping an invalid batch of messages - rather than ending up in a poison pill situation requiring manual intervention.
https://rmoff.net/2019/10/15/skipping-bad-records-with-the-kafka-connect-jdbc-sink-connector/

Quick resiliency similar to the basic elasticsearch connector approach:
https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java#L307

DLQ support can still be enhanced in future.


louisburton:
[clabot:check]

PR #770, opened 2020-01-09
CC-7628 Set conservative timeout for JDBC to stop trying to reach drive

cyrusv:
By default, the driver has no limit for timing out trying to make
a connection (https://docs.oracle.com/javase/7/docs/api/java/sql/DriverManager.html#setLoginTimeout(int))
This fix will set to 1 minute.

Without this fix, the connect framework may hang in the main thread (perhaps when
validating a connector config during Table Recommender), and block the entire
node's REST API. A framework-level fix is ideal, but this fix will alleviate
the issue in the short-term and on older worker versions.

Also considered making this configurable, but 1 minute is pretty conservative

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -218,6 +218,10 @@ public class GenericDatabaseDialect implements DatabaseDialect {
       properties.setProperty("password", dbPassword.value());
     }
     properties = addConnectionProperties(properties);
+    // Timeout is 40 seconds to be as long as possible for customer to have a long connection
+    // handshake, while still giving enough time to validate once in the follower worker,
+    // and again in the leader worker and still be under 90s REST serving timeout
+    DriverManager.setLoginTimeout(40);
     Connection connection = DriverManager.getConnection(jdbcUrl, properties);
     if (jdbcDriverInfo == null) {
       jdbcDriverInfo = createJdbcDriverInfo(connection);


cyrusv:
@rhauch, @wicknicks, do I understand correctly that you two want to review/re-evaluate what the timeout should be here, and I should not merge this yet?

cyrusv:
Some updates on exploring this more:
 - Setting this globally in the connector will prevent some on-prem customers from overriding in their JVM some other way. I'm noticing this in cloud when I try to set globally using a REST extension.
 - Making this configurable may be a good idea if we use the URL strategy, but cannot be configurable at the global level. Specifying timeout in URL will be nice but will require us to test on each driver

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -214,6 +214,7 @@ public Connection getConnection() throws SQLException {
       properties.setProperty("password", dbPassword.value());
     }
     properties = addConnectionProperties(properties);
+    DriverManager.setLoginTimeout(60);
```
wicknicks: what's the reasoning behind 60 seconds here?

cyrusv: In the end, any number is a little arbitrary -- 60 just felt so conservative that no reasonable user would be waiting that long for a connection, and short enough that it does finally release our threads. I personally the 99.99% of connections will clear in under 10 seconds, but you never know with all the crazy routers, firewalls, proxies, etc. The main goal of the PR is to not block herder indefinitely, so I reasoned it is fine to be conservative. @wicknicks

C0urante: This happens during connector validation, right?

When a connector config is submitted to a distributed worker, it gets validated as soon as the worker has a chance to look at it. Then, if the worker isn't the leader, it forwards the config to the leader, which validates it _again_, and then finally writes it to the config topic. When all of this is done, the user's REST request is given a 200-level response.

Given that this means the connector config will possibly (even likely, since in any cluster of two or more workers you have a <=50% chance of hitting the leader with a request) be validated twice, and the REST request timeout in distributed mode is 90 seconds, maybe something like 40-45 seconds would be more appropriate here? That way we could avoid the scenario (which is possible with a driver connection timeout of 60 seconds) of the connection taking 60 seconds to succeed during the first worker's validation, then taking another 60 seconds to succeed during the leader's validation, then getting written to the config topic--even though at that point the user's REST request will have been met with a 500 timeout response.

cyrusv: That's a reasonable suggestion -- 40s is a good idea. If @wicknicks  is onboard i'll set to 40 and we can merge

wicknicks: yeah, @C0urante's suggestion makes sense. please add a comment to the doc, explaining this reasoning. thx in advance!

gharris1727:
@cyrusv Can you please retarget this for 5.0.x?

PR #774, opened 2020-01-17
Add possibility to customize update clause in updset/update mode for Oracle db

demidov163:
In order to solve problem, when we need compare some field in merge(update) clause with field in DB, I suggest new property:
update.clause, where clause set for example: 
update.clause= "PRODUCTS"."CHANGE_DATE" < incoming."CHANGE_DATE" 
incoming - it's constant

This enhancement give possibility to merge without additional querying from DB

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1400,13 +1400,13 @@ public class GenericDatabaseDialect implements DatabaseDialect {
              .delimitedBy(" AND ")
              .transformedBy(ExpressionBuilder.columnNamesWith(" = ?"))
              .of(keyColumns);
-      if (config.values().containsKey(JdbcSinkConfig.UPDATE_CONDITION_CONFIG)) {
-        String condition = config.getString(JdbcSinkConfig.UPDATE_CONDITION_CONFIG);
+      if (config.values().containsKey(JdbcSinkConfig.UPDATE_CLAUSE_CONFIG)) {
+        String condition = config.getString(JdbcSinkConfig.UPDATE_CLAUSE_CONFIG);
         builder.append(" AND ").append(condition);
       }
     } else {
-      if (config.values().containsKey(JdbcSinkConfig.UPDATE_CONDITION_CONFIG)) {
-        String condition = config.getString(JdbcSinkConfig.UPDATE_CONDITION_CONFIG);
+      if (config.values().containsKey(JdbcSinkConfig.UPDATE_CLAUSE_CONFIG)) {
+        String condition = config.getString(JdbcSinkConfig.UPDATE_CLAUSE_CONFIG);
         if (condition != null && !"".equals(condition)) {
           builder.append(" WHERE ").append(condition);
         }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1400,6 +1400,17 @@ public class GenericDatabaseDialect implements DatabaseDialect {
              .delimitedBy(" AND ")
              .transformedBy(ExpressionBuilder.columnNamesWith(" = ?"))
              .of(keyColumns);
+      if (config.values().containsKey(JdbcSinkConfig.UPDATE_CONDITION_CONFIG)) {
+        String condition = config.getString(JdbcSinkConfig.UPDATE_CONDITION_CONFIG);
+        builder.append(" AND ").append(condition);
+      }
+    } else {
+      if (config.values().containsKey(JdbcSinkConfig.UPDATE_CONDITION_CONFIG)) {
+        String condition = config.getString(JdbcSinkConfig.UPDATE_CONDITION_CONFIG);
+        if (condition != null && !"".equals(condition)) {
+          builder.append(" WHERE ").append(condition);
+        }
+      }
     }
     return builder.toString();
   }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.jdbc.dialect;
 
+import io.confluent.connect.jdbc.sink.JdbcSinkConfig;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
@@ -192,7 +193,11 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
       builder.appendList()
              .delimitedBy(",")
              .transformedBy(transform)
-             .of(nonKeyColumns);
+              .of(nonKeyColumns);
+      if (config.values().containsKey(JdbcSinkConfig.UPDATE_CONDITION_CONFIG)) {
+        String condition = config.getString(JdbcSinkConfig.UPDATE_CONDITION_CONFIG);
+        builder.append(" ").append(condition);
+      }
     }
 
     builder.append(" when not matched then insert(");
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -196,7 +196,9 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
               .of(nonKeyColumns);
       if (config.values().containsKey(JdbcSinkConfig.UPDATE_CONDITION_CONFIG)) {
         String condition = config.getString(JdbcSinkConfig.UPDATE_CONDITION_CONFIG);
-        builder.append(" ").append(condition);
+        if (condition != null && !"".equals(condition)) {
+          builder.append(" where ").append(condition);
+        }
       }
     }
 
@@ -212,6 +214,8 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
     builder.append(")");
     return builder.toString();
   }
+
+
 
   @Override
   protected String sanitizedUrl(String url) {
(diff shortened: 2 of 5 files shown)


demidov163:
[clabot:check]

gharris1727:
@demidov163 Can you clarify the expected functionality of this feature, and how it's different than specifying an additional primary key in your config?

demidov163:
@gharris1727 Thanks for your reply, 
The requested functionality is simple - the is need to check received value with value from DB for same PK.
In my case additional primary key column is leaded to increase 2 time(or more else) DB size.
I don't need to have full story of object changes, for me it's appropriate to have current state of object.

gharris1727:
@demidov163 Thanks for clarifying, let me see if I understand you by giving an example that I think is similar to what you've run into. Let me know if this is an accurate example, or if it differs from your scenario significantly.

The data in the topic has two columns, `before` and `after` and represents a change emitted from another system. In the database, you want only the latest record in a chain of updates to be stored in the database, and all the previous updates discarded. This PR would accomplish this by issuing UPDATE statements, where the `before` in the new record is compared to the `after` of the table, in order to determine which row needs to be updated/discarded.

The alternative of using `after` as a primary-key fails to discard the old records, because it identifies each update as a new and independent record. This strategy would need to be combined with some other cleanup functionality 

Do you have any other fields in your data that could be used as a primary key? It could even be inside the field you want to add this clause for, and you could extract it with an SMT.
For example, if the column was a text blob with an embedded JSON object, and there was a unique ID inside that object, you could parse that object, extract the id, and make the ID a root-level field in the outer object.

Overall this approach seems relatively complicated just to work around data without a suitable primary key, and i'm hesitant to add this functionality for maintainability reasons.

If you have control over the data source, I would recommend adding some unique ID that can be used to identify which updates are from the same source, and should overwrite one another.

demidov163:
@gharris1727, thank you for yoor time, a lot of information in last comment. So i try to give more details about model and particular case.
For example, we have following data  model in DB
Table - product, columns - <id, name, status, changed_date>, key - <id> 

Records in Kafka(in partition) 
1) <1, product_1, status_new, 01-01-2020>
2) <1, product_1, status_active, 02-01-2020>
3) <1, product_1, status_disable, 03-01-2020>

Then if i read records by connector in upsert mode, finally i have  <1, product_1, status_disable, 03-01-2020> record in DB, that OK.

Lets imagine, and producer have some trouble, and records in kafka were produced in different order:
Records in Kafka(in partition) 
1) <1, product_1, status_new, 01-01-2020>
2) <1, product_1, status_disable, 03-01-2020>
3) <1, product_1, status_active, 02-01-2020>

Then, connector read records and in upser mode i will have... <1, product_1, status_active, 02-01-2020> record in DB, and it's wrong for me, because i need **<1, product_1, status_disable, 03-01-2020>** 

Currently, without suggested fix,  i can make composite primary key  - <id, change_date> and record  
<1, product_1, status_disable, 03-01-2020> doesn't  lost, but table size will increase dramatically, and job for clean is not good solution....

My fix solve the problem in query level(merge with condition by change_Date field)

gharris1727:
@demidov163 Thanks for your precise example, I understand much better what's going on. 

It looks like the core issue here is the out of order data in your topic, and not the connector's handling of that mis-ordered topic once it tries to insert it. The connector seems to be functioning as intended, as we don't support re-ordering data after it's read from Kafka.

My other concern with this feature is that it inherently drops the conflicting update, when dropping updates won't give the same result in general as inserting the data in the correct order. Consider the following example, where the status column becomes undefined for the 03-01-2020 record:

1. <1, product_1, status_new, 01-01-2020>
2. <1, product_1, status_active, 02-01-2020>
3. <1, product_1, undefined, 03-01-2020>

In this case, the intended behavior of an undefined column is to leave it at whatever value was previously in the database (status_active), while updating the other columns. If this is re-ordered similar to your example:

1. <1, product_1, status_new, 01-01-2020>
2. <1, product_1, undefined, 03-01-2020>
3. <1, product_1, status_active, 02-01-2020>

The state is left as status_new after the second update, and the third update is dropped. The database is left in (status_new) state, different than the case without reordering.

There are two approaches to fixing this issue:
1. Avoid producing 03-01-2020 until after 02-01-2020 has been written to Kafka
This is usually the approach that people take, but it requires additional design consideration to make sure that the records are
2. After out-of-order data is found, re-produce the later updates to the topic again
This is a flavor of the recovery process used for connectors themselves, where failing to write a record will rewind back to the last known good record and continue from there. Here, this will force the connector to re-apply all of the changes in order. 

For example:

1. <1, product_1, status_new, 01-01-2020>
2. <1, product_1, status_disable, 03-01-2020>
3. <1, product_1, status_active, 02-01-2020>
4. <1, product_1, status_disable, 03-01-2020> 

This will put the database back into the proper state after all of that data has been written.

demidov163:
@gharris1727 Your suggestions look good, let's  discuss it
1) >> Avoid producing 03-01-2020 until after 02-01-2020 has been written to Kafka
This is usually the approach that people take, but it requires additional design consideration to make sure that the records are produced in the correct order. << - This restriction should be implemented in producer system, but we can not change it, we don't want depend on bugs of this system.

 2) >> After out-of-order data is found, re-produce the later updates to the topic again
This is a flavor of the recovery process used for connectors themselves, where failing to write a record will rewind back to the last known good record and continue from there. Here, this will force the connector to re-apply all of the changes in order. << in this case we should request updates in external system and a lot of difficulties could be:
 a) this case can be found in post processing validation, so we received incorrect information, used in our calculations with this record and moreover have some incorrect result. the process of recalculation with **correct** updates may take a lot of time and efforts of our support team. 
b) the external system can be down or under a lot of pressure(highload problems)
c) etc 

So we prefer don't save these wrong updates in further. Related your comment "inherently drops the conflicting update" ->  My suggested fix in this request is only **possibility** of customizing queries. People can use it in their case or no, may be they need use another validation, but meaning of this pull request is use **data from db** for validation.

gharris1727:
@demidov163 Let me clarify a couple of points:

> we don't want depend on bugs of this system.

I'm don't believe that this is a bug in the producer system, it's a design decision. You can only guarantee the ordering of messages by controlling when you give them to the producer, and wait for acknowledgment before continuing. By "design consideration", I mean making sure that you only produce data once you're ready for it to appear in Kafka. If this is not possible on the initial data load (such as if the producers are out of your control, or one key can be produced by multiple producers), then an intermediate re-ordering step is necessary to get your data in the exact order you need to process it in.

>  My suggested fix in this request is only possibility of customizing queries. People can use it in their case or no, may be they need use another validation, but meaning of this pull request is use data from db for validation.

I understand that this feature is optional, and that it enables this new behavior of using the current state of the database to determine it's behavior. By "inherently drops the conflicting updates", I mean this feature as-implemented can only result in the update not being applied, because it's a boolean condition in the WHERE clause.

This is not sufficient to effectively re-order the input updates, as demonstrated in my example above. It's only a half measure with some non-obvious problems to other users, and even though those problems don't appear in your case, doesn't mean that they aren't present. I believe this change is more a case of trying to upstream a workaround than it is a reusable and expressive feature, and I have concerns about maintaining this configuration long-term.

ctapmex:
Порядок у публикатора может быть легко нарушен если в кафку вставляются данные, полученные из оффлайн источников (датчки в лесу) . Сейчас связь на них есть, через час нету. Потом появилась снова. 
Другой пример - нам порядок не важен, но мы не хотим загружать в БД часть статусов. Первичный ключ тут не подойдет.

gharris1727:
> The publisher’s order can be easily violated if data obtained from offline sources

Yes, this is a common pattern in Kafka applications, and it's an architecture problem that you need to solve. Architects have solved this problem with a re-ordering step before it gets to the connector, and a feature like this has not been necessary.

> the order is not important to us, but we do not want to load part of the statuses into the database.

This is a case where the correct solution is to use an SMT, for example: https://docs.confluent.io/current/connect/transforms/filter.html .
This is a reusable module for all connectors that gives you the ability to drop messages before writing them to the database. If you don't care about order, that also means that the state of the database does not affect whether you drop a record, and an SMT should work for your case.

demidov163:
Hi @gharris1727, thanks for answer, but 
>> before it gets to the connector
and 
>> This is a case where the correct solution is to use an SMT, for example: https://docs.confluent.io/current/connect/transforms/filter.html .

Doesn't have possibility to perform _batch_  validation, because if we perform validation one by one record, it makes huge impact to performance

gharris1727:
@demidov163 

> Doesn't have possibility to perform batch validation, because if we perform validation one by one record, it makes huge impact to performance

The Connect Framework already does batching of records over the network, and if you've got logic for individually filtering records, batching that logic together isn't going to change the throughput of the filter.

If you've got performance results that compare these two options i'd like to see them, but I find this argument hard to believe.

demidov163:
>> The Connect Framework already does batching of records over the network, a
Yes, and it's perfect!

>>  and if you've got logic for individually filtering records, 
But Default filter implementation use The filter.condition is a predicate specifying JSON path <https://github.com/json-path/JsonPath>_-- as predicate and this validation is static condition.
But we need more complex validation, compare with DB values, and make this validation as bulk query, using all values from batch records of connector call, for performance reason.

Sorry, but filter implementation doesn't do this.

gharris1727:
@demidov163 
Oh I see that I misunderstood what you meant by "batch validation". I mean "individually filtering records" to mean that the filter + the record determine whether the record is dropped, not the state of the database.

As soon as the state of the database is included in the decision to drop the record, order of records now matters, and the records are no longer filtered individually. You are correct that the Filter SMT does not do this (since it's by definition an SMT). I wanted to separate these two cases, where order does and does not matter.

This is all about correctness and not performance, so I don't think it's valid to argue about performance at this time.

gharris1727:
I'd like to close this PR due to inactivity. I'll close it if there are no further comments for a week.

demidov163:
Hi Greg, As i understand, main concern about this PR that we loose history about events.
If our suggestion about this functional - that it's 
* this additional possibility to affect the merge logic for your custom rules, and you can **check income values with DB values** 
* make validation as bulk validation
* optional functional

 If these points can not fit the architecture requirements, so it's pity, but PR can be closed....

demidov163:
hello, any comments? :)

gharris1727:
@demidov163 We won't be moving forward with this PR at this time.
You are free to fork this connector and support this functionality in your fork, however.

PR #775, opened 2020-01-17
Move setFetchSize to GenericDatabaseDialect

ncliang:
Also refactors OOM integration test, add an OOM IT for MySQL

### DIFF
--- a/src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
+++ b/src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
@@ -22,8 +22,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Integration test for Postgres OOM conditions Expects to be run with -Xmx64M or -Xmx128M for
- * minimum runtime but will behave correctly with any heap size (just with less performance)
+ * Integration test for Postgres OOM conditions.
  */
 @Category(IntegrationTest.class)
 public class PostgresOOMIT extends BaseOOMIntegrationTest {
@@ -55,12 +54,6 @@ public class PostgresOOMIT extends BaseOOMIntegrationTest {
     qb.append("FROM generate_series(1, ");
     qb.append(LARGE_QUERY_ROW_COUNT);
     qb.append(") s(i)");
-    log.info(
-        "Large query will generate "
-            + MAX_MEMORY
-            + " bytes across "
-            + LARGE_QUERY_ROW_COUNT + " rows"
-    );
     return qb.toString();
   }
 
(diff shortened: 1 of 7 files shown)


src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
```diff
@@ -24,87 +22,46 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * Integration test for Postgres OOM conditions
- * Expects to be run with -Xmx64M or -Xmx128M for minimum runtime
- * but will behave correctly with any heap size (just with less performance)
+ * Integration test for Postgres OOM conditions Expects to be run with -Xmx64M or -Xmx128M for
+ * minimum runtime but will behave correctly with any heap size (just with less performance)
```
gharris1727: I think you can promote the second half of this comment up to the base OOMIT class

src/test/java/io/confluent/connect/jdbc/source/integration/PostgresOOMIT.java
```diff
@@ -24,87 +22,46 @@
...
-  public void testStreamingReads() throws InterruptedException {
-    props.put(JdbcSourceTaskConfig.TABLES_CONFIG, "");
-    props.put(JdbcSourceConnectorConfig.QUERY_CONFIG, LARGE_QUERY);
-    startTask();
-    assertTrue(task.poll().size() > 0);
+    qb.append("' ");
+    qb.append("FROM generate_series(1, ");
+    qb.append(LARGE_QUERY_ROW_COUNT);
+    qb.append(") s(i)");
+    log.info(
```
gharris1727: This is a silly log message, i don't know why i left it in.

ncliang:
Since `setFetchSize` is just a [hint](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setFetchSize-int-) to the jdbc driver about the size we would like to fetch, I think we can set it to the batch size for all dialects to show our intent to process one batch of results at a time.

In practice, implementations often require this in conjunction with other settings to do streaming. An example is `useCursorFetch=true` for MySQL. This may come at a cost on the server side to use cursors. Instead of forcing this on users in code, I think it would be better to inform users through documentation about this tradeoff. Maybe an entry in FAQ wiki talking about how to resolve OOM exception when reading large tables.

Thoughts?

gharris1727:
>Since setFetchSize is just a hint to the jdbc driver about the size we would like to fetch, I think we can set it to the batch size for all dialects to show our intent to process one batch of results at a time.

SGTM, it sounds like this isn't going to break anything if the driver doesn't support it. It can only improve the memory behavior of the connector, so it's a good addition.

ncliang:
Good catch on the GPL license. We could maybe use `mariadb-java-client` which is LGPL. From what I can tell, LGPL seems to allow linking in from Apache. IANAL, so... if we want to be extra safe, we could just exclude the connector entirely and require the user to install the connector before running the MySQL IT.

PR #784, opened 2020-02-03
MINOR: Only apply fetchsize to source dialect

ncliang:
Fixes the build.


src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -339,7 +342,9 @@ public PreparedStatement createPreparedStatement(
    * @throws SQLException the error that might result from initialization
    */
   protected void initializePreparedStatement(PreparedStatement stmt) throws SQLException {
-    stmt.setFetchSize(config.getInt(JdbcSourceConnectorConfig.BATCH_MAX_ROWS_CONFIG));
+    if (batchMaxRows > 0) {
```
gharris1727: WDYT about putting the `config instanceof JdbcSourceConnectorConfig` here, rather than adding a field with a sentinel value?

ncliang: I actually don't like this pattern of switching behavior based on `instanceof` and wanted to limit it as much as possible to the ctor.

gharris1727: That sounds reasonable to me.

PR #787, opened 2020-02-06
Make metadata allFields ordered consistently with the valueSchema.

norganna:
Currently the use of a HashMap to store the columns loses the ordering of the columns from the schema.

When the create table is generated, the order of the columns is unpredictable.

This change creates a LinkedHashMap which preserves the original column order from the valueSchema and tests to check ordering is functioning correctly.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
@@ -20,6 +20,8 @@ import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -131,9 +133,25 @@ public class FieldsMetadata {
     }
 
     final Map<String, SinkRecordField> allFieldsOrdered = new LinkedHashMap<>();
-    for (Field field : valueSchema.fields()) {
-      String fieldName = field.name();
+    for (String fieldName : JdbcSinkConfig.DEFAULT_KAFKA_PK_NAMES) {
       if (allFields.containsKey(fieldName)) {
+        allFieldsOrdered.put(fieldName, allFields.get(fieldName));
+      }
+    }
+
+    if (valueSchema != null) {
+      for (Field field : valueSchema.fields()) {
+        String fieldName = field.name();
+        if (allFields.containsKey(fieldName)) {
+          allFieldsOrdered.put(fieldName, allFields.get(fieldName));
+        }
+      }
+    }
+
+    ArrayList<String> fieldKeys = new ArrayList<>(allFields.keySet());
+    Collections.sort(fieldKeys);
+    for (String fieldName : fieldKeys) {
+      if (!allFieldsOrdered.containsKey(fieldName)) {
         allFieldsOrdered.put(fieldName, allFields.get(fieldName));
       }
     }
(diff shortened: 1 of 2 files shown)


norganna:
@gharris1727 This is a more complete and accurate replacement for #763.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -129,7 +130,15 @@ public static FieldsMetadata extract(
       );
     }
 
-    return new FieldsMetadata(keyFieldNames, nonKeyFieldNames, allFields);
+    final Map<String, SinkRecordField> allFieldsOrdered = new LinkedHashMap<>();
```
gharris1727: By separating allFields and allFieldsOrdered, you're removing the PK-first ordering of the columns in favor of whatever order the record specifies them in. I think this might violate some assumptions that we have downstream.

I'm also concerned about what happens when allFields contains a key that's not in the record (such as topic/partition/offset), since these are added in a previous step to allFields, but never appear in allFieldsOrdered.

I think the build failures may be linked to this change in semantics.

norganna: Ok, then perhaps this is not the right approach, since the previous method will not preserve original column ordering in the destination table (as PK's will appear first).

Perhaps the approach needed is for the query builder to not use the order from allFields :/

norganna: I have committed additional code to make sure all fields are added in a determinate order to the `allFieldsOrdered` map.

Looks like tests are completing now.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -129,7 +132,31 @@ public static FieldsMetadata extract(
...
+    for (String fieldName : JdbcSinkConfig.DEFAULT_KAFKA_PK_NAMES) {
+      if (allFields.containsKey(fieldName)) {
+        allFieldsOrdered.put(fieldName, allFields.get(fieldName));
+      }
+    }
```
norganna: This will add the default PK fields first, the ordering is determinate because the DEFAULT_KAFKA_PK_NAMES will always be the same.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -129,7 +132,31 @@ public static FieldsMetadata extract(
...
+
+    if (valueSchema != null) {
+      for (Field field : valueSchema.fields()) {
+        String fieldName = field.name();
+        if (allFields.containsKey(fieldName)) {
+          allFieldsOrdered.put(fieldName, allFields.get(fieldName));
+        }
+      }
+    }
```
norganna: This will add the fields from the source DB in the order from the source table and is thus determinate.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -129,7 +132,33 @@ public static FieldsMetadata extract(
...
+
+    if (allFieldsOrdered.size() < allFields.size()) {
+      ArrayList<String> fieldKeys = new ArrayList<>(allFields.keySet());
+      Collections.sort(fieldKeys);
+      for (String fieldName : fieldKeys) {
+        if (!allFieldsOrdered.containsKey(fieldName)) {
+          allFieldsOrdered.put(fieldName, allFields.get(fieldName));
+        }
+      }
+    }
```
norganna: If there's fields missing (possible if pkMode != RECORD_VALUE and valueSchema = NULL) then add the remaining fields sorted by name for deterministic ordering.

PR #794, opened 2020-02-25
CC-6411: Upgrade jdbc driver for postgres to 42.2.10

kkonstantine:


rhauch:
@kkonstantine, do we not want to backport this further back to older branches than just `5.5.x`?

rhauch:
I guess it makes sense to be more conservative about which versions, and we probably don't want to upgrade the PG driver for older patch releases.

kkonstantine:
Thanks @rhauch 
I agree to keep the target branch as is, given that the upgrade hadn't happened for some time and is now beyond just a bugfix upgrade for the postgres driver. 

Added one commit to update the licenses too. Merging.

Issue #788, opened 2020-02-21
Missing connect.decimal.precision Parameter in Avro Schema for Numeric/Decimal Fields


SteadBytes:
Originally submitted to https://github.com/confluentinc/ksql/issues/4602

**Describe the bug**
Numeric/Decimal fields from PostgreSQL are produce invalid Avro schema, missing `connect.parameters.connect.decimal.precision`.

**To Reproduce**

Please see https://github.com/SteadBytes/ksql-numeric-types-issue for setup and minimal reproduction of the initial issue (found when trying to create KSQL streams). The diagnosis was provided on a response to the KSQL issue https://github.com/confluentinc/ksql/issues/4602#issuecomment-589202625

Fetching the numeric schema definition returns (missing `connect.parameters.connect.decimal.precision`:

```
curl -s http://localhost:8081/subjects/num_types_numeric_val-value/versions/1 |jq '.schema|fromjson'

{
  "type": "record",
  "name": "numeric_val",
  "fields": [
    {
      "name": "id",
      "type": "int"
    },
    {
      "name": "a_value",
      "type": {
        "type": "bytes",
        "scale": 20,
        "precision": 64,
        "connect.version": 1,
        "connect.parameters": {
          "scale": "20"
        },
        "connect.name": "org.apache.kafka.connect.data.Decimal",
        "logicalType": "decimal"
      }
    },
    {
      "name": "write_date",
      "type": {
        "type": "long",
        "connect.version": 1,
        "connect.name": "org.apache.kafka.connect.data.Timestamp",
        "logicalType": "timestamp-millis"
      }
    }
  ],
  "
```

**Expected behavior**

**Actual behaviour**
Fetching the numeric schema should return:

```json
{
  "type": "record",
  "name": "numeric_val",
  "fields": [
    {
      "name": "id",
      "type": "int"
    },
    {
      "name": "a_value",
      "type": {
        "type": "bytes",
        "scale": 20,
        "precision": 64,
        "connect.version": 1,
        "connect.parameters": {
          "scale": "20",
          "connect.decimal.precision": "64" // <--- HERE
        },
        "connect.name": "org.apache.kafka.connect.data.Decimal",
        "logicalType": "decimal"
      }
    },
    {
      "name": "write_date",
      "type": {
        "type": "long",
        "connect.version": 1,
        "connect.name": "org.apache.kafka.connect.data.Timestamp",
        "logicalType": "timestamp-millis"
      }
    }
  ],
[...]

PR #803, opened 2020-03-03
Provide Avro Decimal schema connect.decimal.precision parameter

SteadBytes:
Addresses https://github.com/confluentinc/kafka-connect-jdbc/issues/788

Schema Registry `AvroData` converter uses a connect.decimal.precision
parameter https://github.com/confluentinc/schema-registry/blob/master/avro-data/src/main/java/io/confluent/connect/avro/AvroData.java#L252
when performing serde for logical Decimal schema. This is also required by KSQL for the same reason https://github.com/confluentinc/ksql/issues/4602#issuecomment-591369585 .

This adds a new
`ConnectDecimal` class to provide this and replace uses of org.apache.kafka.connect.data.Decimal
where appropriate.

I have tested this using the [example repo](https://github.com/SteadBytes/ksql-numeric-types-issue) from the original issue (loading the new jar into the `connect` docker container) - the schema is generated correctly and the KSQL queries no longer produce errors. I have not run integration tests with databases other than PostgreSQL in the example repo as I this change should be independent of database type - please correct me if I'm wrong! :smiley: 

I wasn't sure on the licence comment other than that the maven style check failed without it - please let me know if that needs changing :+1:

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/data/ConnectDecimal.java
+++ b/src/main/java/io/confluent/connect/jdbc/data/ConnectDecimal.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


SteadBytes:
[clabot:check]

jose-jurado:
I found the same issue working with MySQL DB source and sinking to HDFS.
Hive queries produce an error because precision 64 is not supported:

`MetaException(message:org.apache.hadoop.hive.serde2.avro.AvroSerdeException Invalid precision or scale for decimal type)`

Using a custom JdbcSourceConnector (with a similar change to this PR) fixed the problem for me.

SteadBytes:
- Yes, since this hasnt seen any progress upstream we've been using a fork including this patch (and a few others).

PR #804, opened 2020-03-03
#363 - New MySQL Sink Dialect Update to TEXT

MarkC0x:
This is an updated version of the changes originally submitted by @MrTrustworthy for issue #363 with the original PR being #364 

It switches the column type of STRING from VARCHAR(256) to TEXT.


Eli-Goldberg:
Hi, Thanks for the awesome work !
Sorry If the question doesn't make sense - 
Should this land in the confluent platform? If so, when could we expect it?
Thanks !

PR #805, opened 2020-03-04
Added support for arrays of primitives in Postgres dialect

Faqa:
While this won't help with arrays of complex types, it was fairly straightforward to piggyback on pgjdbc's setObject auto-casting to get at least some support for arrays of values in the sink connector.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -276,4 +276,22 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
     }
   }
 
+  @Override
+  protected boolean maybeBindPrimitive(
+      PreparedStatement statement,
+      int index,
+      Schema schema,
+      Object value
+  ) throws SQLException {
+
+    switch (schema.type()) {
+      case ARRAY:
+        statement.setObject(index, value, Types.ARRAY);
+        return true;
+      default:
+        break;
+    }
+    return super.maybeBindPrimitive(statement, index, schema, value);
+  }
+
 }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -16,6 +16,7 @@
 package io.confluent.connect.jdbc.dialect;
 
 import java.util.Map;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
@@ -276,6 +277,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
     }
   }
 
+  @SuppressWarnings("unchecked")
   @Override
   protected boolean maybeBindPrimitive(
       PreparedStatement statement,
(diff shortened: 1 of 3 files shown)


Faqa:
[clabot:check]

KyleMccr:
I'd love to see this integrated!

avgojzack:
Are there plans to merge this? It's referenced by several issues and I'm running into the same thing.

rionmonster:
Just leaving my two cents here as well. I'd love to see this get merged in as it seems to address a very widespread use-case that an audience much bigger than those chiming in here seem to be encountering as well.

Faqa:
Update - I've fixed the merge conflicts, so this really just needs to be approved by someone with write access.

lv-helpastranger:
Can we get this merged please ?

Yosafat1997:
when this will be release?

Yosafat1997:
@Faqa I'm very interesting with this feature support, but unfortunately based from the pom.xml you provide, I affraid it's behind newest version of development. Beside use beta, I want to use this but using version 5.5.1 of current Confluent stable repo.
I want to test it with real postgresql environment. Can you show me how to build the .jar file but refer it into 5.5.1 version? 


Thanks.

pdasgupt:
What is the plan to merge this feature?

pdasgupt:
When is this planned to be merged?

dlwodnjs42:
Is there anyway to speed up the review process? :(

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -410,4 +415,26 @@ public void shouldSanitizeUrlWithCredentialsInUrlProperties() {
         "jdbc:postgresql://localhost/test?user=fred&password=****&ssl=true"
     );
   }
-}
\ No newline at end of file
+
+  @Test
+  @Override
+  public void bindFieldArrayUnsupported() throws SQLException {
+      // Overridden simply to dummy out the test.
```
ncliang, line 422: I think we should instead move the test out of the `BaseDialectTest` and into the dialect specific tests that this test still applies to. The test that still applies to the postgres dialect is one that tests for array of non-primitive type still throws, which should be added here.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -362,6 +363,81 @@ protected void formatColumnValue(
...
+              Collection<Byte> listVal = (Collection<Byte>) value;
+              newValue = new short[listVal.size()];
+              int currIndex = 0;
+              for (Byte currVal : listVal) {
+                ((short[]) newValue)[currIndex++] = currVal;
+              }
+              break;
+            }
+            case INT16:
+              newValue = ArrayUtils.toPrimitive(((Collection<Short>) value)
```
ncliang, line 398: Any reason you had to convert these to unboxed primitives? The documentation you linked above says arrays of boxed primitives should also work. And, if the Object is already an array, the if clause above doesn't check whether the array contains boxed or unboxed primitives. If the unboxed version works, we could get rid of the `toPrimitive()` calls and not have to pull in commons-lang3.

ncliang (commented):
Thanks for the PR @Faqa . In addition to comments and questions below, could you also add a testcase in `PostgresDatatypeIT` to have integration test coverage for the feature?

joshuagrisham:
Hi,
I just wanted to comment to say I think this is a super good idea but for consumers of this it is important to understand it really means that an Array in the Kafka message can be represented from/to an Array column in a PostgreSQL table.  Like here:  https://www.postgresql.org/docs/current/arrays.html
So an array of numbers as a column in a table, for example, can be mapped against a Kafka message's array of numbers.

For supporting complex messages (e.g. Arrays of Structs or Maps or Maps of Arrays of Structs or etc etc etc) then the way I have been handling is through use of custom SMTs in Connect to convert the complex types to String, and then store that String as a text or JSON column in the database table.  From there, you can do a lot of things with the data... for example I have messages coming in to a table using JdbcSinkConnector and my "Array of Structs" column I have converted to a JSON string, then in the database I have created a view on top of the table where the Sink is pushing that uses PostgreSQL JSON functions to create a recordset / table-like view of the detailed JSON data.  

Now when I consume my data, I read the Sink-populated table for the "header" fields, and then the view created from the JSON string for the 0..n "details" that could be associated with each header.
(see here https://www.postgresql.org/docs/current/functions-json.html for some ideas)

I have created a PR for my transform changes to be available in the standard Cast transform as well: https://github.com/apache/kafka/pull/9493 in case it is helpful to anyone else 😃

TheGreatAbyss:
Thank you for creating this.  I used the array case statement you've written for Clickhouse and works as well.

Since all this is doing is converting a KC Schema array type to a Java array type this doesn't seem to be PostGres specific.  Why not change the PR to be directly on the GenericDatabaseDialect class, since all JDBC drivers it interfaces with will expect a Java Array if its respective database supports primitive array data types?  (And if not it will error as expected)

kevinalh:
The absence of this feature is the main blocker for introducing Kafka in my company. Hope it's merged soon!

pdasgupt:
Thank you for merging this to the master. I built the code but can you please help me how we should send an array of floats using json schema.

I tried something like this float[] and also array[float]

{"schema": {"type": "struct","fields": [{"type":"int64","optional":false,"field":"id"},  {"type":"float[]","optional":false,"field":"temp1"}],"optional":false,"name": "temp"},"payload": {"id":100,"temp1":{30.7,31,5}}}

pdasgupt:
Any suggestions on the usage with json please

PR #812, opened 2020-03-12
correctif issue 783

Schmidt96u:
Failed to compile on Linux version due to syntax error in use of backtick rather than double-quote : TEST n°24.


Schmidt96u:
[clabot:check]

PR #815, opened 2020-03-13
feat: Adding an additional configuration property to determine whethe…

### LINKED ISSUE #814: JDBC Sink Connector configuration for recoverable exception on failure
While working with a coworker to debug an issue with our connector failing we came across https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues. When attempting to utilize this functionality it appears to fail because of the following line https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java#L74. It would be nice if there was a configuration parameter to allow a RetriableException to be thrown so that the connector can continue processing happily.

brbrown25:
…r or not a connect exception should be thrown once max retries has been reached. ISSUE-814.

fixes #814

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -106,6 +106,12 @@ public class JdbcSinkConfig extends AbstractConfig {
       "The time in milliseconds to wait following an error before a retry attempt is made.";
   private static final String RETRY_BACKOFF_MS_DISPLAY = "Retry Backoff (millis)";
 
+  public static final String MAX_RETRY_THROW_FAILURE = "max.retries.throw.failure";
+  private static final boolean MAX_RETRY_THROW_FAILURE_DEFAULT = true;
+  private static final String MAX_RETRY_THROW_FAILURE_DOC =
+      "Whether a connect exception should be thrown after max retries is reached.";
+  private static final String MAX_RETRY_THROW_FAILURE_DISPLAY = "Throw ConnectException on failure";
+
   public static final String BATCH_SIZE = "batch.size";
   private static final int BATCH_SIZE_DEFAULT = 3000;
   private static final String BATCH_SIZE_DOC =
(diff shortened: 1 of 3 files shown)


brbrown25:
I believe this change will allow for users to using the dead letter queues error handling outlined in this blog post. https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues. However I'm not super familiar with that functionality so this may need some tweaks.

brbrown25:
@cjohnson-confluent i need to sync this with latest master but any thoughts on this Pr?

brbrown25:
@rhauch this is the pr I referenced in slack

cjohnson-confluent:
> @cjohnson-confluent i need to sync this with latest master but any thoughts on this Pr?

I have no context for this PR. Did you mean to include someone else?

brbrown25:
> > @cjohnson-confluent i need to sync this with latest master but any thoughts on this Pr?
> 
> I have no context for this PR. Did you mean to include someone else?

I was just trying to get someone to take a look at this and saw you on other prs, hence the tagging you. I've just started chatting about this in the slack channel with @rhauch

rhauch:
The problem with Connect's DLQ functionality for sink connectors is that Connect sends a batch of records to the sink connector, and if any of the records are bad the connector has no choice but to throw an exception.

This is why we created [KIP-610](https://cwiki.apache.org/confluence/display/KAFKA/KIP-610%3A+Error+Reporting+in+Sink+Connectors) that will be released in AK 2.6 (hopefully within a few weeks). This will allow sink connectors to decide whether individual records are bad, report any such records (sending them to the DLQ), and then continuing on. I think this is actually the direction we want to go here.

I think the challenge with this PR is that it's an all-or-nothing thing, and sending a retriable exception (after the connector finishes its retries) will tell Connect to retry the entire batch again.

Why not just increase the number of retries (e.g., `max.retries`), which defaults to 10? Wouldn't this have the same effect?

brbrown25:
Ok I think I understand what you’re getting at. I believe I miss understood the way the DLQ works (it’s been a while since I looked at this) and think it sounds like the linked KIP will provide the needed functionality. Thanks for looking at this!

PR #823, opened 2020-03-24
CC-8418: set default timezone for certain drivers

levzem:
this PR attempts to fix the `timezone region not found` issue for certain drivers such as oracle11g 
https://stackoverflow.com/questions/9156379/ora-01882-timezone-region-not-found

Signed-off-by: Lev Zemlyanov <lev@confluent.io>


levzem:
In that case, there is no reasonable way of fixing this error in code because every solution requires changing the worker settings, correct?

levzem:
Closing because of discussion above

PR #828, opened 2020-04-01
MINOR: Retain chain of SQLException exceptions

kkonstantine:
Instead of just keeping the exception names when a series of SQLException exceptions is thrown, keep the exceptions themselves as well within the `ConnectException`.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
@@ -80,18 +80,20 @@ public class JdbcSinkTask extends SinkTask {
           remainingRetries,
           sqle
       );
-      String sqleAllMessages = "";
+      String sqleAllMessages = "Exception chain:" + System.lineSeparator();
       for (Throwable e : sqle) {
         sqleAllMessages += e + System.lineSeparator();
       }
+      SQLException sqlAllMessagesException = new SQLException(sqleAllMessages);
+      sqlAllMessagesException.setNextException(sqle);
       if (remainingRetries == 0) {
-        throw new ConnectException(new SQLException(sqleAllMessages));
+        throw new ConnectException(sqlAllMessagesException);
       } else {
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -80,18 +80,20 @@ public void put(Collection<SinkRecord> records) {
           remainingRetries,
           sqle
       );
-      String sqleAllMessages = "";
+      String sqleAllMessages = "Exception chain:" + System.lineSeparator();
       for (Throwable e : sqle) {
         sqleAllMessages += e + System.lineSeparator();
       }
+      SQLException sqlAllMessagesException = new SQLException(sqleAllMessages);
+      sqlAllMessagesException.setNextException(sqle);
```
gharris1727, line 88: The way I understand `setNextException` is that it's a list of co-equal exceptions, such as the unrelated exceptions that occur together within a batch. As written, this code effectively appends this new exception to the head of the existing exception list, as another co-equal of the existing exceptions.

Does it make more sense for `sqle` to be the cause of the new exception instead of being the head of the list? I think semantically `sqle` is the cause of the exception, since without it, the new exception would not exist. This would also simplify the logic for separating the summary from the main chain of exceptions.

rhauch: @kkonstantine ?

kkonstantine: That was done on purpose. 
Any existing code that would iterate over the `SQLException` and would expect to traverse the chain, will now work, with the addition of the exception we were adding so far. 

If I appended the new chain as a cause, it could be more accurate, but the envelopment is artificial anyways, and users would need to get the cause before they attempt to iterate over the exceptions. 

Given that the summary is artificial, I'd suggest keeping this as is now.

PR #838, opened 2020-04-17
CC-8823: Add support for the JDBC sink to write to views in addition to tables

rhauch:
This change allows the JDBC sink to optionally write to tables (by default), views, or a combination.
A new low-importance `table.types` configuration property has been added to the JDBC sink config, with `table` and/or `view` as allowed values, with `table` being the default to maintain backward compatibility when people upgrade. Note that the JDBC sink connector does not attempt to alter a view that does not match the sink records' schemas, and in such cases the sink connector always fails.

The changes are backward compatible (mostly through avoiding changes to signatures), so any connector that reused the JDBC sink classes should automatically work with these changes. Any connector that extends the generic dialect and should get all of the functionality for free.


rhauch:
Closed in favor of #839

PR #839, opened 2020-04-17
CC-8823: Add support for the JDBC sink to write to views in addition to tables

rhauch:
This change allows the JDBC sink to optionally write to tables (by default), views, or a combination. A new low-importance `table.types` configuration property has been added to the JDBC sink config, with `table` and/or `view` as allowed values, with `table` being the default to maintain backward compatibility when people upgrade. Note that the JDBC sink connector does not attempt to alter a view that does not match the sink records' schemas, and in such cases the sink connector always fails.

The changes are backward compatible (mostly through avoiding changes to signatures), so any connector that reused the JDBC sink classes should automatically work with these changes. Any connector that extends the generic dialect and should get all of the functionality for free.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -166,9 +166,10 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     this.jdbcUrl = config.getString(JdbcSourceConnectorConfig.CONNECTION_URL_CONFIG);
     this.jdbcUrlInfo = DatabaseDialects.extractJdbcUrlInfo(jdbcUrl);
     if (config instanceof JdbcSinkConfig) {
+      JdbcSinkConfig sinkConfig = (JdbcSinkConfig) config;
       catalogPattern = JdbcSourceTaskConfig.CATALOG_PATTERN_DEFAULT;
       schemaPattern = JdbcSourceTaskConfig.SCHEMA_PATTERN_DEFAULT;
-      tableTypes = new HashSet<>(Arrays.asList(JdbcSourceTaskConfig.TABLE_TYPE_DEFAULT));
+      tableTypes = sinkConfig.tableTypeNames();
       quoteSqlIdentifiers = QuoteMethod.get(
           config.getString(JdbcSinkConfig.QUOTE_SQL_IDENTIFIERS_CONFIG)
       );
@@ -384,6 +385,8 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   public List<TableId> tableIds(Connection conn) throws SQLException {
     DatabaseMetaData metadata = conn.getMetaData();
     String[] tableTypes = tableTypes(metadata, this.tableTypes);
+    String tableTypeDisplay = displayableTableTypes(tableTypes, ", ");
+    log.debug("Using {} dialect to get {}", this, tableTypeDisplay);
 
     try (ResultSet rs = metadata.getTables(catalogPattern(), schemaPattern(), "%", tableTypes)) {
       List<TableId> tableIds = new ArrayList<>();
@@ -396,6 +399,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
           tableIds.add(tableId);
         }
       }
+      log.debug("Used {} dialect to find {} {}", this, tableIds.size(), tableTypeDisplay);
       return tableIds;
     }
   }
@@ -433,6 +437,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
       DatabaseMetaData metadata,
       Set<String> types
   ) throws SQLException {
+    log.debug("Using {} dialect to check support for {}", this, types);
     // Compute the uppercase form of the desired types ...
     Set<String> uppercaseTypes = new HashSet<>();
     for (String type : types) {
@@ -450,7 +455,9 @@ public class GenericDatabaseDialect implements DatabaseDialect {
         }
       }
     }
-    return matchingTableTypes.toArray(new String[matchingTableTypes.size()]);
+    String[] result = matchingTableTypes.toArray(new String[matchingTableTypes.size()]);
+    log.debug("Used {} dialect to find table types: {}", this, result);
+    return result;
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -550,7 +550,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
         tableId.catalogName(),
         tableId.schemaName(),
         tableId.tableName(),
-        new String[]{"TABLE"}
+        tableTypes
     )) {
       final boolean exists = rs.next();
       log.info(
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -62,6 +62,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 import io.confluent.connect.jdbc.dialect.DatabaseDialectProvider.FixedScoreProvider;
 import io.confluent.connect.jdbc.sink.JdbcSinkConfig;
@@ -88,6 +89,7 @@ import io.confluent.connect.jdbc.util.JdbcDriverInfo;
 import io.confluent.connect.jdbc.util.QuoteMethod;
 import io.confluent.connect.jdbc.util.TableDefinition;
 import io.confluent.connect.jdbc.util.TableId;
+import io.confluent.connect.jdbc.util.TableType;
 
 /**
  * A {@link DatabaseDialect} implementation that provides functionality based upon JDBC and SQL.
@@ -533,7 +535,10 @@ public class GenericDatabaseDialect implements DatabaseDialect {
       Connection connection,
       TableId tableId
   ) throws SQLException {
-    log.info("Checking {} dialect for existence of table {}", this, tableId);
+    DatabaseMetaData metadata = connection.getMetaData();
+    String[] tableTypes = tableTypes(metadata, this.tableTypes);
+    String tableTypeDisplay = displayableTableTypes(tableTypes, "/");
+    log.info("Checking {} dialect for existence of {} {}", this, tableTypeDisplay, tableId);
     try (ResultSet rs = connection.getMetaData().getTables(
         tableId.catalogName(),
         tableId.schemaName(),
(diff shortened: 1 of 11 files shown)


rhauch:
System test with all the JDBC connector tests, using this PR branch:
https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/3728/

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -212,9 +215,23 @@
...
+      + TableType.TABLE + "`` and ``" + TableType.VIEW + "`` is allowed. Not all databases "
+      + "support writing to views, and when they do the the sink connector will fail if the "
+      + "view definition does not match the records' schemas (regardless of ``"
+      + AUTO_EVOLVE + "``).";
+
   private static final EnumRecommender QUOTE_METHOD_RECOMMENDER =
       EnumRecommender.in(QuoteMethod.values());
 
+  private static final EnumRecommender TABLE_TYPES_RECOMMENDER =
+      EnumRecommender.in(TableType.values());
```
wicknicks, line 233: want to add this to the identical source config too?

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -128,26 +129,50 @@ boolean amendIfNecessary(
...
         throw new ConnectException(
-            "Cannot ALTER to add missing field " + missingField
-            + ", as it is not optional and does not have a default value"
+            String.format(
+                "%s %s is missing fields (%s) and ALTER %s is unsupported",
+                type.capitalized(),
+                tableId,
+                missingFields,
+                type.jdbcName()
+            )
```
wicknicks, line 147: nice check! since the view schema cannot be altered by the Connector, should we not ignore `auto.evolve` if the table to be written to is a view, and simply drop all the new fields? or have a validation check that fails config validation if we try to set auto.enable when table.types includes views (which may be too strict actually)? either way, I feel it doesn't make sense to bundle view and table updates in a single deployed connector. users will have to spin up different connectors: one for views that can manage schema evolution, and the other one where it is just not supported.

WDYT?

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
Cyril-Engels: Source connector supports more types. Might need a different `values`.

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
rhauch: When records for this "table" are being sent to a view, the proposal is to require that the view exactly match the record value fields, per the check above. IOW, in that case we *are* ignoring the `auto.evolve` setting.

Maybe I don't understand what you're asking here.

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -128,26 +129,50 @@ boolean amendIfNecessary(
...
+          type.capitalized(),
           tableId,
           missingFields
       ));
     }
 
     final List<String> amendTableQueries = dbDialect.buildAlterTable(tableId, missingFields);
     log.info(
-        "Amending table to add missing fields:{} maxRetries:{} with SQL: {}",
+        "Amending %s to add missing fields:{} maxRetries:{} with SQL: {}",
```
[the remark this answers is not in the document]
rhauch: Good catch.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
rhauch: @Cyril-Engels is correct. Plus, I want to *not* change the source connector in any way.

src/test/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialectTest.java
```diff
@@ -118,6 +125,37 @@ protected GenericDatabaseDialect newDialectFor(
...
+
+  protected GenericDatabaseDialect newSinkDialectFor(Set<String> tableTypes) {
+    if (tableTypes != null) {
+      connProps.put(JdbcSinkConfig.TABLE_TYPES_CONFIG, StringUtils.join(tableTypes, ","));
+    } else {
+      connProps.remove(JdbcSinkConfig.TABLE_TYPES_CONFIG);
+    }
+    sinkConfig = new JdbcSinkConfig(connProps);
+    dialect = createDialect(sinkConfig);
+    assertTrue(dialect.tableTypes.containsAll(tableTypes));
```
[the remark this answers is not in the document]
rhauch: ack

src/test/java/io/confluent/connect/jdbc/source/EmbeddedDerby.java
```diff
@@ -27,8 +27,11 @@
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Arrays;
+import java.util.stream.Collectors;
 
 import io.confluent.connect.jdbc.util.BytesUtil;
+import io.confluent.connect.jdbc.util.TableType;
```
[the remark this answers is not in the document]
rhauch: ack

src/test/java/io/confluent/connect/jdbc/source/EmbeddedDerby.java
```diff
@@ -143,6 +146,42 @@ public void dropTable(String name) throws SQLException {
...
+    String statementStr = statement.toString();
+    log.debug("Creating view {} in {} with statement {}", name, this.name, statementStr);
+    stmt.execute(statementStr);
+  }
+
+  /**
+   * Drop a view.
+   * @param name
+   */
+  public void dropView(String name) throws SQLException {
```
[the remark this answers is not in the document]
rhauch: ack. removed the method.

src/test/java/io/confluent/connect/jdbc/source/EmbeddedDerby.java
```diff
@@ -143,6 +146,42 @@ public void dropTable(String name) throws SQLException {
     stmt.execute("DROP TABLE \"" + name + "\"");
   }
 
+  /**
+   * Shorthand for creating a view over a different table
+   * @param name      name of the table
```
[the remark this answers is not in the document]
rhauch: ack. it wasn't tabs, just wrong number of spaces. fixed.

src/test/java/io/confluent/connect/jdbc/util/TableTypeTest.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+package io.confluent.connect.jdbc.util;
+
+import java.util.Arrays;
+import java.util.EnumSet;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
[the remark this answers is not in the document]
rhauch: ack

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
rhauch: Okay, so I talked with @wicknicks offline, and came to these conclusions:

1. If the record has extra fields relative to the table, the connector will use `auto.evolve` to determine whether to attempt to alter the table to match the record or fail with an exception stating that fields are missing from the table. 
2. If the record has extra fields relative to the view, the connector will fail with an exception stating that fields are missing from the view. This reflects the table behavior (with `auto.evolve=false`), since for views `auto.evolve` is ignored as the connector never has enough information to alter the view definition.
3. If the record is missing some columns from the table or view, the connector does not catch this and instead the operations will succeed iff the extra columns have defaults and/or are nullable, but will fail with a runtime exception if the extra columns are required. This is kind of an undocumented feature, and it would actually be somewhat difficult to explain this corner case in the config docs.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -782,7 +804,50 @@ public TableDefinition describeTable(
...
+    String tableTypeDisplay = displayableTableTypes(tableTypes, "/");
+    log.info("Checking {} dialect for type of {} {}", this, tableTypeDisplay, tableId);
+    try (ResultSet rs = connection.getMetaData().getTables(
+        tableId.catalogName(),
+        tableId.schemaName(),
+        tableId.tableName(),
+        tableTypes
+    )) {
+      if (rs.next()) {
+        //final String catalogName = rs.getString(1);
```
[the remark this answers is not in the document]
kkonstantine: ok, thought so. Not a blocker, but maybe I'd wish if it didn't look exactly as code statements if it's only to document the other columns.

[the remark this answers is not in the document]
rhauch: Fixed.

rhauch:
@kkonstantine, @wicknicks, @Cyril-Engels: Could I ask for another round of reviews? I changed the base from `5.1.x` to `5.4.x`, and there were only two minor conflicts related to adjacent lines that had changed on the 5.4.x. 

This is now what I'd like to merge.

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
wicknicks: Thanks for the summary, @rhauch. For points 1 and 3, it makes sense. For 2, though, I don't believe we ignore `auto.evolve` when the destination is a sink. The task crashes with a ConnectException (as seen even in the `PostgresViewIT#testRecordSchemaMoreFieldsThanViewFails` test). 

It's ok to do this, but we just need to set the right expectation. Effectively, users of the connector might have to run separate connectors to sink records to both views and tables. That is only way to enable auto evolution for tables, and disable for the views in the pipeline.

rhauch: > Effectively, users of the connector might have to run separate connectors to sink records to both views and tables.

Remember that `auto.evolve` only applies to tables, and is important only when the columns of *each* table do not match the record. The behavior of writing to views is independent of this setting.

But it is possible to run one connector that writes to views and tables with `auto.evolve=true` and to have the sink connector ALTER tables as needed.

Consider an example where a user runs one JDBC sink connector that sinks to both a table and a view, and they set:
```
auto.evolve=true
table.types=TABLE,VIEW
```
If the record has, say, one more field than what corresponds to the table's columns, the connector will (per `auto.evolve=true`) attempt to ALTER the table to add the missing column. If the database user has privilege to do this, then this should work and the connector should be able to then write the record to the table.

Meanwhile, if the same connector consumed a record that corresponded to the view and all record fields existed in the view, then the connector would be able to write to the view.

I will attempt to address this in a forthcoming DOCS PR.

wicknicks: In your example, what happens if the records that go to the view added new fields. that would trigger an ALTER query, and a subsequent task failure, right?

rhauch: If the sink consumed a record destined for the view, and the view were missing some columns in the record, the sink connector will fail with an error message similar to the following, regardless of the `auto.evolve` setting:
```
VIEW `customers` is missing fields (SinkRecordField{schema=INT64, name='essential_column', isPrimaryKey=false}) and ALTER VIEW is unsupported
```

wicknicks: makes sense. and that's the bit you wanted to add to documentation?

rhauch: Yes. It's impossible to have that much detail in the config docs, and IMO the code is pretty clear considering the order of the checks in this section.

wicknicks: alright cool, as long as the docs are updated to give a hint about this behavior, I think we should be fine. thanks for your time and responses :).

rhauch: Thanks!

PR #844, opened 2020-04-24
CC-8898: Add better error messages when consumed records don’t match connector config

### LINKED ISSUE #789: NPE at io.confluent.connect.jdbc.sink.BufferedRecords.flush(BufferedRecords.java:174)
Thank you for providing a great kafka-connect-jdbc plugin, but when I tried to sink to DB2 in confluent5.4 , there were some NPE throwed out and the sink task is killed and I can,t find the reason, is there possibly a bug here? Could you please have a look at the following stack trace for me? 

[2020-02-21 16:00:20,480] INFO Attempting to open connection #1 to Db2 (io.confluent.connect.jdbc.util.CachedConnectionProvider:87)
[2020-02-21 16:00:20,689] INFO JdbcDbWriter Connected (io.confluent.connect.jdbc.sink.JdbcDbWriter:49)
[2020-02-21 16:00:20,701] ERROR WorkerSinkTask{id=test-sink-to-db2-0} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted. (org.apache.kafka.connect.runtime.WorkerSinkTask:559)
java.lang.NullPointerException
        at io.confluent.connect.jdbc.sink.BufferedRecords.flush(BufferedRecords.java:174)
        at io.confluent.connect.jdbc.sink.BufferedRecords.add(BufferedRecords.java:159)
        at io.confluent.connect.jdbc.sink.JdbcDbWriter.write(JdbcDbWriter.java:66)
        at io.confluent.connect.jdbc.sink.JdbcSinkTask.put(JdbcSinkTask.java:74)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:539)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:322)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:224)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:192)
        at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:177)
        at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:227)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:748)
[2020-02-21 16:00:20,703] ERROR WorkerSinkTask{id=test-sink-to-db2-0} Task threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask:179)
org.apache.kafka.connect.errors.ConnectException: Exiting WorkerSinkTask due to unrecoverable exception.
        at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:561)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:322)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:224)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:192)
        at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:177)
        at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:227)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NullPointerException
        at io.confluent.connect.jdbc.sink.BufferedRecords.flush(BufferedRecords.java:174)
        at io.confluent.connect.jdbc.sink.BufferedRecords.add(BufferedRecords.java:159)
        at io.confluent.connect.jdbc.sink.JdbcDbWriter.write(JdbcDbWriter.java:66)
        at io.confluent.connect.jdbc.sink.JdbcSinkTask.put(JdbcSinkTask.java:74)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:539)
        ... 10 more
[2020-02-21 16:00:20,704] ERROR WorkerSinkTask{id=test-sink-to-db2-0} Task is being killed and will not recover until manually restarted (org.apache.kafka.connect.runtime.WorkerTask:180)

rhauch:
Added record validation that checks records key, key schema, value, and value schema to ensure that they match with the `delete.enabled` and `pk.mode` connector configuration properties.

Added unit tests that check all permutations of records with combinations of these configuration properties. Importantly, all of these unit tests were developed initially to identify the current behavior without the new validation, and thus asserted each permutation was either successfully written to a database or failed with one of the following:

* ConnectException (e.g., “PK mode for table 'dummy' is RECORD_KEY, but record key schema is missing”);
* NPEs; or
* SQLExceptions (e.g., “Values not bound to statement”)

The `ConnectException` instances are okay, and even had a relatively useful error message. But there still is a lot more info that could be added.

Then when validation was added, the tests' assertions remained the same, except that the exceptions thrown in the failure cases are always ConnectException with much more useful error messages; NPEs and the aforementioned SQLException no longer occur.

The more useful error messages include information that is essential for the user to understand and correct the problem:
1. the connector name;
1. the related connector configs and what they require of records in terms of key, key schema, value, and value schemas;
1. what about the record did not satisfy the requirements; and 
1. the coordinates (topic, partition, offset) of the problematic record.

For example:

> The JDBC sink connector 'my-connector' is configured with 'delete.enabled=true' and 'pk.mode=record_key' and therefore requires non-null record keys, but found record at (topic='dummy',partition=0,offset=1001,timestamp=null) with a null key and STRUCT key schema.

**This is targeted at `5.4.x` and later branches.**

Fixes #789

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
@@ -55,6 +55,7 @@ public class BufferedRecords {
   private List<SinkRecord> records = new ArrayList<>();
   private Schema keySchema;
   private Schema valueSchema;
+  private RecordValidator recordValidator;
   private FieldsMetadata fieldsMetadata;
   private PreparedStatement updatePreparedStatement;
   private PreparedStatement deletePreparedStatement;
@@ -74,9 +75,11 @@ public class BufferedRecords {
     this.dbDialect = dbDialect;
     this.dbStructure = dbStructure;
     this.connection = connection;
+    this.recordValidator = RecordValidator.create(config);
   }
 
   public List<SinkRecord> add(SinkRecord record) throws SQLException {
+    recordValidator.validate(record);
     final List<SinkRecord> flushed = new ArrayList<>();
 
     boolean schemaChanged = false;
@@ -101,7 +104,7 @@ public class BufferedRecords {
       valueSchema = record.valueSchema();
       schemaChanged = true;
     }
-    if (schemaChanged) {
+    if (schemaChanged || updateStatementBinder == null) {
       // Each batch needs to have the same schemas, so get the buffered records out
       flushed.addAll(flush());
 
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -473,7 +473,7 @@ public class JdbcSinkConfig extends AbstractConfig {
 
   public JdbcSinkConfig(Map<?, ?> props) {
     super(CONFIG_DEF, props);
-    connectorName = ConfigUtils.connectorName(this);
+    connectorName = ConfigUtils.connectorName(props);
     connectionUrl = getString(CONNECTION_URL);
     connectionUser = getString(CONNECTION_USER);
     connectionPassword = getPasswordValue(CONNECTION_PASSWORD);
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -29,6 +29,7 @@ import java.util.stream.Collectors;
 
 import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig;
 
+import io.confluent.connect.jdbc.util.ConfigUtils;
 import io.confluent.connect.jdbc.util.DatabaseDialectRecommender;
 import io.confluent.connect.jdbc.util.DeleteEnabledRecommender;
 import io.confluent.connect.jdbc.util.EnumRecommender;
@@ -451,6 +452,7 @@ public class JdbcSinkConfig extends AbstractConfig {
             RETRY_BACKOFF_MS_DISPLAY
         );
 
+  public final String connectorName;
   public final String connectionUrl;
   public final String connectionUser;
   public final String connectionPassword;
@@ -471,6 +473,7 @@ public class JdbcSinkConfig extends AbstractConfig {
 
   public JdbcSinkConfig(Map<?, ?> props) {
     super(CONFIG_DEF, props);
+    connectorName = ConfigUtils.connectorName(this);
     connectionUrl = getString(CONNECTION_URL);
     connectionUser = getString(CONNECTION_USER);
     connectionPassword = getPasswordValue(CONNECTION_PASSWORD);
@@ -502,6 +505,10 @@ public class JdbcSinkConfig extends AbstractConfig {
       return password.value();
     }
     return null;
+  }
+
+  public String connectorName() {
+    return connectorName;
   }
 
   public EnumSet<TableType> tableTypes() {
--- a/src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright 2018 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.connect.jdbc.sink;
+
+import io.confluent.connect.jdbc.util.StringUtils;
+
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.sink.SinkRecord;
+
+@FunctionalInterface
+public interface RecordValidator {
+
+  RecordValidator NO_OP = (record) -> { };
+
+  void validate(SinkRecord record);
+
+  default RecordValidator and(RecordValidator other) {
+    if (other == null || other == NO_OP || other == this) {
+      return this;
+    }
+    if (this == NO_OP) {
+      return other;
+    }
+    RecordValidator thisValidator = this;
+    return (record) -> {
+      thisValidator.validate(record);
+      other.validate(record);
+    };
+  }
+
+  static RecordValidator create(JdbcSinkConfig config) {
+    RecordValidator requiresKey = requiresKey(config);
+    RecordValidator requiresValue = requiresValue(config);
+
+    RecordValidator keyValidator = NO_OP;
+    RecordValidator valueValidator = NO_OP;
+    switch (config.pkMode) {
+      case RECORD_KEY:
+        keyValidator = keyValidator.and(requiresKey);
+        break;
+      case RECORD_VALUE:
+      case NONE:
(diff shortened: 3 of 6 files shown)


src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
```diff
@@ -0,0 +1,213 @@
...
+    // Compose the validator that may or may be NO_OP
+    return keyValidator.and(valueValidator);
+  }
+
+  static RecordValidator requiresValue(JdbcSinkConfig config) {
+    return record -> {
+      if (record.value() == null) {
+        throw new ConnectException(
+            String.format(
+                "The JDBC sink connector '%s' is configured with '%s=%s' and '%s=%s' "
```
gharris1727, line 79: And similar for other usages.
```suggestion
                "Sink connector '%s' is configured with '%s=%s' and '%s=%s' "
```

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
```diff
@@ -0,0 +1,213 @@
...
+    };
+  }
+
+  static RecordValidator requiresKey(JdbcSinkConfig config) {
+    return record -> {
+      if (record.key() == null) {
+        throw new ConnectException(
+            String.format(
+                "The JDBC sink connector '%s' is configured with '%s=%s' and '%s=%s' "
+                + "and therefore requires non-null record keys, "
```
gharris1727, line 149: WDYT about combining all of these throws together and parameterizing the message?
It looks like all of these functions are identical except for the `if` condition and this line in the error message.
```suggestion
                + "and %s, "
```

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
```diff
@@ -0,0 +1,213 @@
+/*
+ * Copyright 2018 Confluent Inc.
```
Cyril-Engels, line 2:
```suggestion
 * Copyright 2020 Confluent Inc.
```

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
```diff
@@ -0,0 +1,213 @@
...
+      if (record.value() == null) {
+        throw new ConnectException(
+            String.format(
+                "The JDBC sink connector '%s' is configured with '%s=%s' and '%s=%s' "
+                + "and therefore requires non-null record values that are structs, "
+                + "but found record at (topic='%s',partition=%d,offset=%d,timestamp=%d) "
+                + "with a %s value and %s value schema.",
+                config.connectorName(),
+                JdbcSinkConfig.DELETE_ENABLED,
+                config.deleteEnabled,
```
Cyril-Engels, line 85: Since `JdbcSinkConfig.DELETE_ENABLED` only matters to key, I feel it would add to confusion in the error message.

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
```diff
@@ -0,0 +1,213 @@
...
+                JdbcSinkConfig.PK_MODE,
+                config.pkMode.toString().toLowerCase(),
+                record.topic(),
+                record.kafkaPartition(),
+                record.kafkaOffset(),
+                record.timestamp(),
+                StringUtils.valueTypeOrNull(record.value()),
```
Cyril-Engels, line 92: `record.value()` is always null, maybe we can just hardcode null in error string. Similar for other error messages.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -101,7 +104,7 @@ public BufferedRecords(
       valueSchema = record.valueSchema();
       schemaChanged = true;
     }
-    if (schemaChanged) {
+    if (schemaChanged || updateStatementBinder == null) {
```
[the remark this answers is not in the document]
rhauch: Yes, this will fix #789

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -471,6 +473,7 @@
 
   public JdbcSinkConfig(Map<?, ?> props) {
     super(CONFIG_DEF, props);
+    connectorName = ConfigUtils.connectorName(this);
```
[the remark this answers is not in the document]
rhauch: ack

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
rhauch: I tried that, and it made the code more difficult to understand (since the entire error messages were never constructed in one place), and searching the code for an error message becomes more difficult. I think the clarity outweighs any duplicate trivial code.

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
rhauch: Why not keep the code the same as in other places? Yes its null, but isn't that okay?

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
rhauch: I originally had slightly different error messages for the different cases, but I felt like each message was not clearly stating all of the requirements, which means users might need to try multiple things before they got the connector to work. By putting all of the configuration information in the error message, they're more likely to see everything that affects record validation.

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
gharris1727: This is best fixed in the pom with a year bump, and probably noisy enough to push to a different PR.

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
gharris1727: When I see so much similar code, I assume that there are more differences than there actually are, and play spot-the-difference. I don't need to do this if there is less code.

w.r.t. searching for error messages: The error message is already mostly un-searchable with so much string formatting. If this is valuable (i think it is), then maybe we can take another step to make this more searchable:

If the message had the parameterized message first, such as "Found record with null record keys: The JDBC sink connector is configured...." then I can search for the phrase before the colon, and find a static string in the class.

src/main/java/io/confluent/connect/jdbc/sink/RecordValidator.java
rhauch: TIL: you can have different copyright dates for each file, as long as it's the same as or more recent than the project's inception year.

IIRC, we actually don't want to change all of the files at once.

rhauch:
@gharris1727 and @Cyril-Engels: I think I've addressed all of your comments/suggestions.

PR #866, opened 2020-06-08
CC-8750: Changed the dialects to sanitize all URL properties that case-insensitively contain “password”

rhauch:
Now sanitizing any URL property that case-insensitively contains (and not just equal to) `password`. Also added a few dialect-specific tests to ensure that different dialects are sanitizing DBMS-specific properties.

Applying to 5.0.x and newer.


rhauch:
Thanks, @gharris1727. I've updated the regex to account for property names that contain `password` but end in other characters, by restricting the suffix to not contain the `&` or `=` characters. This way, `passwordmy-password&something-else=foo...` will not sanitize `something-else`. I've added this case in the unit test.

PR #870, opened 2020-06-17
CC-9567: Allow users to define extra connection properties individually rather than only through JDBC URI

rhauch:
Extra connection properties are those that are prefixed with `connection.` and that are not already predefined by the connector’s ConfigDef. These properties are added by default in the GenericDatabaseDialect, since the existing `addConnectionProperties(…)` method was not explicitly overridden by any known connector — and if any did but did not call `super. addConnectionProperties(…)`, they just wouldn’t get this feature by default.

Modified one of the MySQL integration tests to pass a connection property via the extra properties rather than via the JDBC URI.

This is to be merged/applied only in the `master` branch for inclusion in the upcoming 6.0.0 release.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -308,9 +308,15 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   /**
    * Add or modify any connection properties based upon the {@link #config configuration}.
    *
-   * <p>By default this method does nothing and returns the {@link Properties} object supplied as a
-   * parameter, but subclasses can override it to add/remove properties used to create new
-   * connections.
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -319,6 +325,14 @@ protected JdbcDriverInfo createJdbcDriverInfo(Connection connection) throws SQLE
    *     should be returned; never null
    */
   protected Properties addConnectionProperties(Properties properties) {
+    // Get the set of config keys that are known to the connector
+    Set<String> configKeys = config.values().keySet();
+    // Add any configuration property that begins with 'connection.` and that is not known
+    config.originalsWithPrefix("connection.").forEach((k,v) -> {
```
ncliang, line 331: Pull "connection." into private static final?

Issue #81, opened 2016-06-03
Support for uuid type in postgres


cddr:
Tables with columns of type UUID in postgres cannot be "connected" by just listing the table in the whitelist because the UUID type is not handled by the converter.

One workaround is to use a SQL query instead and convert UUIDs to strings instead but obviously this isn't ideal.

xiongtx:
One workaround is to use a [query](http://docs.confluent.io/3.2.0/connect/connect-jdbc/docs/source_config_options.html#mode) that converts the `UUID` to e.g. a `VARCHAR`:

E.g. in Postgres:

```sql
SELECT
    test.id::varchar,
...
```

cddr:
Another option (especially if you have a lot of tables that use unsupported types), is to create a "kafka_connect" schema or something, and use it to house views that do the conversion described above (and all the other "exotic" datatypes that remain unsupported like boolean and json).

The root problem is that many of these types look the same as far as jdbc is concerned. They all come through as `Type.Other` so the DataConverter does not have enough information to know how to handle it. Obviously there might be many different types that show up under this so it's not a trivial fix.

Would it be acceptable to add a map to the job config that provided additional information about how to convert columns whose type is unknown to the native data converter?

We could do this at a global level like...

{ "typeConverters" => { "id" => OtherToString,
                                         "deleted" => OtherToBoolean } }

...and if (for whatever reason, "id" is not *always* a UUID), on a per-table level like...

{ "typeConverters" =>
      { "foo_table" => { "id" => OtherToString,
                                    "deleted" => OtherToBoolean } },
      {  "bar_table" => { "deleted" => OtherToBoolean,
                                     "request_info" => OtherToString } } }

This would also resolve #35

gwenshap:
If I understand correctly, the discussion is around where/how to add UUID support. 
Two reasonable work-arounds were suggested above, but I want to discuss a solution that doesn't require changes to Postgres schemas.

There are basically 2 places that UUID awareness could be added:
1. Connect data model
2. JDBC Connector

If we add UUID as a new data type in the Connect data model, we can also modify Converters to support it when serializing/de-serializing. The problem there is that UUID type isn't all that common - Avro and JSON don't have a type for UUID AFAIK and neither do many data stores.

[...]

Issue #46, opened 2016-02-23
Please support Postgres jsonb


understructure:
Getting this warning for a table with a JSONB column:

`WARN JDBC type 1111 not currently supported`

I haven't tried regular JSON.

+1 for supporting Postgres boolean type:

```
WARN Ignoring record due to SQL error: (io.confluent.connect.jdbc.JdbcSourceTask:73)
org.postgresql.util.PSQLException: Bad value for type byte : f
```

Thanks in advance!

vikasdp:
+1

ryananguiano:
Same issue

robconrad:
+1

Oktaz:
+1

dandanmylady:
+1

khgitting:
I'd also like to see this. Is this another issue that will be easier with https://github.com/confluentinc/kafka-connect-jdbc/issues/303 ?

venergiac:
+1

Deninc:
+1

minh5:
Running into this as well, +1

rhauch:
Thanks for the suggestion, and all the +1s. The naive approach to solving this is to treat `json` and `jsonb` as a string, which would essentially encode the JSON into a single string field. This would work across all converters (not just JSON). Is this sufficient? I've already incorporated this into the [RDBMS-specific dialects|https://github.com/confluentinc/kafka-connect-jdbc/pull/333/files#diff-9a03fcf4950ba15217add89d2a4664a9R113] in #333 (for issue #303), at least on the source connector side.

OTOH, extracting the JSON content into a Connect `Struct` or `Map` is also possible, but likely has other implications and complications with respect to schemas. Here, any variation in JSON structure would cause schema changes that could make things messy, unless we have a "smart schema inference" mechanism that would compute a schema that covers all JSON structures seen so far. But this also has lots of risk and complications.

austinnichols101:
It would be sufficient for my current use-cases.

akashdeep01:
+1

matthewrj:
Currently when using `JdbcSinkConnector`, it attempts to write `jsonb` data as `VARCHAR`.
Connect config:
```
{
    "connector.class": "io.confluent.connect.jdbc.JdbcSinkConnector",
    "table.name.format": "messages",
    "tasks.max": "1",
    "topics": "topic1",
    "transforms": "HoistPayload",
    "transforms.HoistPayload.field": "message_json",
[...]

PR #871, opened 2020-06-18
Explicitly set the PG type when binding a field during sink

jharting:
In a project I am working on we're using the kafka-connect-jdbc connector to materialize an event stream into a PostgreSQL database.

We run into a problem where the incoming fields of type STRING cannot be written to PostgreSQL columns of type uuid, jsonb, etc. The problem manifests itself as:

```
Caused by: org.postgresql.util.PSQLException: ERROR: column "id" is of type uuid but expression is of type character varying
  Hint: You will need to rewrite or cast the expression.
  Position: 123
        at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2533)
        at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2268)
        ... 19 more
```

A workaround we found was to configure implicit casts in the database.
```
CREATE CAST (varchar as uuid) WITH INOUT AS IMPLICIT;
```

The ugly part is that PostgreSQL does not support scoping such casts to a given table/schema but they are always global. As a result, the workaround for a sink connector limitation then affects the entire database.

Therefore, we instead chose to extended the PostgreSqlDatabaseDialect ourselves for it to set the PG type based on the record schema. For example, a message whose key schema is
```
{"type":"string","optional":false, "name": "io.confluent.connect.jdbc.dialect.PostgreSqlDatabaseDialect.pgtype=uuid"}
```

would be written to the database table column of type uuid just fine. The implementation uses org.postgresql.util.PGobject to achieve that.

The idea is that the event stream schema would not necessarily define these PostgreSQL-specific names but an SMT on the sink side would be used to alter the names so that the sink connector can then assign proper types when writing the record to the database

We're using the extended dialect as a workaround for
- https://github.com/confluentinc/kafka-connect-jdbc/issues/46
- https://github.com/confluentinc/kafka-connect-jdbc/issues/81

**This PR is not expected to be merged in its current form**. Instead I am submitting this PR to get feedback from maintainers about:
- is this approach reasonable? are there alternative/better approaches to deal with the current limitation?
- if so, what format should the schema name prefix have?
- what else needs to change for this to be considered for merging?

Thanks in advance

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
             <version>${postgresql.version}</version>
-            <scope>runtime</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>net.sourceforge.jtds</groupId>
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -21,6 +21,7 @@ import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Schema.Type;
+import org.postgresql.util.PGobject;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Time;
 import org.apache.kafka.connect.data.Timestamp;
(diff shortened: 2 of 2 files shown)


jharting:
[clabot:check]

DestroyerAlpha:
@jharting Any updates on this fix?

jharting:
@DestroyerAlpha I never received any response from project maintainers. I think this has been (at least partially for uuid and jsonb fields) implemented in the meantime, though.

DestroyerAlpha:
@jharting I for the love of god could not figure out how to get JSONB fields. I am using Avro serialisation and it throws ` Type property "json" not a valid Avro schema: Could not make an Avro Schema object from json.` Did you figure out how to use it?

PR #875, opened 2020-06-23
CC-9565: Add support for casting stringified UUIDs to `uuid` upon insert/update/upsert in PostgreSQL

rhauch:
## Problem
PostgreSQL is very strict about datatypes, and string literals used in `INSERT` and `UPDATE` statements must be cast to the database type if the statement is to convert/cast the literal to the column type. 

For example, consider the following table definition:
```
CREATE TABLE users (firstName TEXT, lastName TEXT, userId UUID, preferences JSON)
```
PostgreSQL rejects an insert statement such as the following:
```
INSERT INTO users (firstName, lastName, userId, preferences) 
    VALUES 'Jane', 'Doe', '4eb9d40e-c9e0-4524-b242-18d182330fed', '5';
```
The third and fourth string literals must be explicitly cast to `uuid` and `json` types, respectively:
```
INSERT INTO users (firstName, lastName, userId, preferences) 
    VALUES 'Jane', 'Doe', '4eb9d40e-c9e0-4524-b242-18d182330fed'::uuid, '5'::json;
```
or as a prepared statement:
```
INSERT INTO users (firstName, lastName, userId, preferences) 
    VALUES ?, ?, ?::uuid, ?::json;
```

See https://www.postgresql.org/docs/10/datatype-json.html and https://www.postgresql.org/docs/10/datatype-uuid.html for details.

## Solution
This PR modifies the methods on the `DatabaseDialect` interface used to supply the insert, update, and upsert expressions and create the prepared statements to use the table definition (and the corresponding the column definitions). The `GenericDatabaseDialect` class does not use these definitions, but the `PostgreSqlDatabaseDialect` class is modified to add the cast expression (e.g., `::uuid`) to the variable placeholder (e.g, `?`) if the column type matches one of the database types that requires casting, which in this case is either `uuid`, `json`, or `jsonb`.

Modifying the signature of `DatabaseDialect` methods requires some planning. All existing code in the connector is changed to use the methods with the new signature, and default methods delegate to the methods with the older existing signatures for backward compatibility reasons (in case other dialect implementations outside of this project implement these methods). However, the older methods are deprecated to suggest to developers of other projects that implementations of the methods with older signatures should be migrated to the new signatures. The `PostgreSqlDatabaseDialect` overrides the new methods to provide the desired behavior.

This PR adds unit and integration tests that verify this functionality behaves as expected. When run without the modifications, the integration test fails with the same exception noted in [CC-9565](https://confluentinc.atlassian.net/browse/CC-9565), but this same test passes with the proposed changes.

### Does this solution apply anywhere else?
- [ ] yes
- [x] no

There may be other DBMSes that require casting, but that is beyond the scope of this issue.

### Additional tests:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
This PR is targeted to the `master` branch for inclusion in the upcoming `6.0.0` release of the JDBC source/sink connector. It will not be backported.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1439,6 +1439,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   }
 
   @Override
+  @SuppressWarnings("deprecation")
   public String buildInsertStatement(
       TableId table,
       Collection<ColumnId> keyColumns,
@@ -1459,6 +1460,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   }
 
   @Override
+  @SuppressWarnings("deprecation")
   public String buildUpdateStatement(
       TableId table,
       Collection<ColumnId> keyColumns,
@@ -1483,6 +1485,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   }
 
   @Override
+  @SuppressWarnings("deprecation")
   public String buildUpsertQueryStatement(
       TableId table,
       Collection<ColumnId> keyColumns,
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -15,8 +15,10 @@
 
 package io.confluent.connect.jdbc.dialect;
 
+import java.util.Collections;
 import java.util.Map;
 import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
 import org.apache.kafka.connect.data.Schema;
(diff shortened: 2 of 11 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1439,6 +1439,7 @@ protected void free(Clob clob) throws SQLException {
   }
 
   @Override
+  @SuppressWarnings("deprecation")
```
gharris1727, line 1442: Can these be migrated to the new signature before the method is removed, or do they need to stay while the old signature is deprecated?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -239,11 +255,101 @@ protected String getSqlType(SinkRecordField field) {
     }
   }
 
+  @Override
+  protected ColumnDefinition columnDefinition(
```
gharris1727, line 259: What is the significance of this override?

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
rhauch: What do you mean by "these"?

This is mostly a signal to developers to implement the newer method, and IMO there's not really a good reason to remove these old methods (the earliest we could do that is in a future 7.0.0 release).

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
rhauch: I will remove this override. I think it was leftover from a previous approach that didn't work out.

Issue #874, opened 2020-06-23
Required DB2DatabaseDialect to support `Insert or Ignore` Upsert style


RajKumarMondol:
`Connect guarantees "at least once" delivery. Usually, the JDBC sink processes records, inserts them into the backend database, and asynchronously and periodically commits offsets (via normal consumer offset mechanism). However, if anything happens to the Connect worker after records have been inserted into the database but before the offsets are committed, Connect has no knowledge that those records were indeed inserted, and it restarts processing records at the most recently committed offsets. The net effect is that in your case a small number of records will be reprocessed.`
-- comment from @rhauch in issue  #327 

We also faced this issue. The only option was to use `UPSERT`. In our case, we are using the DB2 Database. We had a trigger that was processing on each inserted record and updates the record's status column to **processed**. While `kafka-connect-jdbc` retries with `upsert` it updates that status column back to **pending**;
DB2DatabaseDialect by default supports `INSERT OR UPDATE` but we required `INSERT OR IGNORE`. We have made a modification to `kafka-connect-jdbc` to achieve this.

I have created this issue and planning to create a pull request with different upsert styles and implementing `INSERT OR IGNORE` for DB2DatabaseDialect.

PR #877, opened 2020-06-25
Required DB2DatabaseDialect to support  `Insert or Ignore` Upsert style

RajKumarMondol:
## Problem
We had a trigger that was processing on each inserted record and updates the record's status column to processed. While kafka-connect-jdbc retries with upsert it updates that status column back to pending;
DB2DatabaseDialect by default supports `INSERT OR UPDATE` but we required `INSERT OR IGNORE`

Fix for #874 

## Solution

1. Add UpsertStyle configuration in JdbcSinkConfig
2. Update Db2DatabaseDialect to read UpsertStyle config
3. Update Db2DatabaseDialect to build the merge statement based on the UpsertStyle

- [ x] yes
- [ ] no

##### If yes, where?
1. This UpsertStyle configuration can be used by any other DatabaseDialect to provide support for `INSERT OR UPDATE` and `INSERT OR IGNORE` UpsertStyle

## Test Strategy

1. If UpsertStyle is not provided this will be taken as default which means as it was working before this configuration. So there is no impact on an existing application using `kafka-connect-jdbc`.

- Test config value if this configuration is not provided
- Test config value if this configuration is provided

2. DB2DatabaseDialect to generate the merge statement based on this style.
 
- Test merge sql statement for all type of upsert combination when `upsert.style=insert_or_ignore` 

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

We can merge it to master so that this functionality is available for everyone.  Going forward all other DatabseDialect can also be updated to respect this configuration is possible.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.jdbc.dialect;
 
+import io.confluent.connect.jdbc.sink.JdbcSinkConfig;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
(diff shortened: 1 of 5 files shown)


RajKumarMondol:
[clabot:check]

RajKumarMondol:
Hi @rhauch,
could you review this PR?

RajKumarMondol:
I had signed cla in past and again I have signed the latest versio of the CLA now.
Cla link in email is not reachable should be updated.

Issue #878, opened 2020-06-29
Reuse JDBC connection for all operations on JDBC source connector


ivansjg:
CachedConnectionProvider is used for almost everything, but sometimes is not being used (like in JdbcSourceTask:possibleTablePartitions() when calling to dialect.parseTableIdentifier()), and is requesting a new connection directly to JDBC within the Dialect instance. That means each connector could potentially have more than 2 concurrent connections (as it should have ideally, one per task and one at the worker).
This might not be a problem in almost all cases, but it is a problem when you have a limited number of concurrent connections allowed to the source DB, in that case that limit could be overcome making the connector failing to be started.

With my change, only one connection will be provided at most from the dialect (CachedConnectionProvider is using the dialect under the hood), so each connector will have 2 concurrent connections to the DB as expected (one for the worker and one for each task).

ivansjg:
Pull request related: https://github.com/confluentinc/kafka-connect-jdbc/pull/879

PR #879, opened 2020-06-29
Reuse JDBC connection for all operations on JDBC source connector

ivansjg:
Issue related: https://github.com/confluentinc/kafka-connect-jdbc/issues/878

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -207,8 +207,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     return timeZone;
   }
 
-  @Override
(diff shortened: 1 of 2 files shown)


ivansjg:
[clabot:check]

ncliang:
Thanks for the PR @ivansjg. However, I do not agree with the direction of this PR. Rather than force the reuse of a single connection, connection pooling should be explicitly controlled by the dialect and the generic dialect should default to just create and return a new connection. Since there are many libraries that implement connection pooling, including vendor specific ones like https://docs.oracle.com/cd/B10501_01/java.920/a96654/connpoca.htm , it would be harsh to enforce this 1-connection-per-thread limit globally and bypass any connection pooling that dialects may desire.

szalapski:
@ncliang Can you help me understand how I can implement connection pooling "at the dialect" as you suggest with Oracle? I currently have 38 of these cconnectors and thus 72 connections that stay alive.  I'd really like to get this down.

PR #883, opened 2020-07-07
Adding integration test for JDBC using mysql dialect. 

deepesh1234:
Link to JIRA ticket - 
https://confluentinc.atlassian.net/browse/HCD-329

Added integration test for scenarios like - 
- End to End IT test (success scenario)
- IT test for JDBC unavailability
- IT test for network interruptions

We have used `pumba` to generate network interruptions and service unavailability.


pom.xml
```diff
@@ -178,6 +229,16 @@
                     </compilerArgs>
                 </configuration>
             </plugin>
+            <plugin>
```
C0urante, line 232: Are you sure we need this?

deepesh1234: this plugin was not required, removed it. Thanks

PR #887, opened 2020-07-16
Add the possibility for dialects to provide a SQL type for null values

smaspe:
## Problem

`statement.setObject(index, null);` is discouraged by JDBC, as per [`setObject` documentation](https://docs.oracle.com/javase/10/docs/api/java/sql/PreparedStatement.html#setObject(int,java.lang.Object)), as some dialects may not support it.

## Solution

A `getSqlTypeForSchema(Schema schema)` method is added. This method can be left alone for dialects that don't need it, or overridden when required. Nothing is affected unless a dialect overrides that method.

- [ ] yes
- [ ] no
- [x] I don't understand the question

## Test Strategy

This code has been tested together with a dialect that uses that method and provides SQL types for the different values of `schema`.

##### Testing done:
- [x] Unit tests
- Unit tests have been included in the dialect that use this particular method (not present in this library)
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
This can be included in any future release. If possible in both 6.x and 5.5.x.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1535,7 +1535,12 @@ public class GenericDatabaseDialect implements DatabaseDialect {
       Object value
   ) throws SQLException {
     if (value == null) {
-      statement.setObject(index, null);
+      Integer type = getSqlTypeForSchema(schema);
+      if (type != null) {
+        statement.setNull(index, type);
+      } else {
+        statement.setObject(index, null);
+      }
     } else {
       boolean bound = maybeBindLogical(statement, index, schema, value);
       if (!bound) {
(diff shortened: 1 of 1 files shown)


smaspe:
@kkonstantine any chance of having this reviewed?

ncliang (approved):
Thanks for the PR @smaspe . While the addition of the method that is just used for setting null objects is slightly awkward, I think it is the least disruptive way to add this hook for dialects that do not support `setObject(index, null)`.

smaspe:
Thanks for the review @ncliang ! I agree with your comment, it's a trade-off between looking a bit awkward and having zero impact on any dialect not interested in passing a specific type.

Do you plan on including this in a 5.5.x release ?

ncliang:
@smaspe , as this does not affect existing dialects that are implemented in the connector, I did not consider it a bugfix and thus will not take it back to `5.5.x` or earlier. It will be released with the upcoming `6.0` release.

PR #908, opened 2020-08-21
Incrementing column delay 5.3.3 post

xjrk58:
## Problem

Incrementing mode might fail to identify new rows even for columns that are eventually strictly monotonic. An example would be two transactions in MySQL inserting data with a primary column that is automatically incremented. 

First: `BEGIN; SELECT SLEEP(10); INSERT ....; COMMIT;`
Second: `BEGIN; INSERT ...; COMMIT;`

The first transactions will get assigned lower number, but as it's committed later the row won't be visible soon enough and it will be skipped. 

## Solution

Introduce an option to wait for `poll.interval.ms` before reading a row and committing the offset. This will give database and writer some time to commit the transaction and propagate it to the reader. If this interval is as long as maximum transaction time + maximum propagation time we will be always reading all rows inserted into the table.

- [ ] yes
- [X] no

## Test Strategy

- [x] Check if all existing unit tests still run fine
- [x] Add unit tests to cover new code
- [x] Verify by manual test that the new behavior with flag enabled actually detects all new rows
- [x] Verify that no rows are detected as new twice 

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [x] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -310,6 +310,21 @@ public interface DatabaseDialect extends ConnectionProvider {
    */
   void applyDdlStatements(Connection connection, List<String> statements) throws SQLException;
 
+
+  /**
+   * Build the SELECT prepared statement expression returning maximum id for the given table
+   * and the first column.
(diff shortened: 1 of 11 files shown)


xjrk58:
[clabot:check]

PR #910, opened 2020-08-27
CC-11507: Always parse Date in UTC

ncliang:
## Problem
AK Date logical type expects the native Date object to be in UTC time. It complains loudly if it receives native Date from another timezone: https://github.com/apache/kafka/blob/trunk/connect/api/src/main/java/org/apache/kafka/connect/data/Date.java#L62

## Solution
Parse Date in UTC timezone, ignoring the `db.timezone` specified.

- [ ] yes
- [x] no

## Test Strategy
Updated unittests. Verified end-to-end against postgres and mysql using IT in `connet-ksql-integration`

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [x] System tests
- [ ] Manual tests

## Release Plan

 
Targeting `5.0.x` as the timezone fix was backported very far back. Will merge the fix forward to other branches. Fix will be released with next bugfix release on perspective branches.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1246,12 +1246,14 @@ public class GenericDatabaseDialect implements DatabaseDialect {
 
       // Date is day + month + year
       case Types.DATE: {
-        return rs -> rs.getDate(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
+        return rs -> rs.getDate(col,
+            DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC)));
       }
 
       // Time is a time of day -- hour, minute, seconds, nanoseconds
       case Types.TIME: {
-        return rs -> rs.getTime(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
+        return rs -> rs.getTime(col,
+            DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC)));
       }
 
       // Timestamp is a date + time
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1252,8 +1252,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
 
       // Time is a time of day -- hour, minute, seconds, nanoseconds
       case Types.TIME: {
-        return rs -> rs.getTime(col,
-            DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC)));
+        return rs -> rs.getTime(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
       }
 
       // Timestamp is a date + time
--- a/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskConversionTest.java
+++ b/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskConversionTest.java
@@ -269,7 +269,7 @@ public class JdbcSourceTaskConversionTest extends JdbcSourceTaskTestBase {
   @Test
   public void testTime() throws Exception {
     GregorianCalendar expected = new GregorianCalendar(1970, Calendar.JANUARY, 1, 23, 3, 20);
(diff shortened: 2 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1246,12 +1246,14 @@ protected ColumnConverter columnConverterFor(
...
-        return rs -> rs.getDate(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
+        return rs -> rs.getDate(col,
+            DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC)));
       }
 
       // Time is a time of day -- hour, minute, seconds, nanoseconds
       case Types.TIME: {
-        return rs -> rs.getTime(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
+        return rs -> rs.getTime(col,
+            DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC)));
```
gharris1727, line 1256: I think this change is explicitly incorrect change, because interpreting times as the specified timezone is an explicitly supported functionality of this connector.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1246,12 +1246,14 @@ protected ColumnConverter columnConverterFor(
 
       // Date is day + month + year
       case Types.DATE: {
-        return rs -> rs.getDate(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
+        return rs -> rs.getDate(col,
+            DateTimeUtils.getTimeZoneCalendar(TimeZone.getTimeZone(ZoneOffset.UTC)));
```
gharris1727, line 1250: I think this change may be dangerous, because the behavior of this parameter is non-standard.
Also, I think that timezone conversions for dates _does_ make sense, as two different timezones experience different dates at the same instants.

ncliang: Is that true though? It is 8/27/2020 for me right now here in Seattle, but 8/28/2020 in Taiwan. Same instant in time, same millis since EPOCH, different dates.

Since the `Date` object is just a thin wrapper around millis since EPOCH, and EPOCH is defined in UTC timezone, what timezone the driver decides to use to calculate the start of day for that date is important. The reason we were seeing the hours/minutes when we get to framework Date logical type is likely because we chose the start of day to be start of the day in timezone rather than in UTC.

If we find drivers that are not implementing the timezone calculation properly, we can deal with them individually in dialects. But, I think the general case should hold that we should ask for the Date in UTC if we expect it in UTC at framework level.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
ncliang: Thinking about this more, I actually think this change may be a no-op. For Dates, the timestamp of start of day differs depending on timezone. But, times are just millis since EPOCH and what timezone we ask for them in should not make a difference. I can back out the changes for Time and just keep the one for Date.

ncliang:
Thanks for review @gharris1727 . I have backed out my changes to `Time` which I believe are no-ops. I still believe that the changes to `Date` are necessary though since framework is expecting Dates to contain start-of-day to be the start in UTC and not local timezone.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskConversionTest.java
```diff
@@ -259,7 +269,7 @@ public void testNullableDate() throws Exception {
   @Test
   public void testTime() throws Exception {
     GregorianCalendar expected = new GregorianCalendar(1970, Calendar.JANUARY, 1, 23, 3, 20);
-    expected.setTimeZone(TimeZone.getTimeZone("UTC"));
+    expected.setTimeZone(timezone);
```
ncliang, line 272: Note that these testcases were explicitly validating the objects were in UTC time. They failed once I updated them to test with non-UTC timezone.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
gharris1727: After thinking through the cases, this seems to be the safest change to implement. This does depend on the driver implementation to do the conversion, but any sufficiently strange implementations we can follow up with dialect fixes.

PR #914, opened 2020-09-02
Add a flag to enabling waiting for longer transactions (Fix for #915)

xjrk58:
## Problem (see #915 )

Incrementing mode might fail to identify new rows even for columns that are eventually strictly monotonic. An example would be two transactions in MySQL inserting data with a primary column that is automatically incremented. 

First: `BEGIN; SELECT SLEEP(10); INSERT ....; COMMIT;`
Second: `BEGIN; INSERT ...; COMMIT;`

The first transactions will get assigned lower number, but as it's committed later the row won't be visible soon enough and it will be skipped. 

## Solution

Introduce an option (`incrementing.relaxed.monotonic`) to wait for `poll.interval.ms` before reading a row and committing the offset. This will give database and writer some time to commit the transaction and propagate it to the reader. If this interval is as long as maximum transaction time + maximum propagation time we will be always reading all rows inserted into the table.

- [ ] yes
- [X] no

## Test Strategy

- [ ] Check if all existing unit tests still run fine
- [x] Add unit tests to cover new code
- [x] Verify by manual test that the new behavior with flag enabled actually detects all new rows
- [x] Verify that no rows are detected as new twice 

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [x] System tests
- [x] Manual tests

## Release Plan

 

This feature will be merged to master and it's backwards compatible (flag will be `false` by default).


ncliang:
Thanks for the PR @xjrk58 . I think the `timestamp.delay.interval.ms` flag is what you are looking for.

xjrk58:
@ncliang `timestamp.delay.interval.ms` doesn't work in this case. Also in order to get `timestamp.delay.interval.ms` to work you need to have reliable timestamp. This PR helps in case you are using **incrementing** mode only. It doesn't influence `timestamp` or `incrementing+timestamp` mode.

xjrk58:
In case you do NOT have reliable timestamp there is no way to fetch inserted rows in neither mode.

xjrk58:
@ncliang Also `timestamp.delay.interval.ms` introduces only "delay", which doesn't help at all in case you have multiple transactions with highly variable time between begin and commit even if you have reliable timestamp. In case of MySQL timestamp is generated at the beginning of the transaction, so rows inserted long(er) transactions might be skipped again any mode.

PR #918, opened 2020-09-11
CC-8794: MS SQL Server datetime give repeated messages due to rounding error

SajanaW:

## Problem
Ms Sql Server datetime columns have a precision error that causes our connector to loop on the most recent record. This error cannot be fixed by increasing precision because precision is set at a driver level.

### References
http://www.dbdelta.com/sql-server-2016-and-azure-sql-database-v12-breaking-change/

## Solution
 We've decided to forbid the use of `datetime` for MSSQL Server (post 2016) and this pr includes a runtime validation for it

- [ ] yes
- [ x] no

## Test Strategy
Tested this using IT tests and different MS SQL Drivers locally

##### Testing done:
- [ ] Unit tests
- [ x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,8 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.package.home>target/${project.artifactId}-${project.version}-package</project.package.home>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
+        <connect.utils.version>0.1.11</connect.utils.version>
+        <confluent.licensing.version>0.5.2</confluent.licensing.version>
     </properties>
 
     <repositories>
@@ -82,7 +84,12 @@
             <artifactId>connect-api</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-runtime</artifactId>
+            <version>${kafka.version}</version>
+            <scope>provided</scope>
+        </dependency>
          <!-- JDBC drivers, only included in runtime so they get packaged -->
         <dependency>
             <groupId>org.xerial</groupId>
@@ -163,6 +170,59 @@
             <version>2.5.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.microsoft.sqlserver</groupId>
+            <artifactId>mssql-jdbc</artifactId>
+            <version>8.4.1.jre11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>connect-licensing-extensions</artifactId>
+            <version>${confluent.licensing.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-runtime</artifactId>
+            <version>${kafka.version}</version>
+            <type>test-jar</type>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_${kafka.scala.version}</artifactId>
+            <version>${kafka.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_${kafka.scala.version}</artifactId>
+            <version>${kafka.version}</version>
+            <type>test-jar</type>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <version>${kafka.version}</version>
+            <classifier>test</classifier>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>connect-utils</artifactId>
+            <version>${connect.utils.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.12.2</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
@@ -174,7 +234,6 @@
                 <configuration>
                     <compilerArgs>
                         <arg>-Xlint:all</arg>
-                        <arg>-Werror</arg>
                     </compilerArgs>
                 </configuration>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -234,6 +234,7 @@
                 <configuration>
                     <compilerArgs>
                         <arg>-Xlint:all</arg>
+                        <arg>-Werror</arg>
                     </compilerArgs>
                 </configuration>
             </plugin>
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1641,7 +1641,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   }
 
   @Override
-  public void validateSpecificColumnsTypes(
+  public void validateTimestampColumns(
           ResultSetMetaData rsMetadata,
           List<ColumnId> timestampColumns
   ) throws ConnectException { }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1641,9 +1641,9 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   }
 
   @Override
-  public void validateTimestampColumns(
+  public void validateSpecificColumnTypes(
           ResultSetMetaData rsMetadata,
-          List<ColumnId> timestampColumns
+          List<ColumnId> columns
   ) throws ConnectException { }
 
   protected List<String> extractPrimaryKeyFieldNames(Collection<SinkRecordField> fields) {
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1640,6 +1640,12 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     return Collections.singletonList(builder.toString());
   }
 
+  @Override
+  public void validateSpecificColumnsTypes(
+          ResultSetMetaData rsMetadata,
+          List<ColumnId> timestampColumns
+  ) throws ConnectException { }
+
   protected List<String> extractPrimaryKeyFieldNames(Collection<SinkRecordField> fields) {
     final List<String> pks = new ArrayList<>();
     for (SinkRecordField f : fields) {
--- a/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
@@ -366,10 +366,10 @@ public class SqlServerDatabaseDialect extends GenericDatabaseDialect {
 
     if (versionWithBreakingDatetimeChange()) {
       try {
-        for (int i = 1; i < rsMetadata.getColumnCount(); i++) {
-          if (rsMetadata.getColumnTypeName(i).equals(DATETIME)) {
+        for (int i = 0; i < rsMetadata.getColumnCount(); i++) {
+          if (rsMetadata.getColumnTypeName(i + 1).equals(DATETIME)) {
             for (ColumnId id: timestampColumns) {
-              if (id.name().equals(rsMetadata.getColumnName(i))) {
+              if (id.name().equals(rsMetadata.getColumnName(i + 1))) {
                 throw new ConnectException(
                         "A DATETIME column is configured for " + TIMESTAMP_COLUMN_NAME_CONFIG
                         + " with Sql Server. DATETIME is not supported. Use DATETIME2 instead.");
--- a/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
@@ -64,6 +64,9 @@ public class SqlServerDatabaseDialect extends GenericDatabaseDialect {
   private static final DateTimeFormatter DATE_TIME_FORMATTER =
       DateTimeFormatter.ofPattern(DATE_TIME_FORMAT);
 
+  private static final int MSSQL_2016_VERSION = 13;
+  private static final int PRE_MSSQL_2016_VERSION = 12;
+
   /**
    * The provider for {@link SqlServerDatabaseDialect}.
    */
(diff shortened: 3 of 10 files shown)


src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -143,6 +153,57 @@ protected void createPreparedStatement(Connection db) throws SQLException {
...
+  }
+
+  /**
+   * Do runtime checks on table metadata
+   * If SqlServer dialect (Sql Server 2016 or older), time stamp mode configured,
+   * and time stamp column is datetime, kill task
+   * @param metadata table meta data
+   * @throws ConnectException if run time check failed
+   */
+  private void testForUnsupportedDatabaseColumnTypes(
```
mageshn, line 176: This method should be in the dialect classes.

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -91,6 +94,21 @@ public SqlServerDatabaseDialect(AbstractConfig config) {
     jtdsDriver = jdbcUrlInfo == null ? false : jdbcUrlInfo.subprotocol().matches("jtds");
   }
 
+  public boolean sqlServer2016OrLater() {
```
cyrusv, line 97: Let's call this by "versionWithBreakingDatetimeChange" or something with more context than just the year of the release. There may be multiple releases in that year, and more directly what the function does.

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -143,6 +153,57 @@ protected void createPreparedStatement(Connection db) throws SQLException {
...
+      String schemaName = tableId != null ? tableId.tableName() : null; // backwards compatible
+      ResultSetMetaData metadata = resultSet.getMetaData();
+      testForUnsupportedDatabaseColumnTypes(metadata);
+      schemaMapping = SchemaMapping.create(schemaName, metadata, dialect);
+    }
+  }
+
+  /**
+   * Do runtime checks on table metadata
+   * If SqlServer dialect (Sql Server 2016 or older), time stamp mode configured,
```
cyrusv, line 171: older? I thought newer had the issue?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
cyrusv: +1

src/test/java/io/confluent/connect/jdbc/integration/BaseConnectorIT.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.kafka.connect.runtime.rest.entities.ConnectorStateInfo;
+import org.apache.kafka.connect.util.clusters.EmbeddedConnectCluster;
+import org.apache.kafka.test.IntegrationTest;
+import org.apache.kafka.test.TestUtils;
+import org.junit.experimental.categories.Category;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Category(IntegrationTest.class)
+public abstract class BaseConnectorIT {
```
cyrusv, line 31: Is this not centralized somewhere? It does not seem specific to JDBC at all.

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+    private Map<String, String> props;
+
+    private static final String USER = "sa"; // test creds
+    private static final String PASS = "reallyStrongPwd123"; // test creds
+
+    private Connection connection;
+
+    @ClassRule
+    public static final FixedHostPortGenericContainer mssqlServer =
+            new FixedHostPortGenericContainer<>("microsoft/mssql-server-linux:latest")
```
cyrusv, line 83: nice

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+        //Create table
+        Class.forName("com.microsoft.sqlserver.jdbc.SQLServerDriver");
+        connection = DriverManager.getConnection(MSSQL_URL, USER, PASS);
+        startConnect();
+
+    }
+
+    @After
+    public void close() throws SQLException {
+        connect.deleteConnector(CONNECTOR_NAME);
```
cyrusv, line 99: What happens if you stopConnect() without deleting the connector? Seems like the code for this line and stopConnect should be co-located.

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+            log.info("Successfully created table");
+        } catch (Exception ex) {
+            log.error("Could not create table");
+            throw ex;
+        }
+    }
+
+    private void writeSingleRowWithTimestamp(
+            Timestamp t,
+            boolean stringRecord,
```
cyrusv, line 264: These boolean arguments are kind of un-intuitive, and their implementation is a bit awkward. Maybe more legible, and still good code re-use, to just accept the entire INSERT statement as a String. It may feel like you need to abstract out the actual sql statements, but since we are testing sql implementation details, I think it's fine to surface the call itself.

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+        props.put(CONFLUENT_TOPIC_BOOTSTRAP_SERVERS_CONFIG, connect.kafka().bootstrapServers());
+        props.put(CONFLUENT_TOPIC_REPLICATION_FACTOR_CONFIG, "1");
+
+        return props;
+    }
+
+    private void createDateTimeTable(
+            String timestampColType,
+            String regularColType,
+            boolean timestampModeOnly
```
cyrusv, line 242: I would probably just pass in the CREATE string as an arg instead of this bool, it's kind of hard to read

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+     * Verify that after a number of poll intervals the number of records in topic
+     * still equal to the number of records written to the MSSQL Server instance.
+     *
+     * This test is for a datetime to datetime2 comparison error in MSSQL Server (2016 and after).
+     * Datetime is converted recursively to a higer precision (eg- 3.33 MS would be 3.33333 ),
+     * while java.sql.Timestamp converts by tacking on 0s (3.33000).
+     */
+
+    @Test
+    public void verifyDATETIME2DoesNotKillTask() throws Exception {
```
cyrusv, line 180: I think CAPSLOCK is not standard Java style here

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+                    topic);
+            //Assert that records in topic == NUM_RECORDS_PRODUCED
+            assertEquals(NUM_RECORDS_PRODUCED, records.count());
+
+        }
+        // clean up table
+        deleteTable();
+    }
+
+    private Map<String, String> configProperties(boolean timestampModeOnly) {
```
cyrusv, line 211: Let's pass the actual mode instead of the bool. It looks goofy when you call it with `configProperties(true)`

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+        KAFKA_TOPICS.forEach(topic -> connect.kafka().createTopic(topic, 1));
+
+        // Configure source connector
+        configureAndWaitForConnector();
+
+        java.sql.Timestamp t = java.sql.Timestamp.from(
+                ZonedDateTime.of(2016, 12, 8, 19, 34, 56, 2000000, utc).toInstant()
+        );
+
+        writeSingleRowWithTimestamp(t, true, true);
```
cyrusv, line 130: This is goofy to read, can you make the method sig/params more intuitive to read?

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+     * This test is for a datetime to datetime2 comparison error in MSSQL Server (2016 and after).
+     * Datetime is converted recursively to a higher precision (eg- 3.33 MS would be 3.33333 ),
+     * while java.sql.Timestamp converts by tacking on 0s (3.33000).
+     */
+    // Verify connect error thrown
+    @Test
+    public void verifyDateTimeTSModeKillsTasks() throws Exception {
+        // Setup up props for the source connector
+        props = configProperties(true);
+        createDateTimeTable("DATETIME", "VARCHAR(255)", true);
```
cyrusv, line 118: This is goofy to read, can you make the method sig/params more intuitive to read?

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,314 @@
...
+
+        writeSingleRowWithTimestamp(t, true, true);
+
+        connect.assertions().assertConnectorIsRunningAndTasksHaveFailed(
+                CONNECTOR_NAME,
+                Math.min(KAFKA_TOPICS.size(), TASKS_MAX),
+                "failed to verify that tasks have failed"
+        );
+
+        deleteTable();
```
cyrusv, line 138: This can be in teardown method, so it runs even if there is a failure in the test

cyrusv (changes requested):
Strategy looks good. Two main points on the first pass:
 - Let's be more thoughtful where we put this test, and try to put it in the dialect. Let's read all the methods we might override on the dialect class and pick the appropriate one, or have a discussion if there is no appropriate one.
 - The test methods are kind of hard to read bc the helper methods take these weird bools and string literals. Let's focus more on legibility than re-use. When someone wants to know what we support, they should first look what we test against. And when they read the ITs, it should read like documentation as a primary goal. It is fine to assume the reader knows about SQL since this entire DB is about connecting to SQL, so it's not necessary to hide the sql statements behind a function.

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
SajanaW: I was going to do it but I held back because I thought moving this over to the Dialect would obscure the Connect Exception that would get thrown (because it would not be in the "main thread" of execution if that makes sense). 
Would you agree with me if I leave this function here but move it's logic into the Dialect class (this function would use a bool to determine if it needs to throw a Connect Exception)?

pom.xml
```diff
@@ -163,6 +170,59 @@
             <version>2.5.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.microsoft.sqlserver</groupId>
+            <artifactId>mssql-jdbc</artifactId>
+            <version>8.4.1.jre11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
mageshn, line 179: Why do we need licensing dependency?

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
```diff
@@ -0,0 +1,298 @@
...
+
+        sql = "INSERT INTO " + MSSQL_Table + " (start_time, record) "
+                + "values (?, ?)";
+        PreparedStatement insertStmt = connection.prepareStatement(sql);
+        insertStmt.setTimestamp(1, t);
+        insertStmt.setTimestamp(2, t);
+        executeSQL(insertStmt);
+
+
+        Thread.sleep(Duration.ofSeconds(30).toMillis());
```
cyrusv, line 238: Are you waiting for something specific here? Can we do a wait-and-probe type strategy, instead of just 30 seconds?

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1640,6 +1640,12 @@ public String buildDropTableStatement(
     return Collections.singletonList(builder.toString());
   }
 
+  @Override
+  public void validateSpecificColumnsTypes(
+          ResultSetMetaData rsMetadata,
+          List<ColumnId> timestampColumns
```
cyrusv, line 1646: If this is specific to timestamps, can we change the name to `validateTimestampColumns`, or if it is for all columns, then make this param name more generic?

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -63,6 +67,15 @@
   private static final String DATE_TIME_FORMAT = "yyyy-MM-dd HH:mm:ss.SSSSSSS ZZZZZ";
   private static final DateTimeFormatter DATE_TIME_FORMATTER =
       DateTimeFormatter.ofPattern(DATE_TIME_FORMAT);
+  private static final int MSSQL_2016_VERSION = 13;
+  private static final int PRE_MSSQL_2016_VERSION = 12;
```
cyrusv, line 71: I'm not sure defining this by its year of release is wise. Doesn't it make more sense to use the version number itself, instead of the year? It's more specific.

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -305,6 +338,53 @@ public String buildUpsertQueryStatement(
...
+          List<ColumnId> timestampColumns
+  ) throws ConnectException {
+    if (verifiedSqlServerTimestamp) {
+      return;
+    }
+
+    if (versionWithBreakingDatetimeChange()) {
+      try {
+        for (int i = 0; i < rsMetadata.getColumnCount(); i++) {
+          if (rsMetadata.getColumnTypeName(i + 1).equals(DATETIME)) {
```
cyrusv, line 370: why i+1? Maybe we can make the for loop better, and leave a comment?

cyrusv (commented):
Looks pretty good, just some minor comments

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
SajanaW: The indexing for Columns in rsMetadata is [indexed at 1 ](https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSetMetaData.html#getColumnName(int))(instead of 0). 
I'll leave a comment

src/test/java/io/confluent/connect/jdbc/integration/BaseConnectorIT.java
SajanaW: No, unfortunately not, all IT tests have a Base Connector class somewhere

pom.xml
SajanaW: It's being used in the IT tests to get config keys for some Embedded Cluster config
`import static io.confluent.connect.utils.licensing.LicenseConfigUtil.CONFLUENT_TOPIC_REPLICATION_FACTOR_CONFIG;`
Since these are in test scope it shouldn't matter

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
SajanaW: Thanks for the thoughtful comment! 
I think it should be renamed to `validateTimestampColumns`. 
The pattern here is a TableQuerier uses the dialect classes to verify if the columns selected for the mode (timestamp, incrementing or both) in config are valid against the metadata of the table. So I think the play is to have two no-ops in the `GenericDialectClass`, one `validateTimestampColumns` and one `validateIncrementingColumn` (but we should add the latter when it's needed)

src/test/java/io/confluent/connect/jdbc/source/integration/MSSQLDateTimeIT.java
SajanaW: We're just verifying that when `datetime2` is used the number of rows written to MSSQL is the number rows read from it. The way to do this is just wait for long enough that the `TableQuerier` would have polled the MSSQL Table more than once checking for new records. There's no real wait-and-probe strategy cause in the test `pass` case  no matter how long we wait we should still have the same number of records read! I'll leave a comment

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
SajanaW: Chatted offline about this: MSSQL Server tends to define their versions in terms of release years. Ex MSSQL Server 2016 is used to refer that version more than say MSSQL Server 13. TLDR: Keeping as is

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
mageshn: Actually I like the original generic name....each dialect might different restrictions for columns and this method in the dialect can have logic to check anything specific to the DB, right?

cyrusv (approved):
Great Work!

mageshn (approved):
LGTM for the dialect abstraction concern that I raised 👍

PR #924, opened 2020-09-24
MINOR: Fix exception when failing to read tables

C0urante:
The `ConfigException` class doesn't have a constructor that accepts a child exception, but it does have a constructor that appears to, which accepts [a `String` and an `Object`](https://github.com/apache/kafka/blob/149897976214a769a35cf87eecf9c02221f8d597/clients/src/main/java/org/apache/kafka/common/config/ConfigException.java#L32-L34).

Using that constructor with a message and an exception cause error messages like this to be logged:
```
Invalid value org.postgresql.util.PSQLException: FATAL: password authentication failed for user "fm_service_api_journal_usage" for configuration Couldn't open connection to jdbc:postgresql://db.my-env.local/db
```

The fix here replaces the `ConfigException` with a `ConnectException`, which should fix how errors are logged when the connector has trouble reading the list of tables from the source database.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -26,6 +26,7 @@ import org.apache.kafka.common.config.ConfigDef.Width;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.types.Password;
 import org.apache.kafka.common.utils.Time;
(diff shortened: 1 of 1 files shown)


ncliang (approved):
Thanks @C0urante . I believe the TableRecommender is being removed in master, but the fix should apply to previous branches.

C0urante:
Thanks @ncliang! I've been able to build and test locally successfully; think the Jenkins failure is a Jenkins issue and not related to these changes. Will merge.

PR #926, opened 2020-09-25
Pe 7413/connect image

rajat189:
## Description
To create a custom image with updated jars of JDBC connectors


gharris1727:
@rajat189 Is this intended to be upstreamed at this time?

PR #927, opened 2020-09-29
MINOR: Change TableRecommender to return empty list if unable to connect and obtain list of available tables

rhauch:
## Problem
The source connector's `TableRecommender` does not have visibility to the non-parsed configuration properties (e.g., those that don't have a ConfigDef). This means that any extra `connection.*` properties enabled via #870 are simply not visible to any recommenders.

## Solution
Rather than fail if the `TableRecommender` is not able to connect and list available tables, simply return an empty list. Any connection-related properties will be reported on the connection-related properties themselves, so failing with an error is really not helpful and, in the case of extra properties, the recommender actually prevents otherwise valid configurations from being used.

TBD: The `TableRecommender` has been only semi-functional for some time, and it can lead to excessive validation times. It may be better to simply remove this altogether.

- [ ] yes
- [x] no

## Test Strategy
Existing unit tests

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -19,6 +19,7 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.ZoneId;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


rhauch:
See #930 for a discussion about this approach versus removing the TableRecommender altogether.

Either #930 or this PR should be used, and the other closed.

PR #928, opened 2020-09-29
CC-3610: Add Oracle and SQL Server JDBC drivers, which are now distribution friendly

rhauch:
## Problem
We were previously unable to distribute the MS SQL Server JDBC driver or the Oracle JDBC driver with the JDBC source or sink connector. This means that users had to install these before using the connector.

However, the licenses on these two JDBC drivers are now compatible with our distribution mechanism.

## Solution
Include these two JDBC drivers in the JDBC source and sink connector plugin archive.

* [MS SQL Server JDBC Driver license](https://github.com/microsoft/mssql-jdbc/blob/dev/LICENSE) as denoted in the [POM file in Maven Central for version 8.4.1](https://repo1.maven.org/maven2/com/microsoft/sqlserver/mssql-jdbc/8.4.1.jre8/mssql-jdbc-8.4.1.jre8.pom)
* [Oracle Free Use license](https://www.oracle.com/downloads/licenses/oracle-free-license.html), same as denoted in the [Oracle 8 Production BOM for version 19.7.0.0](https://www.oracle.com/database/technologies/maven-central-guide.html)

##### Does this solution apply anywhere else?
- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan
These can be included as early as 10.0.0.

PR #929, opened 2020-09-29
CC-11861: Changed to use 10.0.0-SNAPSHOT for the next release on `master`

rhauch:
## Problem
We need the release maven plugin for independent releases. Also, some adjustment in version variables

## Solution
Modify the POM and license files.

This is based upon #928, which updates the license files.

- [x] yes
- [ ] no

##### If yes, where?
The other connectors that have been removed from CP 6.0.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests


rhauch:
Superseded by #931

PR #933, opened 2020-10-02
CC-5879: 

SajanaW:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -22,5 +22,5 @@
               files="(DataConverter|GenericDatabaseDialect|JdbcSourceTask).java"/>
 
     <suppress checks="ParameterNumber"
-              files="(ColumnDefinition|GenericDatabaseDialect|SqlServerDatabaseDialect|PostgreSqlDatabaseDialect).java"/>
+              files="(ColumnDefinition|GenericDatabaseDialect|SqlServerDatabaseDialect|PostgreSqlDatabaseDialect|TimestampIncrementingTableQuerier).java"/>
 </suppressions>
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -7,13 +7,13 @@
 <suppressions>
 
     <suppress checks="CyclomaticComplexity"
-              files="(BufferedRecords|DataConverter|DatabaseDialect|FieldsMetadata|HanaDialect|JdbcSourceTask|MySqlDatabaseDialect|OracleDatabaseDialect|PostgreSqlDatabaseDialect|PreparedStatementBinder|SqlServerDatabaseDialect|SqliteDatabaseDialect|TimestampIncrementingTableQuerier|VerticaDatabaseDialect|SapHanaDatabaseDialect|TableId|ColumnDefinition|TableMonitorThread).java"/>
+              files="(BufferedRecords|DataConverter|DatabaseDialect|FieldsMetadata|HanaDialect|JdbcSourceTask|MySqlDatabaseDialect|OracleDatabaseDialect|PostgreSqlDatabaseDialect|PreparedStatementBinder|SqlServerDatabaseDialect|SqliteDatabaseDialect|TimestampIncrementingTableQuerier|VerticaDatabaseDialect|SapHanaDatabaseDialect|TableId|ColumnDefinition|TableMonitorThread|JdbcSourceConnector).java"/>
 
     <suppress checks="ClassDataAbstractionCoupling"
               files="(DbDialect|JdbcSourceTask|GenericDatabaseDialect).java"/>
 
     <suppress checks="NPathComplexity"
-              files="(BufferedRecords|DataConverter|FieldsMetadata|JdbcSourceTask|GenericDatabaseDialect).java"/>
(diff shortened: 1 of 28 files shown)


gharris1727 (changes requested):
Consider the following cases to check the upgrade path here:
1. There's no override in a subclass
2. There's an override in a subclass
3. There's an override in a subclass, and it calls it's super method.
4. After naively migrating the override
5. After completely migrating the override
6. After naively migrating the override with a super method call
7. After completely migrating the override with a super method call

Note:
* C is the connection, and X is whatever other arguments the method normally took.
* Naively migrating is delegating from Subclass(C,X) to Subclass(X) by discarding the connection
* Completely migrating is moving the logic from Subclass(X) to Subclass(C, X) and making it connection-aware.

Before:
1. JDBC calls Generic(X), executes Generic(X)
2. JDBC calls Generic(X), executes Subclass(X)
3. JDBC calls Generic(X), executes Subclass(X), calls Super(X), executes Generic(X)

## Copying Code (current approach)

After copying code from Generic(X) to Generic(C,X):
1. JDBC calls Generic(C,X), executes Generic(C,X)
2. JDBC calls Generic(C,X), **executes Subclass(C,X) that doesn't exist**
3. JDBC calls Generic(C,X), **executes Subclass(C,X) that doesn't exist**
4. JDBC calls Generic(C,X), executes Subclass(C,X), executes Subclass(X)
5. JDBC calls Generic(C,X), executes Subclass(C,X)
6. JDBC calls Generic(C,X), executes Subclass(C,X), executes Subclass(X), calls Super(X), executes Generic(X)
7. JDBC calls Generic(C,X), executes Subclass(C,X), calls Super(C,X), executes Generic(C,X) 

**This would be a regression**, since the subclass overridden behavior would revert to the generic behavior. Ideally we want: Generic(C,X) to execute Subclass(X) if it exists, but if Subclass(C,X) has happened, it should behave normally.

Additionally, the strategy of delegating Generic(C,X) to Generic(X) by discarding the connection has the same overall behavior, except that cases 1, 5, and 7 have an additional delegating step in them. This also doesn't get us a single shared connection, as the connection is discarded and the old logic is used,.

## Sentinel Values from Generic(X)
We could try to distinguish if Subclass(X) existed by returning `null` or throwing an exception from Generic(X). If we call Generic(X) and get null, we know that it wasn't overridden, and we can use the connection-sensitive code. Otherwise if it's non-null, use those results since they came from the subclass.
1. JDBC calls Generic(C,X), calls Generic(X), executes Generic(X), gets null, executes Generic(C,X)
2. JDBC calls Generic(C,X), calls Generic(X), executes Subclass(X), gets non-null
3. JDBC calls Generic(C,X), calls Generic(X), executes Subclass(X), calls Super(X), **executes Generic(X), gets `null` when it didn't expect to**
4. JDBC calls Generic(C,X), executes Subclass(C,X), executes Subclass(X)
5. JDBC calls Generic(C,X), executes Subclass(C,X)
6. JDBC calls Generic(C,X), executes Subclass(C,X), executes Subclass(X), calls Super(X), **executes Generic(X), gets `null` when it didn't expect to**
7. JDBC calls Generic(C,X), executes Subclass(C,X), calls Super(C,X), executes Generic(C,X)

**This would also be a regression**, as subclasses that relied on their superclass for their implementation would fail/change behavior. To be backwards compatible, both methods (Generic(X) and Subclass(X)) need to be valid, but distinguishable. I think this is only really possible with reflection.

## Two-Step
If we still used Generic(X) as the entry point, we could keep the old behavior temporarily, and later roll out the change once dialects have migrated.
Step 1: Add new Generic(C,X) method, and use a suitable connection to delegate from Generic(X) to Generic(C,X):
1. JDBC calls Generic(X), executes Generic(X), calls Generic(C,X), executes Generic(C,X)
2. JDBC calls Generic(X), executes Subclass(X)
3. JDBC calls Generic(X), executes Subclass(X), calls Super(X), executes Generic(X), calls Generic(C,X), executes Generic(C,X)
4. Same as 2
5. Same as 2
6. Same as 3
7.  Same as 3

This would effectively make this change a preparatory deprecation in advance of rolling out the database connection improvement, assuming that the "use a suitable connection" step above will prevent us from sharing a connection.

Step 2: change the JDBC connector from calling Generic(X) to Generic(C,X), and remove the deprecated method:
1. JDBC calls Generic(C,X), executes Generic(C,X)
2. JDBC calls Generic(C,X), **executes Subclass(C,X) doesn't exist** (but after marking other method as deprecated for a whole release)
3. JDBC calls Generic(C,X), **executes Subclass(C,X) doesn't exist** (but after marking other method as deprecated for a whole release)
4. JDBC calls Generic(C,X), executes Subclass(C,X), executes Subclass(X)
5. JDBC calls Generic(C,X), executes Subclass(C,X)
6. JDBC calls Generic(C,X), executes Subclass(C,X), executes Subclass(X), **calls Super(X) compilation error**
7. JDBC calls Generic(C,X), executes Subclass(C,X), calls Super(C,X), executes Generic(C,X)

**This would be incompatible**, but removing the deprecated methods would be incompatible anyway, and can accompany the roll-out of this feature.

TL;DR:
If we want to be backwards compatible and actually provide a deprecation path, we either need a 2-step rollout, or reflection. The current strategy is backwards-incompatible, but keeps around all of the stale code anyway.

PR #936, opened 2020-10-07
MINOR: Include maven release plugin to the right profile

kkonstantine:
## Problem
The plugin is included only in the standalone profile. 

## Solution
Needs to be included overall

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -392,6 +392,16 @@
                     </execution>
                 </executions>
             </plugin>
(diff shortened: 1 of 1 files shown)


rhauch (approved):
LGTM

PR #943, opened 2020-10-15
Implement Insert with Ignore statement on database dialects

blcksrx:
In some cases such as the existence of duplicating records, the sink connector task is going to fail if the `insert.mode=insert` in such a case. Also, the `upsert` mode is not efficient and fast enough in such cases with a large amount of data and data-append only(Such as logs or any event-based records). So here is the deal: Just insert the data into the database without any interruption on connectors task and  other effects that interruption on tasks might happen

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


blcksrx:
> [clabot:check]

Issue #374, opened 2018-03-01
Task is still running after connector is paused


michaelhuang:
Hi,  

When the task gets no new records from mysql, it can not be paused.

> WorkerSourceTask doesn't know the connector is paused, it will know this only after leaving of poll() method. But by design pool() method should block if no data is currently available.

https://issues.jboss.org/browse/DBZ-516

Connector config:
```
topic.prefix=test_jdbc_
name=mysql-local-source
incrementing.column.name=id
table.whitelist=users
timestamp.column.name=modified
tasks.max=1
connection.url=jdbc:mysql://localhost:3306/oltp?user=hzf&password=hzf
connector.class=io.confluent.connect.jdbc.JdbcSourceConnector
mode=timestamp+incrementing
#task ids: 0
```

Connector status:
```
connectorState:  PAUSED
workerId: 127.0.0.1:8083
numberOfTasks: 1
tasks:
  - taskId: 0
    taskState: RUNNING
    workerId: 127.0.0.1:8083
```

Traceback:
```
[2018-03-01 16:31:21,439] INFO Connector mysql-local-source target state change (org.apache.kafka.connect.runtime.distributed.DistributedHerder:1107)
[2018-03-01 16:31:21,440] INFO Setting connector mysql-local-source state to PAUSED (org.apache.kafka.connect.runtime.Worker:566)
[2018-03-01 16:31:21,440] INFO Stopping table monitoring thread (io.confluent.connect.jdbc.JdbcSourceConnector:166)
[2018-03-01 16:56:24,689] INFO 127.0.0.1 - - [01/Mar/2018:08:56:24 +0000] "GET /connectors/mysql-local-source/status HTTP/1.1" 200 171  3 (org.apache.kafka.connect.runtime.rest.RestServer:60)
[2018-03-01 16:57:00,709] INFO 127.0.0.1 - - [01/Mar/2018:08:57:00 +0000] "GET /connectors/mysql-local-source/status HTTP/1.1" 200 170  2 (org.apache.kafka.connect.runtime.rest.RestServer:60)
[2018-03-01 16:57:08,153] INFO WorkerSourceTask{id=mysql-local-source-0} Committing offsets (org.apache.kafka.connect.runtime.WorkerSourceTask:306)
[...]

PR #947, opened 2020-10-22
CC-12468: Periodically return control flow to WorkerSourceTask

ncliang:
## Problem
As described in https://github.com/confluentinc/kafka-connect-jdbc/issues/374, CC-12468, and SourceTask [javadoc](https://github.com/apache/kafka/blob/trunk/connect/api/src/main/java/org/apache/kafka/connect/source/SourceTask.java#L48), `poll()` should block when no data is available, but periodically return control flow to caller to give it a chance to pause the connector. Failure to do so caused pause to not be reflected until any result was polled from the database. The workaround is to delete the connector and recreate to resume. But, it is not as convenient as pause/resume.

## Solution
Keep track of consecutive empty results. Return `null` to WorkerSourceTask once consecutive empty results pass a threshold.

- [x] yes
- [ ] no

##### If yes, where?
Other source connectors will have the same issue and may need to be updated accordingly.

## Test Strategy
Added integration test for source connector pause/resume.

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Backport the bugfix as far back as `5.0.x`

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -303,7 +303,7 @@ public class JdbcSourceTask extends SourceTask {
   public List<SourceRecord> poll() throws InterruptedException {
     log.trace("{} Polling for new data");
 
-    Map<String, Integer> consecutiveEmptyResults = tableQueue.stream().collect(
+    Map<TableQuerier, Integer> consecutiveEmptyResults = tableQueue.stream().collect(
         Collectors.toMap(Function.identity(), (q) -> 0));
     while (running.get()) {
       final TableQuerier querier = tableQueue.peek();
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -53,6 +53,8 @@ import io.confluent.connect.jdbc.util.Version;
  * generates Kafka Connect records.
  */
 public class JdbcSourceTask extends SourceTask {
+  // When no results, periodically return control flow to caller to give it a chance to pause us.
+  private static final int CONSECUTIVE_EMPTY_RESULTS_BEFORE_RETURN = 3;
 
   private static final Logger log = LoggerFactory.getLogger(JdbcSourceTask.class);
 
@@ -300,6 +302,7 @@ public class JdbcSourceTask extends SourceTask {
   public List<SourceRecord> poll() throws InterruptedException {
     log.trace("{} Polling for new data");
 
+    int consecutiveEmptyResults = 0;
     while (running.get()) {
       final TableQuerier querier = tableQueue.peek();
 
@@ -309,6 +312,7 @@ public class JdbcSourceTask extends SourceTask {
             + config.getInt(JdbcSourceTaskConfig.POLL_INTERVAL_MS_CONFIG);
         final long now = time.milliseconds();
         final long sleepMs = Math.min(nextUpdate - now, 100);
+
         if (sleepMs > 0) {
           log.trace("Waiting {} ms to poll {} next", nextUpdate - now, querier.toString());
           time.sleep(sleepMs);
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -334,8 +338,18 @@ protected void closeResources() {
         }
 
         if (results.isEmpty()) {
+          consecutiveEmptyResults++;
           log.trace("No updates for {}", querier.toString());
-          continue;
+
+          if (consecutiveEmptyResults >= CONSECUTIVE_EMPTY_RESULTS_BEFORE_RETURN) {
```
gharris1727, line 344: Once it passes this check, it will start to tight loop with the framework. Perhaps we can set the context.timeout() to back-off slightly in this case.

gharris1727: Also, rather than waiting for 3 empty results, we could wait for one pass through all of the table queries, so that tables with no updates don't prevent tables with updates from being serviced.

ncliang: I don't think we get a `timeout` in the source context to play with. The intent of the check is to try a bit to prevent going into the tight loop in framework. The idea to wait for a pass through all table queries seems good. I think I'll go a step further and require that all table queries have returned empty for at least the number of consecutive queries before returning control to framework.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -300,6 +303,8 @@ protected void closeResources() {
   public List<SourceRecord> poll() throws InterruptedException {
     log.trace("{} Polling for new data");
 
+    Map<String, Integer> consecutiveEmptyResults = tableQueue.stream().collect(
+        Collectors.toMap(TableQuerier::toString, (q) -> 0));
```
gharris1727, line 307:
```suggestion
        Collectors.toMap(Function.identity(), (q) -> 0));
```

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -334,8 +340,19 @@ protected void closeResources() {
         }
 
         if (results.isEmpty()) {
+          consecutiveEmptyResults.compute(querier.toString(), (k, v) -> v + 1);
```
gharris1727, line 343:
```suggestion
          consecutiveEmptyResults.compute(querier, (k, v) -> v + 1);
```

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -334,8 +340,19 @@ protected void closeResources() {
...
+          if (Collections.min(consecutiveEmptyResults.values())
+              >= CONSECUTIVE_EMPTY_RESULTS_BEFORE_RETURN) {
+            log.trace("More than " + CONSECUTIVE_EMPTY_RESULTS_BEFORE_RETURN
+                + " consecutive empty results for all queriers, returning");
+            return null;
+          } else {
+            continue;
+          }
+        } else {
+          consecutiveEmptyResults.put(querier.toString(), 0);
```
gharris1727, line 355:
```suggestion
          consecutiveEmptyResults.put(querier, 0);
```

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -300,6 +303,8 @@ protected void closeResources() {
   public List<SourceRecord> poll() throws InterruptedException {
     log.trace("{} Polling for new data");
 
+    Map<String, Integer> consecutiveEmptyResults = tableQueue.stream().collect(
```
gharris1727, line 306:
```suggestion
    Map<TableQuerier, Integer> consecutiveEmptyResults = tableQueue.stream().collect(
```

gharris1727 (commented):
`TableQuerier::toString()` is not final, and may be changed in the future to include some dynamic state that might break this fix. I think we could just use the querier itself as the key, as the default/future implementations of equals/hashCode will keep us safe.

ncliang:
Thanks for review @gharris1727 . Yeah, I did not see `hashCode` implemented on `TableQuerier` so I went for `toString` to get a hash that is value based. It should not matter here though since we are just using the objects from `tableQueue` as the key.

gharris1727 (approved):
LGTM!

PR #959, opened 2020-11-10
CC-12626: validate topic.prefix to disallow spaces in between

liukrimhrim:
## Problem
The source connector did not validate topic.prefix. If there is a space in it, the connector will eventually end up with an exception, with a message totally irrelevant to the the very root cause.

## Solution
Validate the config in the first place to disallow spaces in between, and automatically trim leading and tailing spaces.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
targeting 5.0.x, will push to master.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -37,11 +37,13 @@ import io.confluent.connect.jdbc.util.QuoteMethod;
 import io.confluent.connect.jdbc.util.TableId;
 import io.confluent.connect.jdbc.util.TimeZoneValidator;
 
+import java.util.regex.Pattern;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Recommender;
 import org.apache.kafka.common.config.ConfigDef.Type;
+import org.apache.kafka.common.config.ConfigDef.Validator;
 import org.apache.kafka.common.config.ConfigDef.Width;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.utils.Time;
@@ -52,6 +54,7 @@ import org.slf4j.LoggerFactory;
 public class JdbcSourceConnectorConfig extends AbstractConfig {
 
   private static final Logger LOG = LoggerFactory.getLogger(JdbcSourceConnectorConfig.class);
+  private static Pattern INVALID_CHARS = Pattern.compile("[^a-zA-Z0-9._-]");
 
   public static final String CONNECTION_URL_CONFIG = "connection.url";
   private static final String CONNECTION_URL_DOC =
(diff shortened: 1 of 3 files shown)


C0urante:
Does this issue arise only with spaces or with any prefix that would eventually lead to an invalid topic name? Based on https://github.com/apache/kafka/blob/960e64072592bc1ad939518b58dbdfb17068680d/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L36-L47 it looks like there are a few more cases we might want to cover here, such as catching prefixes that would lead to names with lengths greater than the maximum allowed, and ones with any characters that are not alphanumeric or `-`, `_`, or `.`.

liukrimhrim:
> Does this issue arise only with spaces or with any prefix that would eventually lead to an invalid topic name? Based on https://github.com/apache/kafka/blob/960e64072592bc1ad939518b58dbdfb17068680d/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L36-L47 it looks like there are a few more cases we might want to cover here, such as catching prefixes that would lead to names with lengths greater than the maximum allowed, and ones with any characters that are not alphanumeric or `-`, `_`, or `.`.

Thanks for the review. Added checks for those cases.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -565,6 +568,33 @@ private static final void addConnectorOptions(ConfigDef config) {
...
+                    "Topic prefix must not start with '_', '-' or '.'.");
+              }
+            }
+
+            if (((String) value).length() > 249) {
+              throw new ConfigException(name, value,
+                  "Topic prefix length must not exceed max topic name length, 249 chars");
+            }
+
+            if (((String) value).trim().contains(" ")) {
```
C0urante, line 592: Shouldn't we check for all of the other possible invalid characters here as well?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -565,6 +568,33 @@ private static final void addConnectorOptions(ConfigDef config) {
...
+
+            if (!((String) value).isEmpty()) {
+              char first = ((String) value).charAt(0);
+              if (Stream.of('_', '-', '.').collect(Collectors.toSet()).contains(first)) {
+                throw new ConfigException(name, value,
+                    "Topic prefix must not start with '_', '-' or '.'.");
+              }
+            }
+
+            if (((String) value).length() > 249) {
```
C0urante, line 587: If we're invoking `trim()` below, probably best to do it here as well.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -565,6 +568,33 @@ private static final void addConnectorOptions(ConfigDef config) {
...
         Type.STRING,
+        "",
+        new Validator() {
+          @Override
+          public void ensureValid(final String name, final Object value) {
+            if (value == null) {
+              throw new ConfigException(name, value, "Topic prefix must not be null.");
+            }
+
+            if (!((String) value).isEmpty()) {
```
C0urante, line 579: If we're invoking `trim()` below, probably best to do it here as well.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -565,6 +571,36 @@ private static final void addConnectorOptions(ConfigDef config) {
...
+              throw new ConfigException(name, value, "Topic prefix must not be null.");
+            }
+
+            String trimmed = ((String) value).trim();
+
+            if (!trimmed.isEmpty()) {
+              char first = trimmed.charAt(0);
+              if (Stream.of('_', '-', '.').collect(Collectors.toSet()).contains(first)) {
+                throw new ConfigException(name, value,
+                    "Topic prefix must not start with '_', '-' or '.'.");
```
C0urante, line 588: Source for this?

liukrimhrim: > Source for this?

I don't know why I put the limitation, just removed it.

C0urante (approved):
Thanks for sticking with this Jinxin, LGTM

liukrimhrim:
> Thanks for sticking with this Jinxin, LGTM

Thanks for all the comments!

PR #984, opened 2020-12-23
MINOR: Fix JdbcSinkTask::version

C0urante:
Version is showing up as `null` in logs, but just for the sink task (and not the sink connector or the source connector or task). This should fix it by using the same strategy for the sink task's `version` method that's used by the other relevant classes.


ncliang (approved):
LGTM. Thanks @C0urante

C0urante:
Given that the fix is trivial and compiles locally, I'm going to ignore the build error and merge. Thanks @ncliang!

PR #997, opened 2021-01-27
Use fixed length type for primary keys in MySQL dialect (fixes #996)

### LINKED ISSUE #996: MySQL CREATE TABLE.. fails when specified key column is of type STRING
With the reasonable change from VARCHAR(256) to TEXT for MySQL DDL statements (see #363) the usage of a STRING field as primary key is not possible anymore - MySQL expects key to have a length and fails with

```
BLOB/TEXT column 'xxx' used in key specification without a key length
```
Keys should be created as VARCHAR(256) - this does not solve the problems of #551 but at least brings back the possibility of using a string key with `auto.create=true`

baunz:
See #996 for description


baunz:
[clabot:check]

adamnmcc (approved):

adamnmcc:
Hi, please can we get this reviewed and inlcuded? this will solve a lot of problems for mysql users

rarup1:
Is this going to be merged into master? Would be pretty useful!

PR #1001, opened 2021-02-04
MINOR: Log complete stack traces for all sub-exceptions when failing task

C0urante:


src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -82,12 +82,22 @@ public void put(Collection<SinkRecord> records) {
...
       SQLException sqlAllMessagesException = new SQLException(sqleAllMessages);
       sqlAllMessagesException.setNextException(sqle);
       if (remainingRetries == 0) {
+        log.error(
+            "Failing task after exhausting retries; "
+              + "encountered {} exceptions on last write attempt",
+            totalExceptions);
+        int exceptionCount = 1;
+        for (Throwable e : sqle) {
+          log.error("Exception {}:", exceptionCount++, e);
```
gharris1727, line 99: I understand that this is an error condition, but this log volume might be extreme for most cases. I think it's typical to run at INFO level, and drop down to debug for more info; so this might be a better fit for DEBUG severity.
```suggestion
          log.debug("Exception {}:", exceptionCount++, e);
```

C0urante: Fair, I've downgraded to `DEBUG` but added a note to the `ERROR`-level message that prompts users to enable `DEBUG` logging for more detail.

C0urante:
@gharris1727 Gonna merge now as the only outstanding comment has been addressed; I'm happy to address any follow-up if you're not convinced by the current approach, just let me know.

PR #1003, opened 2021-02-08
MINOR: Fix logger classes for dialects

C0urante:
## Problem
Inaccurate log namespaces caused by `GenericDatabaseDialect` subclasses inheriting the logger from their parent. Results in logs with weird things like `PostgreSqlDatabaseDialect:571` when the Postgres dialect has, to date, only 423 lines total.

## Solution
Give each database dialect its own logger.

- [ ] yes
- [x] no

## Test Strategy
Jenkins build.

##### Testing done:
- [x] Unit tests
- [x] Integration tests

## Release Plan
Target 5.0.x, port forward through to master.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -120,7 +120,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     }
   }
 
-  protected final Logger log = LoggerFactory.getLogger(getClass());
+  private final Logger log = LoggerFactory.getLogger(GenericDatabaseDialect.class);
   protected final AbstractConfig config;
 
   /**
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -120,11 +120,13 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     }
   }
 
-  // This field should be private as it should only be used by this class and subclasses should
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -120,7 +120,12 @@ public DatabaseDialect create(AbstractConfig config) {
     }
   }
 
-  protected final Logger log = LoggerFactory.getLogger(getClass());
+  // This field should be private as it should only be used by this class and subclasses should
+  // instantiate their own loggers. However, it was originally declared as protected, and changing
+  // it to private now would break existing dialects that inherit from this class and rely on this
+  // field. So we leave it as protected now, but strongly recommend that all dialects use their own
+  // loggers (as all dialects in this code base do).
+  protected final Logger log = LoggerFactory.getLogger(GenericDatabaseDialect.class);
```
gharris1727, line 128: hmm we could include `@Deprecated` but then it would mark all of our uses as deprecated too...
I guess we could also re-assign it to a private variable and ignore that one deprecation warning, and then rename all of the other accessors here.

gharris1727 (approved):
this always bothered me
one comment, up to you, on whether we should enforce this deprecation.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
C0urante: SGTM. I doubled the instantiation logic for `glog` instead of assigning `log` to it since the idiom for loggers is that they are `private static final` and `log` is not declared `static`, but apart from that I went with this change. Mind taking one more quick look?

gharris1727 (approved):
LGTM, thanks!

PR #1004, opened 2021-02-09
CCDB-3681: pick binding method wisely according to column def

liukrimhrim:
## Problem
Currently for string value, jdbc sink always use setString() to bind, and for bytes value, always use setBytes() to bind. This is causing several escalations. Please see https://confluentinc.atlassian.net/browse/CCDB-3681 for more details.

## Solution
Now choose binding method wisely according to column def. For string value, if the column def is CLOB or NCLOB, use setCharacterStream() to bind, whereas if the column def is NVARCHAR2, use setNString() to bind, otherwise use setString(); For bytes value, if the column def is BLOB, use setBlob(), otherwise use setBytes().

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
targeting 5.0.x, will pint merge to master.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -390,7 +390,6 @@ public interface DatabaseDialect extends ConnectionProvider {
 
   /**
    * Create a component that can bind record values into the supplied prepared statement.
-   *
    * @param statement      the prepared statement
    * @param pkMode         the primary key mode; may not be null
    * @param schemaPair     the key and value schemas; may not be null
@@ -399,19 +398,44 @@ public interface DatabaseDialect extends ConnectionProvider {
    * @return the statement binder; may not be null
    * @see #bindField(PreparedStatement, int, Schema, Object)
    */
+  @Deprecated
   StatementBinder statementBinder(
+      PreparedStatement statement,
+      JdbcSinkConfig.PrimaryKeyMode pkMode,
+      SchemaPair schemaPair,
+      FieldsMetadata fieldsMetadata,
+      JdbcSinkConfig.InsertMode insertMode
+  );
+
+  /**
+   * Create a component that can bind record values into the supplied prepared statement. By
+   * default, the behavior is the same as the other overloaded method with the extra parameter
+   * tableDefinition. This overloading method is introduced to deprecate the other overloaded
+   * method eventually.
+   *
+   * @param statement      the prepared statement
+   * @param pkMode         the primary key mode; may not be null
+   * @param schemaPair     the key and value schemas; may not be null
+   * @param fieldsMetadata the field metadata; may not be null
+   * @param tableDefinition the table definition; may be null
+   * @param insertMode     the insert mode; may not be null
+   * @return the statement binder; may not be null
+   * @see #bindField(PreparedStatement, int, Schema, Object)
+   */
+  default StatementBinder statementBinder(
       PreparedStatement statement,
       JdbcSinkConfig.PrimaryKeyMode pkMode,
       SchemaPair schemaPair,
       FieldsMetadata fieldsMetadata,
       TableDefinition tableDefinition,
       JdbcSinkConfig.InsertMode insertMode
-  );
+  ) {
+    return statementBinder(statement, pkMode, schemaPair, fieldsMetadata, insertMode);
+  }
 
   /**
    * Method that binds a value with the given schema at the specified variable within a prepared
    * statement.
-   *
    * @param statement the prepared statement; may not be null
    * @param index     the 1-based index of the variable within the prepared statement
    * @param schema    the schema for the value; may be null only if the value is null
@@ -419,6 +443,7 @@ public interface DatabaseDialect extends ConnectionProvider {
    * @throws SQLException if there is a problem binding the value into the statement
    * @see #statementBinder
    */
+  @Deprecated
   void bindField(
       PreparedStatement statement,
       int index,
@@ -428,7 +453,9 @@ public interface DatabaseDialect extends ConnectionProvider {
 
   /**
    * Method that binds a value with the given schema at the specified variable within a prepared
-   * statement.
+   * statement. By default, the behavior is the same as the other overloaded method with the extra
+   * parameter colDef. This overloading method is introduced to deprecate the other overloaded
+   * method eventually.
    *
    * @param statement the prepared statement; may not be null
    * @param index     the 1-based index of the variable within the prepared statement
@@ -438,13 +465,15 @@ public interface DatabaseDialect extends ConnectionProvider {
    * @throws SQLException if there is a problem binding the value into the statement
    * @see #statementBinder
    */
-  void bindField(
+  default void bindField(
       PreparedStatement statement,
       int index,
       Schema schema,
       Object value,
       ColumnDefinition colDef
-  ) throws SQLException;
+  ) throws SQLException {
+    bindField(statement, index, schema, value);
+  }
 
   /**
    * A function to bind the values from a sink record into a prepared statement.
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -423,6 +423,25 @@ public interface DatabaseDialect extends ConnectionProvider {
       PreparedStatement statement,
       int index,
       Schema schema,
+      Object value
+  ) throws SQLException;
+
+  /**
+   * Method that binds a value with the given schema at the specified variable within a prepared
+   * statement.
+   *
+   * @param statement the prepared statement; may not be null
+   * @param index     the 1-based index of the variable within the prepared statement
+   * @param schema    the schema for the value; may be null only if the value is null
+   * @param value     the value to be bound to the variable; may be null
+   * @param colDef    the Definition of the column to be bound; may be null
+   * @throws SQLException if there is a problem binding the value into the statement
+   * @see #statementBinder
+   */
+  void bindField(
+      PreparedStatement statement,
+      int index,
+      Schema schema,
       Object value,
       ColumnDefinition colDef
   ) throws SQLException;
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -404,6 +404,7 @@ public interface DatabaseDialect extends ConnectionProvider {
       JdbcSinkConfig.PrimaryKeyMode pkMode,
       SchemaPair schemaPair,
       FieldsMetadata fieldsMetadata,
+      TableDefinition tableDefinition,
       JdbcSinkConfig.InsertMode insertMode
   );
 
@@ -422,7 +423,8 @@ public interface DatabaseDialect extends ConnectionProvider {
       PreparedStatement statement,
       int index,
       Schema schema,
-      Object value
+      Object value,
+      ColumnDefinition colDef
   ) throws SQLException;
 
   /**
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -98,6 +98,7 @@ import io.confluent.connect.jdbc.util.TableId;
  * how a column value is converted to a field value for use in a {@link Struct}. To also change the
  * field's type or schema, also override the {@link #addFieldToSchema} method.
  */
+@SuppressWarnings("checkstyle:ClassDataAbstractionCoupling")
 public class GenericDatabaseDialect implements DatabaseDialect {
 
   protected static final int NUMERIC_TYPE_SCALE_LOW = -84;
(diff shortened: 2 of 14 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -81,13 +87,38 @@ protected boolean maybeBindPrimitive(
...
+    if (colDef == null) {
+      return super.maybeBindPrimitive(statement, index, schema, value, colDef);
+    }
+
     if (schema.type() == Type.STRING) {
-      statement.setNString(index, (String) value);
+      if (colDef.type() == Types.CLOB || colDef.type() == Types.NCLOB) {
+        statement.setCharacterStream(index, new StringReader((String) value));
+        return true;
+      } else if (colDef.type() == Types.NVARCHAR) {
```
mageshn, line 101: How about `NCHAR` ?

liukrimhrim: > How about `NCHAR` ?

good point! I added nchar.

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -76,18 +82,44 @@ protected String checkConnectionQuery() {
...
+        return true;
+      } else if (colDef.type() == Types.NVARCHAR || colDef.type() == Types.NCHAR) {
+        statement.setNString(index, (String) value);
+        return true;
+      } else {
+        return super.maybeBindPrimitive(statement, index, schema, value, colDef);
+      }
+    }
+
+    if (schema.type() == Type.BYTES && colDef.type() == Types.BLOB) {
```
gharris1727, line 110: Is there a reason to make BYTES data insertion into a non-BLOB column an error in the connector instead of the driver? surely the driver should know better than us what type conversions are possible.

gharris1727 (changes requested):

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -98,6 +98,7 @@
  * how a column value is converted to a field value for use in a {@link Struct}. To also change the
  * field's type or schema, also override the {@link #addFieldToSchema} method.
  */
+@SuppressWarnings("checkstyle:ClassDataAbstractionCoupling")
```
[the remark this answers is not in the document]
liukrimhrim: already added

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -404,6 +404,7 @@ StatementBinder statementBinder(
       JdbcSinkConfig.PrimaryKeyMode pkMode,
       SchemaPair schemaPair,
       FieldsMetadata fieldsMetadata,
+      TableDefinition tableDefinition,
```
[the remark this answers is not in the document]
Cyril-Engels: +1

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -76,18 +82,44 @@ protected String checkConnectionQuery() {
...
+      Object value,
+      ColumnDefinition colDef
   ) throws SQLException {
+    if (colDef == null) {
+      return super.maybeBindPrimitive(statement, index, schema, value, colDef);
+    }
+
     if (schema.type() == Type.STRING) {
-      statement.setNString(index, (String) value);
+      if (colDef.type() == Types.CLOB || colDef.type() == Types.NCLOB) {
```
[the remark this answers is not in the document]
liukrimhrim: > an NCLOB is considered a Character String and not an N String?

Good point. Didn't notice there's a setNCharacterStream() as well

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -76,18 +82,44 @@ protected String checkConnectionQuery() {
...
   protected boolean maybeBindPrimitive(
       PreparedStatement statement,
       int index,
       Schema schema,
-      Object value
+      Object value,
+      ColumnDefinition colDef
   ) throws SQLException {
+    if (colDef == null) {
+      return super.maybeBindPrimitive(statement, index, schema, value, colDef);
```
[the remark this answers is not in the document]
Cyril-Engels: Looks like this is just the old behavior to cover the KAFKA case https://github.com/confluentinc/kafka-connect-jdbc/pull/1004/files#diff-b8e818973a89932b3776d9d80dcd9ce6f464b7c781423ffbc4b381325375f081R101

After reverting the signature change in Dialect interface this should be good.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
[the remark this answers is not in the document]
aakashnshah: If you add them there, you still get the error without this additional flag?

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -76,18 +82,44 @@ protected String checkConnectionQuery() {
...
     if (schema.type() == Type.STRING) {
-      statement.setNString(index, (String) value);
+      if (colDef.type() == Types.CLOB || colDef.type() == Types.NCLOB) {
+        statement.setCharacterStream(index, new StringReader((String) value));
+        return true;
+      } else if (colDef.type() == Types.NVARCHAR || colDef.type() == Types.NCHAR) {
+        statement.setNString(index, (String) value);
+        return true;
+      } else {
+        return super.maybeBindPrimitive(statement, index, schema, value, colDef);
```
aakashnshah, line 106: Do you need this line/case if you do at the end of the overarching conditional statement anyway?

src/test/java/io/confluent/connect/jdbc/dialect/BaseDialectTest.java
```diff
@@ -469,19 +480,25 @@ public void bindFieldNull() throws SQLException {
   @Test(expected = ConnectException.class)
   public void bindFieldStructUnsupported() throws SQLException {
     Schema structSchema = SchemaBuilder.struct().field("test", Schema.BOOLEAN_SCHEMA).build();
-    dialect.bindField(mock(PreparedStatement.class), 1, structSchema, new Struct(structSchema));
+    ColumnDefinition colDef = mock(ColumnDefinition.class);
```
aakashnshah, line 483: It seems like you mock this each time. Any chance you can pull it out to lessen duplication of code?

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
liukrimhrim: yes, because we want to skip executing the rest

liukrimhrim:
@gharris1727 @mageshn Originally I wanted to include an IT for the change, but there seems to be some compatibility issues causing me not able to run Embedded Connect cluster. Hence, let's keep things simple, and just push changes with this PR and push the IT with another PR. I have verified with unit tests that the intended binding methods are getting called. And I have manually verified that the binding methods works with oracle.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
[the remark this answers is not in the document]
gharris1727: Please address this and remove these @SuppressWarnings annotations.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -425,6 +426,26 @@ void bindField(
...
+   *
+   * @param statement the prepared statement; may not be null
+   * @param index     the 1-based index of the variable within the prepared statement
+   * @param schema    the schema for the value; may be null only if the value is null
+   * @param value     the value to be bound to the variable; may be null
+   * @param colDef    the Definition of the column to be bound; may be null
+   * @throws SQLException if there is a problem binding the value into the statement
+   * @see #statementBinder
+   */
+  void bindField(
```
gharris1727, line 441: This is a backwards incompatible change, and anyone not inheriting from GenericDatabaseDialect won't have an implementation of this method.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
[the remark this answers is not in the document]
gharris1727: This is still a backwards incompatible change, even if you've fixed bindField. Please address this function.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -103,6 +103,7 @@ public BufferedRecords(
           config.pkMode,
           schemaPair,
           fieldsMetadata,
+          dbStructure.tableDefinition(connection, tableId),
```
gharris1727, line 106: This is not a safe change to make, because if someone implemented DatabaseDialect and implemented the old method, their implementation would never be called.

gharris1727 (changes requested):

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
liukrimhrim: > This is not a safe change to make, because if someone implemented DatabaseDialect and implemented the old method, their implementation would never be called.

I think now it is safe. dialects that didn't overwrite the overloaded statementBinder in the interface, shall be redirect to the old behavior.

liukrimhrim:
@gharris1727 I updated the PR based on our discussion on Friday. Now I think the changes should be backward compatible.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -115,7 +142,13 @@ protected int bindKeyFields(SinkRecord record, int index) throws SQLException {
       case RECORD_VALUE: {
         for (String fieldName : fieldsMetadata.keyFieldNames) {
           final Field field = schemaPair.valueSchema.field(fieldName);
-          bindField(index++, field.schema(), ((Struct) record.value()).get(field));
+          if (tabDef == null) {
+            bindField(index++, field.schema(), ((Struct) record.value()).get(field));
```
gharris1727, line 146: We should only call the new method, and let it delegate to the old method as a fallback.
When we don't have a table definition, we can pass-through the `null`.
Also, move this logic into the local `bindField` method and pass in the fieldName to deduplicate this logic.

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -77,14 +89,80 @@ protected String checkConnectionQuery() {
   }
 
   @Override
+  public StatementBinder statementBinder(
```
gharris1727, line 92: Hmm, this is unfortunate that we need this other method to enable the new functionality. I think this means that the rollout needs an additional phase:
1. this PR to add the deprecated functions
2. migration of all of the leaf dialects including adding this override
3. Adding the override to Generic, removing the old methods
4. Removing all of these unnecessary overrides from the leaf dialects

It seems unavoidable, since we can't provide a non-null tableDefinition by default without erasing a downstream dialect's functionality.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -46,13 +49,27 @@ public PreparedStatementBinder(
       SchemaPair schemaPair,
       FieldsMetadata fieldsMetadata,
       JdbcSinkConfig.InsertMode insertMode
+  ) {
+    this(dialect, statement, pkMode, schemaPair, fieldsMetadata,
+        null, insertMode);
```
gharris1727, line 54: nit:
```suggestion
    this(
        dialect, 
        statement, 
        pkMode, 
        schemaPair, 
        fieldsMetadata,
        null,
        insertMode
    );
```

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -93,20 +110,21 @@ protected int bindKeyFields(SinkRecord record, int index) throws SQLException {
 
       case KAFKA: {
         assert fieldsMetadata.keyFieldNames.size() == 3;
-        bindField(index++, Schema.STRING_SCHEMA, record.topic());
-        bindField(index++, Schema.INT32_SCHEMA, record.kafkaPartition());
-        bindField(index++, Schema.INT64_SCHEMA, record.kafkaOffset());
+        bindField(index++, Schema.STRING_SCHEMA, record.topic(), null, "");
+        bindField(index++, Schema.INT32_SCHEMA, record.kafkaPartition(), null, "");
+        bindField(index++, Schema.INT64_SCHEMA, record.kafkaOffset(), null, "");
```
gharris1727, line 115: Can you actually fill in the real parameters here? The names of these columns are in `JdbcSinkConfig.DEFAULT_KAFKA_PK_NAMES`, and we should use the real table definition if it's available.

liukrimhrim: > Can you actually fill in the real parameters here? The names of these columns are in `JdbcSinkConfig.DEFAULT_KAFKA_PK_NAMES`, and we should use the real table definition if it's available.

Since the table definition is null, the last parameter will just get ignored.

gharris1727 (approved):
LGTM with green build

PR #1039, opened 2021-04-13
CCDB-3880: Set default schema to dbo for MicrosoftSqlServer Dialect

SajanaW:

## Problem
Currently if table name does not include a schema, when the connector looks for the table in the table, it looks for that table in any schema. However when executing an Insert or Update SQL statement with just the table name, SqlServer assumes the schema is dbo. 

This leads to a bug where you might specify a table without a schema which is not in the schema dbo and the connector fails with `Invalid Object name` instead of `table does not exist`. 

See https://confluentinc.atlassian.net/browse/CCDB-3880 

## Solution
Set schema to dbo when table name does not include a schema. 

- [ ] yes
- [ ] no

##### Testing done:
- [x ] Unit tests
- [x ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Pint merge


src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -108,6 +108,25 @@ public SqlServerDatabaseDialect(AbstractConfig config) {
...
   }
 
+  @Override
+  public TableId parseTableIdentifier(String fqn) {
+    List<String> parts = identifierRules().parseQualifiedIdentifier(fqn);
+    if (parts.isEmpty()) {
+      throw new IllegalArgumentException("Invalid fully qualified name: '" + fqn + "'");
+    }
+    if (parts.size() == 1) {
+      return new TableId(null, "dbo", parts.get(0));
```
gharris1727, line 118: WDYT about calling `super.parseTableIdentifier(fqn)` and then if the schema is null, returning a replacement TableId?

gharris1727 (commented):
an IT here seems unecessary but I wont turn it down :)

gharris1727 (approved):
LGTM, thanks for the fix @SajanaW!

tushargoyal22:
if I specify `table.name.format` as just table_name which is `raft_tbl` in my case then it dumps the data in the table under the default schema `dbo`

whereas when I tried to specify `table.name.format` as `guest.raft_tbl` where `guest` is one of the schemas in my database, **com.microsoft.sqlserver.jdbc.SQLServerException: Database 'guest' does not exist. Make sure that the name is entered correctly.** , I am not sure **why it is looking for Database `guest`**


JDBC sink connector config for reference

```
{
    "name": "SinkConnector3",
    "config": {
        "connector.class": "io.confluent.connect.jdbc.JdbcSinkConnector",
        "tasks.max": 1,
        "topics": "raft.public.goyaltu.subscription.manager.test_client",

        "key.converter": "io.confluent.connect.avro.AvroConverter",
        "key.converter.schema.registry.url": "http://mwkafka-schema-registry-staging.deshaw.com:5036",
        "value.converter": "io.confluent.connect.avro.AvroConverter",
        "value.converter.schema.registry.url": "http://mwkafka-schema-registry-staging.deshaw.com:5036",

        "transforms": "AlignSchemaWithRegistry",
        "transforms.AlignSchemaWithRegistry.type": "deshaw.raft.connect.transforms.AlignSchemaWithRegistry",
        "transforms.AlignSchemaWithRegistry.schema.registry.urls": "http://mwkafka-schema-registry-staging.deshaw.com:5036",

        "connection.url": "jdbc:sqlserver://riskdb;databaseName=sandbox;integratedSecurity=true;authenticationScheme=JavaKerberos;encrypt=true;trustServerCertificate=true;",

        "dialect.name": "SqlServerDatabaseDialect",

        "insert.mode": "upsert",
        "table.name.format": "guest.raft_tbl",
        "pk.mode": "record_key",
        "pk.fields": "id",
        "batch.size": "2",
        "poll.interval.ms": 60000,

        "tasks.recovery.mechanism": "STATE_BASED",


        "consumer.override.client.id": "raft.goyaltu.sinkconnector.jdbc.client",
        "consumer.override.group.id": "raft.test.sinkconnector.consumer",

        "consumer.override.java.security.auth.login.config": "/prod/tools/middleware/kafkautils/dist/etc/config/kafka/zookeeper_client_jaas.conf",
        "consumer.override.security.protocol": "SASL_PLAINTEXT",
        "consumer.override.sasl.mechanism": "GSSAPI",
        "consumer.override.sasl.kerberos.service.name": "kafka",
        "consumer.override.sun.security.jgss.native": true,
        "consumer.override.sun.security.jgss.lib": "/usr/libexec/libgsswrap.so",
        "consumer.override.javax.security.auth.useSubjectCredsOnly": false
    }
}

```

PR #1040, opened 2021-04-17
CCDB-3740: Fix bug fitting numerics into double primitive type

ncliang:
## Problem
There was a bug when handling `best_fit` numeric mapping that prevented any values with precision >= 19 from being mapped to double

## Solution
Fix the bug by checking scale first to determine if we can fit in integer type. If not, fit to double. The case left that will still fall through to map to Decimal is if we decide to map to integer type based on scale <=0 , but we cannot fit into any primitive integer type. In this case, we prefer to not losing precision and revert to use Decimal still.

- [ ] yes
- [x] no

## Test Strategy
Unittest expectations were updated.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Since the behavior change is not backwards-compatible, it will be merged to master and released with next major version bump.


C0urante:
I see the benefit of this change but it seems to come at a pretty heavy cost. The existing checks for precision help prevent loss of precision (though they aren't entirely effective as IIUC you can't get 18 digits of precision with a 64-bit floating point decimal). Defaulting to 64-bit floats for any kind of arbitrary-precision decimal data that can't be fit into an integer type might cause loss of precision in even more cases, such as when the precision of the database column is at least 18.

I think this change is ultimately more intuitive and would probably choose to opt in to it if starting an application using the connector from scratch, but given the potential loss of precision not everyone may feel this way, so we might want to hold off on hardcoding it. Do you think another value for the `numeric.mapping` property might be an acceptable way to allow users to opt in to this new behavior? It'd also save us the trouble of having to perform a major version bump on the next release, and be easier on existing users whose pipelines would be negatively impacted by a sudden change in schema type.

ncliang:
Thanks for review @C0urante . I think what you describe makes sense. I added a value `best_fit_eager_double` to the `numeric.mapping` options to have doubles behave as expected and cast to FLOAT64 connect type despite potential of losing accuracy. PTAL?

C0urante (approved):
LGTM, thanks Nigel. The docs additions are super informative!

PR #1044, opened 2021-04-21
FF-2998: Postgres primitive array support.

jcustenborder:
## Problem

Postgres wants arrays.

## Solution

Give Postgres arrays.

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [X] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.jdbc.dialect;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 import org.apache.kafka.common.config.AbstractConfig;
@@ -46,6 +47,7 @@ import io.confluent.connect.jdbc.util.ExpressionBuilder.Transform;
 import io.confluent.connect.jdbc.util.IdentifierRules;
 import io.confluent.connect.jdbc.util.TableDefinition;
 import io.confluent.connect.jdbc.util.TableId;
+import org.apache.kafka.connect.errors.DataException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -372,7 +374,6 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
     }
   }
 
-  @SuppressWarnings("unchecked")
   @Override
   protected boolean maybeBindPrimitive(
       PreparedStatement statement,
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -366,6 +372,74 @@ protected void formatColumnValue(
...
+      Object value
+  ) throws SQLException {
+
+    switch (schema.type()) {
+      case ARRAY: {
+        Class<?> valueClass = value.getClass();
+        Object newValue = null;
+        if (valueClass.isArray()) {
+          newValue = value;
+        } else {
```
ncliang, line 390: Maybe check that valueClass is a Collection here, and have another else that throws if we try to bind a non-array, non-collection type to an array?

src/test/java/io/confluent/connect/jdbc/sink/integration/PostgresDatatypeIT.java
```diff
@@ -213,6 +216,118 @@ public void testWriteToTableWithUuidColumn() throws Exception {
...
+          assertJDBCArray(rs, "friendnames", struct);
+        }
+      }
+    }
+  }
+
+  @Test
+  public void testWriteToTableWithIntArrayColumnMissingFields() throws SQLException, InterruptedException {
+    createTableWithIntArrayColumnsMissing();
+    props.put(JdbcSinkConfig.AUTO_CREATE, "true");
```
ncliang, line 257: auto_create is not required for this testcase. testAutoCreateTableWithPrimitiveTypeArrayColumn should be its own testcase.

ncliang (commented):
Thanks @jcustenborder . I see that you are building on https://github.com/confluentinc/kafka-connect-jdbc/pull/805 and addressing comments on that PR. I have some more minor comments here. High level, how much effort do you think it would be to add this support in the generic dialect? Is it mostly testing to figure out which driver implementations do not support binding primitive arrays with `bindObject` and array type? We have some IT that are testing with MySql and MSSQL dialects, and tests using embedded Derby. If we just covered these dialects with some manual testing for others, I think that could give us enough confidence to move it to the generic dialect. This does not need to be part of this PR, but could be covered in a followup PR as well.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
jcustenborder: Unified array and list codepaths.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -366,6 +374,80 @@ protected void formatColumnValue(
...
+          case FLOAT32:
+            newValue = valueCollection.toArray(new Float[0]);
+            break;
+          case INT64:
+            newValue = valueCollection.toArray(new Long[0]);
+            break;
+          default:
+            break;
+        }
+
```
ncliang, line 437:
```suggestion
```

ncliang (approved):
Thanks @Faqa , @jcustenborder , the change LGTM. We can expand to support arrays in other dialects in another PR.

jcustenborder:
Thanks buddy!

aakashnshah:
@Iskuskov I'm not sure if we have timelines for this, usually we do these releases when a new version is warranted (typically when a feature is merged to master like this). However, I don't have enough context on this to know if it is a feature that needs to be immediately released. I tagged the people above that seem to be most knowledgable about this, hopefully they can give a better answer.

Iskuskov:
@jcustenborder, hi! thank you very much for the great feature once again! Is it correct this PR supports arrays only on sink side, not on source?

NathanNam:
This feature should be available in 10.2.0 or above.

Issue #563, opened 2019-01-07
numeric.mapping doesn't appear to work with Confluent Platform 5.1 and MySQL


rmoff:
I can't get `numeric.mapping` to work with MySQL and Confluent Platform 5.1. Steps to reproduce below. 

Create MySQL table: 

```
use demo;

create table transactions (
	txn_id INT,
	customer_id INT,
	amount DECIMAL(5,2),
	currency VARCHAR(50),
	txn_timestamp VARCHAR(50)
);

insert into transactions (txn_id, customer_id, amount, currency, txn_timestamp) values (3, 2, 17.13, 'EUR', '2018-04-30T21:30:39Z');

```

Inspect table: 

```
mysql> describe transactions;
+---------------+--------------+------+-----+---------+-------+
| Field         | Type         | Null | Key | Default | Extra |
+---------------+--------------+------+-----+---------+-------+
| txn_id        | int(11)      | YES  |     | NULL    |       |
| customer_id   | int(11)      | YES  |     | NULL    |       |
| amount        | decimal(5,2) | YES  |     | NULL    |       |
| currency      | varchar(50)  | YES  |     | NULL    |       |
| txn_timestamp | varchar(50)  | YES  |     | NULL    |       |
+---------------+--------------+------+-----+---------+-------+
5 rows in set (0.00 sec)
```

Create connector

```
curl -X POST http://localhost:8083/connectors -H "Content-Type: application/json" -d '{
          "name": "jdbc_source_mysql_12a",
          "config": {
                  "connector.class": "io.confluent.connect.jdbc.JdbcSourceConnector",
                  "connection.url": "jdbc:mysql://mysql:3306/demo",
                  "connection.user": "connect_user",
                  "connection.password": "asgard",
                  "topic.prefix": "mysql-12a-",
                  "numeric.mapping": "best_fit",
                  "table.whitelist" : "demo.transactions",
                  "mode":"bulk",
                  "poll.interval.ms" : 3600000
                  }
          }'
```          

Even though `"numeric.mapping": "best_fit"`, Kafka Connect stores the `DECIMAL(5,2)` as a `Decimal`, serialised to bytes in Avro: 

```
$ curl -s "http://localhost:8081/subjects/mysql-12a-transactions-value/versions/1"|jq '.schema|fromjson.fields[] | select (.name == "amount")'
{
  "name": "amount",
  "type": [
    "null",
    {
      "type": "bytes",
      "scale": 2,
[...]

PR #1047, opened 2021-04-27
WIP: feat(jdbc): Support specifying default precision and max scale for DE…

justinrlee:
…CIMAL data fields

- [x] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -19,7 +19,7 @@
               files="(DataConverter|FieldsMetadata|JdbcSourceTask|GenericDatabaseDialect).java"/>
 
     <suppress checks="MethodLength"
-              files="(DataConverter|GenericDatabaseDialect|JdbcSourceTask).java"/>
+              files="(DataConverter|GenericDatabaseDialect|JdbcSourceTask|JdbcSourceConnectorConfig).java"/>
 
     <suppress checks="ParameterNumber"
               files="(ColumnDefinition|GenericDatabaseDialect|SqlServerDatabaseDialect|PostgreSqlDatabaseDialect).java"/>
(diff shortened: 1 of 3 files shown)


justinrlee:
Still need to add tests, and also need to sign the CLA.  But I'm hoping somebody can tell me if this is the right way to address this, or if there's a better way.

justinrlee:
Maybe addresses the item identified here: https://stackoverflow.com/questions/58517842/kafka-connect-jdbc-oracle-number-fields-has-precision-of-64-which-fails-to-rea
and referenced here: https://github.com/confluentinc/kafka-connect-jdbc/issues/563

justinrlee:
[clabot:check]

ncliang:
Thanks for the PR @justinrlee . I am worried that this exposes implementation details to the user that can be hidden in the connector dialect implementation. For instance, one would have to know to configure the max precision and scale to 131072 and 16383 for a numeric that does not have this specified in postgres, based on reading the postgres documentation https://www.postgresql.org/docs/10/datatype-numeric.html#DATATYPE-NUMERIC-TABLE . The same would apply when using other dialects such as Oracle and MS SQL. While I agree with the approach to introduce the change in a backwards-compatible way by disabling it in flags by default, I think we can do a bit more leg work and have the individual database limits applied in their perspective dialects. These changes would go in the Oracle or Postgres dialect implementations with the known limits of the specific dialect, with a high level config option `use.dialect.default.numeric.precision` or similar config added to toggle this behavior on or off. WDYT?

justinrlee:
Closing in favor of https://github.com/confluentinc/kafka-connect-jdbc/pull/1052 (which has been merged) and https://github.com/jcustenborder/kafka-connect-transform-common/pull/75 (pending)

PR #1052, opened 2021-05-10
FF-5018: JDBC Source Connector needs to provide column precision for DECIMAL (NUMERIC) data types

justinrlee:
## Problem

We currently have precision available to us, but we don't add it to the Connect Schema.  This one-line change adds the precision.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1096,7 +1096,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
         glog.debug("DECIMAL with precision: '{}' and scale: '{}'", precision, scale);
         scale = decimalScale(columnDefn);
         SchemaBuilder fieldBuilder = Decimal.builder(scale);
-        // fieldBuilder.parameter("connect.decimal.precision", Integer.toString(precision));
+        fieldBuilder.parameter("connect.decimal.precision", Integer.toString(precision));
         if (optional) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1096,6 +1096,7 @@ protected String addFieldToSchema(
         glog.debug("DECIMAL with precision: '{}' and scale: '{}'", precision, scale);
         scale = decimalScale(columnDefn);
         SchemaBuilder fieldBuilder = Decimal.builder(scale);
+        fieldBuilder.parameter("connect.decimal.precision", Integer.toString(precision));
```
ncliang, line 1099: nit: move `connect.decimal.precision` to a private static final String.

ncliang (approved):
Thanks @justinrlee , LGTM except for a minor nit. I can merge this change once it is addressed.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
justinrlee: Okay, updated (using the variable name used in ksql: https://github.com/confluentinc/ksql/blob/d15026fe7c8993673152d740d46b16253896d8e8/ksql-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java#L33)

Thanks!

PR #1053, opened 2021-05-11
CCDB-3981: Add a null check for get after auto create for JDBC connector

SajanaW:

## Problem
There is a possibility for an NPE to happen if the DDL statements happen without an exception, but describe/get table returns null in the database.  Generally RDS guarantees DDLs are consistent with the get but we've encountered a database where this happens. 

 
## Solution
Add more null checks before and after `amendIfNecessary ` 

- [ ] yes
- [ ] no

##### Testing done:
- [x ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Pint merge up

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
@@ -111,10 +111,12 @@ public class DbStructure {
 
     final TableDefinition tableDefn = tableDefns.get(connection, tableId);
     if (tableDefn == null) {
-      throw new ConnectException(
-          "Unable to auto create table " + tableId
-              + " . DDL statement for auto creation was successful"
-                  + " but table does not exist in database."
+      throw new SQLException(
+          "Table with id " + tableId +  "does not exist in database."
+              + " Since we've either already auto-created the table or confirmed"
+              + " it existed before this point, it must mean this user does not"
+              + " have the permission to read the column data for this table"
+              + " (assuming the database is ACID compliant)."
       );
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -110,6 +110,15 @@ boolean amendIfNecessary(
     //   We also don't check if the data types for columns that do line-up are compatible.
 
     final TableDefinition tableDefn = tableDefns.get(connection, tableId);
+    if (tableDefn == null) {
+      throw new SQLException(
+          "Table with id " + tableId +  "does not exist in database."
+              + " Since we've either already auto-created the table or confirmed"
+              + " it existed before this point, it must mean this user does not"
+              + " have the permission to read the column data for this table"
+              + " (assuming the database is ACID compliant)."
```
gharris1727, line 119: this is unnecessary :)
```suggestion
              + " have the permission to read the column data for this table."
```

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -179,7 +188,17 @@ boolean amendIfNecessary(
...
 
-    tableDefns.refresh(connection, tableId);
+    if (tableDefns.refresh(connection, tableId) == null) {
+      // Refresh should not make the table null.
+      throw new SQLException(
+        "Table with id " + tableId +  "does not exist in database."
+                + " Since we've either already auto-created the table or confirmed"
+                + " it existed before this point, it must mean this user does not"
+                + " have the permission to read the column data for this table"
+                + " (assuming the database is ACID compliant)."
```
gharris1727, line 198:
```suggestion
                + " have the permission to read the column data for this table."
```

gharris1727 (approved):
LGTM

src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -179,7 +186,13 @@ boolean amendIfNecessary(
       );
     }
 
-    tableDefns.refresh(connection, tableId);
+    if (tableDefns.refresh(connection, tableId) == null) {
+      // Refresh should not make the table null.
+      throw new ConnectException(
+              "Unable to auto create or update table. Ddl statements to create or update"
+                  + " the table was successful but table does not exist. Table id" + tableId
```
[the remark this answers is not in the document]
SajanaW: I think we decided this could be because of a permission issue.

SajanaW (commented):
As discussed offline,  the error messages for the null checks here should reflect that the user might not have permission to describe the tables in consideration.  All tables this connector writes to should provide GET after CREATE. If we find that a GET fails after CREATE, this must mean that the user does not have permission to describe the tables in consideration.

PR #1059, opened 2021-05-13
CC-14541 : Exclude zk and netty from packaging

mageshn:
## Problem

Vulnerabilities with below versions as reported by twist lock

org.apache.zookeeper_zookeeper 3.4.13-- Please upgrade to the latest version listed in the following CVEs 
io.netty_netty 3.10.6.Final-- Please upgrade to the latest version listed in the following CVEs 

## Solution

Explicitly mark the ZK dependencies as `test`. common pom defines them in dependency management and hence they are marked as `compile` by maven and included in the final package. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -198,6 +198,21 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+
+        <!-- common pom defines them in dependency management and hence they're pulled in compile scope.
+        Explicitly define them as test to prevent from them form being packaged. -->
+        <dependency>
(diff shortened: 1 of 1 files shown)


ncliang (approved):
Thanks @mageshn . The approach LGTM. Should the fix go further back than `5.2.x` ?

mageshn:
The CVE's were reported only in these versions. I think we can backport if a need arises.

PR #726, opened 2019-10-17
'numeric.mapping' config to handle NUMERIC and DECIMAL fields

pushpavanthar:
This PR contains changes for numeric.mapping config to support both NUMERIC and DECIMAL fields. It resolves open issues [563](https://github.com/confluentinc/kafka-connect-jdbc/issues/563) and [311](https://github.com/confluentinc/kafka-connect-jdbc/issues/311).

PR #1065, opened 2021-05-18
'numeric.mapping' config to handle NUMERIC and DECIMAL fields and fix decimal schema precision issue

cdmikechen:
## Problem
Based on PR https://github.com/confluentinc/kafka-connect-jdbc/pull/726 and PR https://github.com/confluentinc/kafka-connect-jdbc/pull/803 .

## Solution
`decimal` and `numeric` type can use same config to transform date type from decimal to int/long/float/double. And also fix decimal type issue.

- [x] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Maybe 10.2.0


cdmikechen:
[clabot:check]

cdmikechen:
@ncliang Hi~
I submitted new codes based on the old PR [ https://github.com/confluentinc/kafka-connect-jdbc/pull/726]. it seems that the old PR committer has not been processed for some time.

ncliang:
Thanks for the PR @cdmikechen . I'm concerned about the implications for backwards compatibility with this change. While having DECIMAL behavior match NUMERIC behavior is arguably correct and the behavior that many users expect, we may have users depending on current DECIMAL behavior that would be broken with this change. Can we introduce this change behind a config property, `apply.numeric.mapping.to.decimal` or what not, with a default of `false`?

cdmikechen:
@ncliang 
Your suggestion is very good ! 
I'll change it later

PR #1070, opened 2021-05-25
Introduce option to enable/disable HOLDLOCK when using MERGE INTO in MS SQL Server

spinatelli:
## Problem

Using "with (HOLDLOCK)" when performing a MERGE INTO statement in MS SQL Server in certain cases brings up performance issues (https://sqlsunday.com/2021/05/04/how-merge-can-deadlock-you/).
The HOLDLOCK is currently hardcoded into the SQL statement of the Sink connector [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java#L327) and cannot be disabled without having a custom JDBC connector.

## Solution
This PR introduces a configuration option (`mssql.use.merge.holdlock`) that allows to enable/disable the use of HOLDLOCK in the MERGE statement. The default is `true`, meaning this PR should be backwards compatible.

NOTE: the unit tests had to be updated, as the JDBCSourceConfig was used everywhere, even when testing Sink features. 

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
No plan

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
@@ -325,7 +325,7 @@ public class SqlServerDatabaseDialect extends GenericDatabaseDialect {
     ExpressionBuilder builder = expressionBuilder();
     builder.append("merge into ");
     builder.append(table);
-    if (this.config.getBoolean(JdbcSinkConfig.MSSQL_USE_MERGE_HOLDLOCK)) {
+    if (((JdbcSinkConfig) this.config).useHoldlockInMerge) {
       builder.append(" with (HOLDLOCK)");
     }
     builder.append(" AS target using (select ");
--- a/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.jdbc.dialect;
 
+import io.confluent.connect.jdbc.sink.JdbcSinkConfig;
 import org.apache.kafka.common.config.AbstractConfig;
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -324,7 +325,10 @@ public String buildUpsertQueryStatement(
     ExpressionBuilder builder = expressionBuilder();
     builder.append("merge into ");
     builder.append(table);
-    builder.append(" with (HOLDLOCK) AS target using (select ");
+    if (this.config.getBoolean(JdbcSinkConfig.MSSQL_USE_MERGE_HOLDLOCK)) {
```
aakashnshah, line 328:
```suggestion
    if (this.config.useHoldlockInMerge) {
```

aakashnshah (commented):
Looks pretty good overall, I'm not sure if we want to a add configuration specifically for one dialect and if there are any alternatives.

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
spinatelli: Needs casting to JdbcSinkConfig

((JdbcSinkConfig) this.config).useHoldlockInMerge

spinatelli:
Hi Aakash, thanks for the suggestion and the code check.
As for the dialect-specific configuration, if there are any alternatives that are better suited I am happy to update the code with the needed implementation.
Would be great to have an indication about that :)

spinatelli:
Hi @aakashnshah, do you have any further input from your side about the dialect-specific configurations? :)

ManasjyotiSharma (approved):
LGTM; awesome work! Thanks @spinatelli for your efforts here.

PR #1082, opened 2021-06-09
Informix DB support added

vigneshtvp:
This will enable informix Dialect support on Kafka connect

## Problem
Current Kafka connect doesn't support Informix DB

## Solution
I have added support for informix DB dialect

- [] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


vigneshtvp:
In Informix DB, kafka-connect-jdbc qenerated query for view should have different delimiter(colon & dot) like (select * from database_name:Schema_name.Table_name), In other DB its always dotted delimiter. Along with Informix dialect, this issue also taken cared

PR #1091, opened 2021-07-08
Change pom repository reference value to https

kppullin:
Similar to https://github.com/confluentinc/common/pull/341, change the pom's repository value to `https` as now required by default with newer versions of maven (3.8.1+, ref: https://issues.apache.org/jira/browse/MNG-7118).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
         <repository>
             <id>confluent</id>
             <name>Confluent</name>
-            <url>http://packages.confluent.io/maven/</url>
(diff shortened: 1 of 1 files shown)


kppullin:
It's worth noting that this change doesn't completely fix compilation issues by itself, since the version of the parent pom referenced in this project also uses `http` and not `https`.  While this has been fixed in the parent pom, until the parent version is updated as well, those using newer versions of maven will have to hack around the `https` requirement.

NickUfer:
Any progress on this?

PR #1096, opened 2021-07-28
CCDB-4193: Terminate task on SQLNonTransientException

ncliang:
## Problem
Some `SQLException`s are non-transient and should result in failing the task rather than log and continue.

## Solution
Explicitly catch and rethrow `SQLNonTransientException` wrapped in a `ConnectException`

- [ ] yes
- [x] no

## Test Strategy
New unittest was added

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Bugfix to be taken back to `10.0.x`

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.jdbc.source;
 
+import java.sql.SQLNonTransientException;
 import java.util.TimeZone;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.utils.SystemTime;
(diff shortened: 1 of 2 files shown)


C0urante (approved):
LGTM, thanks Nigel.

I really wish there were a cleaner way to handle the exceptions here since we're essentially duplicating the logic for the `catch (SQLNonTransientException sqle)` and `catch (Throwable t)` branches, but I can't think of anything off the top of my head and it's not worth blocking. I suspect you may have thought about this too and arrived at the same conclusion.

Regardless, this is certainly good enough to merge as-is.

mukeshus:
Is there any specific setting that we need to do it in the jdbc connector config? The connector is not failing still even if there is a SQL exception. My connector version - 10.5.1

PR #1107, opened 2021-08-25
CCDB-4247: Close resources correctly on stop

ncliang:
## Problem
Since KAFKA-10792 fixed the source task `stop()` to be called on the task thread, the previous assumption that `stop()` will be called in a separate thread is no longer correct. Instead, the shutdown steps should be done in the `stop()` method directly. This method will be called by framework on the same task thread as `poll()` , once `poll()` relinquishes control back to framework.

## Solution
Detect whether `stop()` is called from task thread. If so, do shutdown immediately.

- [ ] yes
- [x] no

## Test Strategy
Unit test expectation was updated

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Since KAFKA-10792 was taken back to `2.5` , we similarly take this change back to `5.5.x` .


gharris1727 (approved):
amazing

PR #1110, opened 2021-09-03
Add topic.suffix config prop for JDBC Source

desiderantes:
## Problem
Currently there's a `topic.prefix ` config prop, but there's not corresponding suffix option, and useful addition for things like topic versioning and general naming conventions

## Solution

A new topic.suffix config option has been added that allows users to specify a suffix to be added to all topics. If using a custom query and `topic.prefix` is specifying a full topic name, this option is ignored.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -22,5 +22,5 @@
               files="(DataConverter|GenericDatabaseDialect|JdbcSourceTask).java"/>
 
     <suppress checks="ParameterNumber"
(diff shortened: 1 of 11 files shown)


desiderantes:
[clabot:check]

desiderantes:
Please tell me if this is something of interest, so that I can continue trying to make this work

PR #1112, opened 2021-09-10
Fix Sybase dialect upsert syntax

Cyril-Engels:
## Problem
With the current merge query, user will hit the following SQL exception.
```
WARN Write of 1 records failed, remainingRetries=10 (io.confluent.connect.jdbc.sink.JdbcSinkTask:92)
java.sql.BatchUpdateException: Incorrect syntax near the keyword 'with'.
```
According to the [docs](https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc00938.1570/html/locking/locking103.htm), Sybase does support HOLDLOCK syntax directly after table name. But appending HOLDLOCK to the table name also fails in the case of merge. While the [documentation for merge command](https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc36272.1600/doc/html/san1393051041603.html) did not mention HOLDLOCK at all.

The other problem is that the driver also complains about the trailing semicolon.

## Solution
Remove HOLDLOCK from merge query.
Remove semicolon from merge query.
 

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java
@@ -250,32 +250,32 @@ public class SybaseDatabaseDialect extends GenericDatabaseDialect {
     builder.append(table);
     builder.append(" AS target using (select ");
     builder.appendList()
-        .delimitedBy(", ")
-        .transformedBy(ExpressionBuilder.columnNamesWithPrefix("? AS "))
-        .of(keyColumns, nonKeyColumns);
+           .delimitedBy(", ")
+           .transformedBy(ExpressionBuilder.columnNamesWithPrefix("? AS "))
+           .of(keyColumns, nonKeyColumns);
     builder.append(") AS incoming on (");
     builder.appendList()
-        .delimitedBy(" and ")
-        .transformedBy(this::transformAs)
-        .of(keyColumns);
+           .delimitedBy(" and ")
+           .transformedBy(this::transformAs)
+           .of(keyColumns);
     builder.append(")");
     if (nonKeyColumns != null && !nonKeyColumns.isEmpty()) {
       builder.append(" when matched then update set ");
       builder.appendList()
-          .delimitedBy(",")
-          .transformedBy(this::transformUpdate)
-          .of(nonKeyColumns);
+             .delimitedBy(",")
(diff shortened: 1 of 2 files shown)


liukrimhrim (approved):
LGTM

gharris1727:
@Cyril-Engels @liukrimhrim 
This doesn't look safe, and would change the concurrency semantics of the merge:
https://weblogs.sqlteam.com/dang/2009/01/31/upsert-race-condition-with-merge/
Did sybase change the syntax between versions, and change the built-in synchronization?

At a minimum, this change is not safe to backport, as it will produce race conditions on older versions of sybase.

Cyril-Engels:
> @Cyril-Engels @liukrimhrim
> This doesn't look safe, and would change the concurrency semantics of the merge:
> https://weblogs.sqlteam.com/dang/2009/01/31/upsert-race-condition-with-merge/
> Did sybase change the syntax between versions, and change the built-in synchronization?
> 
> At a minimum, this change is not safe to backport, as it will produce race conditions on older versions of sybase.

@gharris1727 I'm afraid this merge statement probably never worked. 
At least in the last 2 versions, the official doc did not mention HOLDLOCK keyword in merge command. The latest version of Sybase is 16.0, which was published in 2014.

gharris1727:
Ah, `MERGE [ INTO ] <target_table> [ WITH ( <merge_hint> ) ]` is part of SqlServer: https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver15 and not sybase. The page I linked earlier also calls out sqlserver. It appears that maybe this was copy-pasted from sqlserver to sybase, and never tested? Very strange.

Cyril-Engels:
> Ah, `MERGE [ INTO ] <target_table> [ WITH ( <merge_hint> ) ]` is part of SqlServer: https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver15 and not sybase. The page I linked earlier also calls out sqlserver. It appears that maybe this was copy-pasted from sqlserver to sybase, and never tested? Very strange.

Yes, that's what I thought too. This change was tested against a sybase container https://hub.docker.com/r/jaschweder/sybase/tags?page=1&ordering=last_updated

PR #1113, opened 2021-09-16
MINOR: fix the build

ncliang:
## Problem
All `5.0.x`+ builds failing with 
```
01:33:34  [2021-09-04 08:33:34,264] ERROR mysql: /tmp/MariaDB4j/base/bin/mysql: error while loading shared libraries: libncurses.so.5: cannot open shared object file: No such file or directory (ch.vorburger.exec.ManagedProcess:71)
```

## Solution
Root cause is https://github.com/confluentinc/jenkins-common/pull/373/files changing the default jenkins node to `docker-debian-10-jdk8` , which likely does not have `libncurses` package installed. Change it back to previous label we were using - `docker-oraclejdk8`

- [ ] yes
- [x] no

##### If yes, where?
Other connectors either already have specified nodeLabel `docker-oraclejdk8` or are fine building with the new `docker-debian-10-jdk8` image.

## Test Strategy
Rely on PR builder to run in Jenkins

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,7 @@
 #!/usr/bin/env groovy
 common {
   slackChannel = '#connect-warn'
+  nodeLabel = 'docker-oraclejdk8'
   upstreamProjects = 'confluentinc/common'
   pintMerge = true
 }


gharris1727 (approved):
I think this is acceptable because the connector is no longer packaged with CP.
Otherwise I think that we should try and migrate to the debian images by asking them to add libncurses, or installing it ourselves, or figuring out how to remove that dependency.

ddasarathan (approved):

PR #1117, opened 2021-09-20
Revert "MINOR: fix the build"

gharris1727:
Reverts confluentinc/kafka-connect-jdbc#1113

We should now have libcurses in the build image.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,6 @@
 #!/usr/bin/env groovy
 common {
   slackChannel = '#connect-warn'
-  nodeLabel = 'docker-oraclejdk8'
   upstreamProjects = 'confluentinc/common'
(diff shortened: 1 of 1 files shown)


ncliang:
I think I would prefer if we still explicitly specified the label to use so that we are not unknowingly affected by upstream default label changes again in the future. WDYT @gharris1727

jkao97:
> I think I would prefer if we still explicitly specified the label to use so that we are not unknowingly affected by upstream default label changes again in the future. WDYT @gharris1727

 If so please change to `docker-debian-jdk8` instead.

jkao97:
I merged changes to change nodelabel here after green build: 
https://github.com/confluentinc/kafka-connect-jdbc/pull/1119

This PR can be closed.

PR #1118, opened 2021-09-23
Include the TableAlterOrCreateException in creatOrAmend path for sink…

ddasarathan:
… connectors

## Problem
Caused by: java.lang.NullPointerException: Cannot invoke "io.confluent.connect.jdbc.util.TableDefinition.columnNames()" because "tableDefn" is null
	at io.confluent.connect.jdbc.sink.DbStructure.amendIfNecessary(DbStructure.java:151)
	at io.confluent.connect.jdbc.sink.DbStructure.createOrAmendIfNecessary(DbStructure.java:80)
	at io.confluent.connect.jdbc.sink.BufferedRecords.add(BufferedRecords.java:123)
	at io.confluent.connect.jdbc.sink.JdbcDbWriter.write(JdbcDbWriter.java:74)
	at io.confluent.connect.jdbc.sink.JdbcSinkTask.put(JdbcSinkTask.java:84)

## Solution
This is because TableAlterOrCreateException is unhandled and is silently ignored. 

- [ ] yes
- [x] no

## Test Strategy
Test with unit tests. 

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
@@ -60,7 +60,7 @@ public class DbStructure {
       final Connection connection,
       final TableId tableId,
       final FieldsMetadata fieldsMetadata
-  ) throws SQLException {
+  ) throws SQLException, TableAlterOrCreateException {
     if (tableDefns.get(connection, tableId) == null) {
       // Table does not yet exist, so attempt to create it ...
       try {
@@ -75,6 +75,9 @@ public class DbStructure {
         } catch (SQLException e) {
           throw sqle;
         }
+      } catch (TableAlterOrCreateException te) {
+        log.warn(te.getMessage());
+        throw te;
       }
     }
     return amendIfNecessary(config, connection, tableId, fieldsMetadata, config.maxRetries);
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/sink/DbStructure.java
```diff
@@ -75,6 +75,9 @@ public boolean createOrAmendIfNecessary(
         } catch (SQLException e) {
           throw sqle;
         }
+      } catch (TableAlterOrCreateException te) {
+        log.warn(te.getMessage());
+        throw te;
```
liukrimhrim, line 80: Maybe I missed something, how is this different from what was before, other than a warn msg? how would this help with the NPE issue.

ddasarathan: If we do not catch this exception here, it does not bubble up. Instead, we try to proceed with table creation and that results in a NPE

ddasarathan: Now that this is caught and thrown, it is handled gracefully and results in an error that either table is not already created or auto create is not configured.

liukrimhrim: @ddasarathan explained to me via zoom. Now the code makes sense to me.

liukrimhrim (approved):
LGTM, thanks @ddasarathan

ddasarathan:
Thanks @liukrimhrim

ddasarathan:
Closed because https://github.com/confluentinc/kafka-connect-jdbc/pull/1120 was created against `10.0.x`

PR #1124, opened 2021-10-18
Fix the JDBC data loss issue, by updating the internal offset value

ddasarathan:
## Problem
The connector experiences data loss when it sees JDBC connection timeouts. This has particularly happened when `querier.next()`  is called midway through a batch. 
The error is

```
2021-10-01 17:13:04,414 ERROR (io.confluent.connect.jdbc.source.JdbcSourceTask:419) null - SQL exception while running query for table: <xxxx> 
java.sql.SQLRecoverableException: Closed Connection: next
at oracle.jdbc.driver.InsensitiveScrollableResultSet.ensureOpen(InsensitiveScrollableResultSet.java:110)
at oracle.jdbc.driver.InsensitiveScrollableResultSet.next(InsensitiveScrollableResultSet.java:404)
at io.confluent.connect.jdbc.source.TimestampTableQuerier.next(TimestampTableQuerier.java:107)
at io.confluent.connect.jdbc.source.JdbcSourceTask.poll(JdbcSourceTask.java:383)
```
The problem is because we cleanup the resultSet, stmt and release the locks. None of the records processed from the entire ResultSet is committed. But there are two important state objects that are not cleaned up properly in `TimestampTableQuerier` class, the `nextRecord` and `offset`

These values store state from the previous queries ResultSet and affect the offset committing logic.

## Solution
The solution is to update an internal `committedOffset` everytime after a successful `poll()` run. A failure in `poll()` would reset all the state variables. 

- [ ] yes
- [x] no

## Test Strategy
- Unit tests
- Manual tests

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
 Release a new minor version
 Merging to Master 
 NA

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -387,7 +387,7 @@ public class JdbcSourceTask extends SourceTask {
         if (!hadNext) {
           // If we finished processing the results from the current query, we can reset and send
           // the querier to the tail of the queue
-          resetAndRequeueHead(querier);
+          resetAndRequeueHead(querier, false);
         }
 
         if (results.isEmpty()) {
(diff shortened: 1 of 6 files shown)


ddasarathan:
Outdated

Issue #1088, opened 2021-06-30
Close database connection when sink connector is idle


philipp94831:
We use Aurora serverless databases which automatically scale to zero if no connections to the database are active. However, the JDBC sink connector does not close the database connection if it is idle for some time. Using a connection pooling library such as Hikari allows to properly close idle connections (https://github.com/brettwooldridge/HikariCP#:~:text=%F0%9F%94%A2-,minimumIdle,-This%20property%20controls). It would be nice to have similar behavior with the sink connector

PR #1128, opened 2021-10-21
1088: Add connection ttl config

Bradleywboggs:
## Problem
Original issue: https://github.com/confluentinc/kafka-connect-jdbc/issues/1088

The current implementation of the source and sink connectors maintains a single database connection indefinitely across all connectors and tasks, only cycling if the connection becomes unhealthy. However,  in some circumstances, long-running connections can eat up too many resources and eventually crash the db server, resulting in outage, data corruption,
 or data loss.

### Overview
 - Whenever `CachedConnnectionProvider::getConnection` is called, check if the connectionTTL is expired, and if so, cycle the connection.
 - During long polling intervals (source) or when no new messages are available (sink), check  if  connection is expired, if so, close it until there is more work to do.

#### Change to Config
- Add a config property to the source and sink connectors: `connection.ttl.ms`.
  If this property is not present, default to the current infinite connection behavior 
  (the ttl value for an infinite connection is -1)

#### Changes to CachedConnectionProvider
- Pass `connectionTTL` to the `CachedConnnectionProvider` constructor from: `JdbcSourceConnector`, `JdbcSourceTask`, `JdbcDBWriter` . 
- Add additional internal property to the `CachedConnectionProvider`: `connectionStartTs` which is set with the current timestamp in millis whenever  a connection object is successfully returned from `provider.getConnection()`.
- Add `connectionIsExpired` method to the CachedConnectionProvider class, which checks the `connectionTTL` against the current time and the `connectionStartTs`.
- Add a check for `connectionIsExpired` in the `CachedConnectionProvider::getConnection()` method. If `false` cycle the connection.

##### Change to JdbcSourceTask
- Add a check for `connectionProvider.connectionIsExpired` prior to `time.sleep()` in `poll` method, and then `connectionProvider.close()` if it's past its ttl.

##### Change to JdbcSinkTask and JdbcWriter
- When `JdbcSinkTask::put` has empty `records`, call new method `JdbcWriter::handleIdling` which checks for an expired connection and closes it if it's expired.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [x] System tests
- [x] Manual tests

## Release Plan

This would be a part of a new minor version release, as it adds functionality but is completely backwards compatible with the existing interface and behavior.

This would be a merge to master.
 
This doesn't require a revert or roll-back.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
@@ -121,7 +121,7 @@ public class JdbcDbWriter {
   public void handleIdling() {
     if (cachedConnectionProvider.connectionIsExpired()) {
       log.debug("No results found. Closing database connection until there's work to do...");
-      closeQuietly();
+      cachedConnectionProvider.close();
     }
   }
 }
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
@@ -71,7 +71,8 @@ public class JdbcSinkTask extends SinkTask {
   @Override
   public void put(Collection<SinkRecord> records) {
     if (records.isEmpty()) {
-      // closes the database connection if there's no work to do and the connection.ttl.ms is expired
+      // closes the database connection
+      // if there's no work to do and the connection.ttl.ms is expired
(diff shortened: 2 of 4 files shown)


Bradleywboggs:
@ncliang Do you have any additional pointers on trying to get this to work as I need, based on my config above? Otherwise, you can feel free to close the PR if you don't think this is right for the project, and I'll just operate from my fork for now. Thanks again!!

PR #1130, opened 2021-10-27
MINOR: call `setCharacterStream` API with length

ncliang:
## Problem
Performance regression when writing short strings into CLOB type columns

## Solution
Call the `setCharacterStream` and `setNCharacterStream` variants that take the length of the stream as argument. These variants can use the length to determine the most efficient binding to use. From javadoc:
> Consult your JDBC driver documentation to determine if it might be more efficient to use a version of setCharacterStream which takes a length parameter.

- [ ] yes
- [x] no

## Test Strategy
Unit tests were updated. Manually verified performance regression fixed.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
Backport fix all the way back to `5.0.x` when #1004 introduced this logic.


aakashnshah (approved):
This is awesome, thanks Nigel! LGTM

Issue #675, opened 2019-07-05
long table names cause NPE


markteehan:
I am JDBC Sinking all topics into Postgres. Long table names such as 
k__confluent-controlcenter-5-3-0-1-metrics-trigger-measurement-rekey
fail with NPE and short topic names (such as connect-offsets-1359-A) succeed.

If i create this table manually, it succeeds.
CREATE TABLE "k__confluent-controlcenter-5-3-0-1-metrics-trigger-measurement-rekey" ("line" BYTEA NULL)

The connector appears to fail when the topic name exceeds a certain length.

     {
            "name": "jdbc_sink_NONE",
             "config": {
                               "connector.class":"io.confluent.connect.jdbc.JdbcSinkConnector"
                     ,                    "name":"jdbc_sink_NONE"
                     ,            "topics.regex":"connect(.*)"
                     ,       "table.name.format":"k_${topic}"
                     ,             "insert.mode":"insert"
                     ,          "connection.url":"jdbc:postgresql://localhost/db_gdelt?currentSchema=public"
                     ,          "connection.user":"gdelt"
                     ,      "connection.password":"gdelt_pw"
                     ,        "poll.interval.ms":"3000"
                     ,         "numeric.mapping":"best_fit"
                     ,               "tasks.max":"1"
                     ,             "auto.create":"true"
                     ,             "auto.evolve":"false"
                     ,            "value.schema":"{\"name\":\"schema.struct\" ,\"type\":\"STRUCT\",\"isOptional\":true,\"fieldSchemas\":{\"KEY\":{\"type\":\"STRING\",\"isOptional\":true},\"VALUE\":{\"type\":\"INT64\",\"isOptional\":true}"
                     , "transforms":"HoistField","transforms.HoistField.type":"org.apache.kafka.connect.transforms.HoistField$Value","transforms.HoistField.field":"line"
                     }
      }

```[2019-07-05 15:24:28,994] DEBUG starting (io.confluent.monitoring.clients.interceptor.MonitoringInterceptor:176)
[2019-07-05 15:24:29,030] DEBUG Received 500 records. First record kafka coordinates:(_confluent-controlcenter-5-3-0-1-metrics-trigger-measurement-rekey-0-123800). Writing them to the database... (io.confluent.connect.jdbc.sink.JdbcSinkTask:68)
[...]

PR #1143, opened 2021-12-07
CCDB-4454: Automatically truncate long table names to respect PostgreSQL length limits

C0urante:
[Jira](https://confluentinc.atlassian.net/browse/CCDB-4454), [GitHub issue](https://github.com/confluentinc/kafka-connect-jdbc/issues/675)

## Problem
When a table is created in PostgreSQL, if its name exceeds the maximum permitted length (see https://www.postgresql.org/docs/14/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), it will be truncated automatically. For example, a table named "toolongtoolong" would be truncated to "too" if the maximum permitted length were 3 bytes.

Unfortunately, this truncation does not take place automatically when the connector checks to see if a table exists, so even if the connector is able to successfully create a table, if the name was truncated during that creation, the next time the connector checks to see if the table exists, it will appear as if the table is missing (since no table with the non-truncated name exists in the database).

This causes a null pointer exception in `DbStructure::amendIfNecessary`.

## Solution
Determine the maximum-permitted length of a table name (see https://stackoverflow.com/questions/27865770/how-long-can-postgresql-table-names-be/27865772#27865772), then automatically truncate table names.

- [ ] yes
- [x] no

## Test Strategy
Unit tests are added to verify the logic for computing the maximum table name length and performing truncation based on it.
I also reproduced the issue locally using a Dockerized PostgreSQL instance and version 10.2.5 of the connector, and verified that with these changes, the connector was able to function properly.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
Merge to 5.2.x, `pint merge` forward.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -15,23 +15,6 @@
 
 package io.confluent.connect.jdbc.dialect;
 
-import java.util.Map;
-import org.apache.kafka.common.config.AbstractConfig;
-import org.apache.kafka.connect.data.Date;
-import org.apache.kafka.connect.data.Decimal;
-import org.apache.kafka.connect.data.Schema;
-import org.apache.kafka.connect.data.Schema.Type;
-import org.apache.kafka.connect.data.SchemaBuilder;
-import org.apache.kafka.connect.data.Time;
-import org.apache.kafka.connect.data.Timestamp;
-
-import java.sql.Connection;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Types;
-import java.util.Collection;
-
 import io.confluent.connect.jdbc.dialect.DatabaseDialectProvider.SubprotocolBasedProvider;
 import io.confluent.connect.jdbc.sink.metadata.SinkRecordField;
 import io.confluent.connect.jdbc.source.ColumnMapping;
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -77,6 +79,70 @@ public PostgreSqlDatabaseDialect(AbstractConfig config) {
...
+      result = Integer.MAX_VALUE;
+    }
+    return result;
+  }
+
+  @Override
+  public TableId parseTableIdentifier(String fqn) {
+    TableId result = super.parseTableIdentifier(fqn);
+    if (maxIdentifierLength > 0 && result.tableName().length() > maxIdentifierLength) {
+      String newTableName = result.tableName().substring(0, maxIdentifierLength);
```
ddasarathan, line 131: Could we unintentionally cause table name collisions by doing this?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -280,4 +286,118 @@ public void shouldSanitizeUrlWithCredentialsInUrlProperties() {
...
+    int actualMaxLength = PostgreSqlDatabaseDialect.computeMaxIdentifierLength(connection);
+
+    assertEquals(expectedMaxLength, actualMaxLength);
+  }
+
+  @Test
+  public void shouldGracefullyHandleErrorWhenComputingMaxTableNameLength() throws Exception {
+    Statement statement = mock(Statement.class);
+    when(statement.executeQuery("SELECT length(repeat('1234567890', 1000)::NAME);"))
+        .thenThrow(new SQLException("I plead the fifth"));
```
ddasarathan, line 313: :)

ddasarathan (approved):
It looks like we need to document this change as it could otherwise lead to unintended truncation resulting in a collision. Otherwise LGTM

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
C0urante: (:

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
C0urante: Yes, this is definitely possible. @gharris1727 even noted it a couple years ago on the [GitHub issue](https://github.com/confluentinc/kafka-connect-jdbc/issues/675#issuecomment-566337721) linked in the description:

> ... it could cause table collisions, if two topics differ by only their truncated suffixes.

I think the approach here is still reasonable since we are only adopting to behavior that PostgreSQL applies downstream. If users encounter issues (such as collision) because of table name truncation, it's difficult to argue that it's the connector's fault when they'd see the exact same effects if they ran the same DDL/DML statements against the database by hand.

One approach we might consider is adding a configuration property that gives users some control over this situation, including but not limited to these options:
- Fail if the table name is too long
- Truncate if the table is too long
- Attempt to deduce the maximum table length and truncate accordingly, but disable truncation if the table length cannot be determined

I opted for an approach that could be backported as it did not introduce a new configuration property and, IMO, has reasonable behavior. But this is a bit borderline and if you have different thoughts on how we can/should handle this I think it's worth it to have that discussion so that we can (hopefully) get this right the first time.

C0urante:
Thanks @ddasarathan. I'll file a docs ticket to add this info if we end up going this route. Curious what your thoughts are on whether this is the best approach for the bug; LMKWYT in the comment thread.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
ddasarathan: I agree with the approach you have in this PR. If you have a follow-up PR on the docs repo, that should be good enough. This way it's easy to explain to unsuspecting customers.

Issue #1142, opened 2021-12-03
Entire Kafka Connect cluster stuck because of a stuck connector


colinsmetz:
## Context

* Kafka Connect 6.1.2
* JDBC connector 10.2.1

## Current behaviour

We have stumbled upon an issue on a running cluster with multiple source connectors:

* One of our connectors was a JDBC source connector connected to an SQL Server database (using the [mssql-jdbc](https://github.com/microsoft/mssql-jdbc) driver).
* It turns out that the SQL server instance had a problem causing all queries to be stuck forever, including `sp_tables` to list the tables.
* It made [this call to `metadata.getTables`](https://github.com/confluentinc/kafka-connect-jdbc/blob/v10.2.1/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java#L417) hang forever, and the connector could never start nor fail.
* After some time, the entire Kafka Connect cluster was unavailable:
  * Workers seemed unable to communicate with each other and elect a leader.
  * The workers left their consumer group, and the entire consumer group was deleted after a while.
  * The REST API was not responding (`{"error_code":500,"message":"Request timed out"}` for most requests).
* Pausing (just before the deletion of the consumer group) or deleting the problematic connector allowed the cluster to run normally again.

We found [this issue](https://issues.apache.org/jira/browse/KAFKA-9374) in Kafka that seemed related to our problem, but even though we should have that fix on our cluster, our problem still appeared.

## Expected behaviour

We expected the connector to:

* Timeout after being stuck on the `getTables` for too long
* Fail the connector, without breaking the whole cluster

PR #1145, opened 2021-12-09
CCDB-4476: Prevent blocking on database connection in JdbcSourceConnector::taskConfigs

C0urante:
[Jira](https://confluentinc.atlassian.net/browse/CCDB-4476), [GitHub issue](https://github.com/confluentinc/kafka-connect-jdbc/issues/1142)

## Problem
The connector can block in its `taskConfigs` method while [invoking `tableMonitorThread::tables`](https://github.com/confluentinc/kafka-connect-jdbc/blob/26cc0b1e309f15f368fbf6d2cd6c8665770c1608/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java#L149), which is [synchronized](https://github.com/confluentinc/kafka-connect-jdbc/blob/26cc0b1e309f15f368fbf6d2cd6c8665770c1608/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java#L96) and therefore blocks on the completion of any pending calls to the [also-synchronized `updateTables` method](https://github.com/confluentinc/kafka-connect-jdbc/blob/26cc0b1e309f15f368fbf6d2cd6c8665770c1608/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java#L138). This method (`updateTables`) may block for a long or even indefinite time as it interacts with the database and may hang on network connectivity issues, deadlock, etc.

With https://github.com/apache/kafka/pull/8069, we made Connect workers more resilient to connectors that would hang during invocations of methods like start, stop, and validate; the impact was lowered from blocking the worker's herder thread (which would cause some REST requests to time out and the worker to eventually fall out of the cluster) to just causing a thread to be allocated to that connector while it remained blocked. With https://github.com/apache/kafka/pull/9669, we fixed a similar issue where the worker's herder thread would block on invocations of `SourceTask::stop` (all other task interactions (except for `version`) such as `start`, `poll`, `put` were already performed safely without risk of blocking the worker's herder thread).

However, some connector methods (which were deemed less likely to hang) have not been safeguarded in this fashion yet, including `taskConfigs`.

As a result, if a JDBC source connector blocks during an invocation of `taskConfigs` (due to a block in the table monitoring thread), the entire worker's herder thread becomes blocked, which causes most REST requests to fail and for the worker to fall out of the cluster and have its tasks and connectors reassigned (even though they will still be running on the current worker as well).

## Solution
Modify the table monitoring thread logic to remove all `synchronized` blocks and instead employ a snapshot-based approach. When `TableMonitorThread::tables` is invoked, no attempt is made to block for any currently-in-progress reads of the database's list of tables to complete, and only the latest complete set of tables is utilized. If no table read has succeeded yet, an `INFO`-level log message is emitted and the connector generates an empty set of task configs. If a table read has succeeded but there are no eligible tables to read from (after applying any user-supplied whitelist/blacklist rules), the connector also generates an empty set of task configs, but emits a `WARN`-level log message as well.

Additionally, calls are added to [`ConnectorContext::raiseError`](https://kafka.apache.org/30/javadoc/org/apache/kafka/connect/connector/ConnectorContext.html#raiseError(java.lang.Exception)) when the table monitor thread encounters unrecoverable errors (i.e., duplicate table names and unexpected failures). This signals to Connect that no further calls to `taskConfigs` should be made (see the logic in the [`DistributedHerder`](https://github.com/apache/kafka/blob/e1dba7af57ec69eb2bbf5a8f4d009502139f0e24/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L1552-L1555) and [`StandaloneHerder`](https://github.com/apache/kafka/blob/e1dba7af57ec69eb2bbf5a8f4d009502139f0e24/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/standalone/StandaloneHerder.java#L393-L396) classes), which is better than the current behavior, where the connector throws an exception from `taskConfigs` and the Connect worker responds by waiting for 250 milliseconds and then invoking the method again until it finally succeeds.

- [ ] yes
- [x] no

## Test Strategy
Existing unit tests are augmented to account for this new logic, and a new case is added that explicitly verifies that `TableMonitorThread::tables` does not block even when an in-progress call to `updateTables` is blocked.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
Merge to 5.2.x, `pint merge` forward. Will likely publish new independent releases on the 10.0.x, 10.1.x, and 10.2.x branches once https://github.com/confluentinc/kafka-connect-jdbc/pull/1144 and https://github.com/confluentinc/kafka-connect-jdbc/pull/1143 are either rejected or merged as well.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
@@ -28,9 +28,11 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 /**
@@ -45,10 +47,9 @@ public class TableMonitorThread extends Thread {
   private final ConnectorContext context;
   private final CountDownLatch shutdownLatch;
   private final long pollMs;
-  private Set<String> whitelist;
-  private Set<String> blacklist;
-  private List<TableId> tables;
-  private Map<String, List<TableId>> duplicates;
+  private final Set<String> whitelist;
+  private final Set<String> blacklist;
+  private final AtomicReference<List<TableId>> tables;
 
   public TableMonitorThread(DatabaseDialect dialect,
       ConnectionProvider connectionProvider,
@@ -64,8 +65,7 @@ public class TableMonitorThread extends Thread {
     this.pollMs = pollMs;
     this.whitelist = whitelist;
     this.blacklist = blacklist;
-    this.tables = null;
-
+    this.tables = new AtomicReference<>();
   }
 
   @Override
@@ -77,8 +77,7 @@ public class TableMonitorThread extends Thread {
           context.requestTaskReconfiguration();
         }
       } catch (Exception e) {
-        context.raiseError(e);
-        throw e;
+        throw fail(e);
       }
 
       try {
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -170,39 +185,28 @@ private synchronized boolean updateTables() {
...
-      // first table lookup
-      return previousTables != null;
-    }
+  /**
+   * Fail the connector with an unrecoverable error and stop the table monitoring thread
+   * @param t the cause of the failure
+   * @return a {@link RuntimeException} that can be thrown from the calling method, for convenience
+   */
+  private RuntimeException fail(Throwable t) {
+    String message = "Encountered an unrecoverable rrorwhile reading tables from the database ";
```
aakashnshah, line 201:
```suggestion
    String message = "Encountered an unrecoverable error while reading tables from the database";
```

aakashnshah (approved):
LGTM, with one minor comment.

C0urante:
Thanks @aakashnshah!

PR #1149, opened 2022-01-07
CCDB-4340 - Add support for the new Timestamp granularity features in the jdbc source connector

ddasarathan:
## Problem
Background
Oracle DB stores a timestamp with microseconds granularity. The JDBC connector produces events that have the same timestamps when the connector rounds up timestamp to milliseconds, causing downstream applications not to sort events properly and not to be able to compare records with higher precision. 

## Solution
Adding micros and millis format negatively impacts the rollout as the offset management becomes backward-incompatible due to loss of precision. So we will have three new config types,

- `connect-logical` (default)
- `nanos-long` (nanos is epoch as Int64)
- `nanos-string` (nanos is epoch as String)
- `nanos-iso-datetime-string` (`yyyy-MM-dd'T'HH:mm:ss.n` for eg., `2022-01-05T12:07:56.702041000`)

We also discussed not allowing the customers to define the iso string format as it might risk losing the column's precision, leading to incompatibility during rollouts. 

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -1391,7 +1391,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
       case Types.TIMESTAMP: {
         return rs -> {
           Timestamp timestamp = rs.getTimestamp(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
-          return tsGranularity.convert.apply(timestamp);
+          return tsGranularity.fromTimestamp.apply(timestamp);
         };
       }
 
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -91,6 +91,10 @@ import io.confluent.connect.jdbc.util.TableDefinition;
 import io.confluent.connect.jdbc.util.TableId;
 import io.confluent.connect.jdbc.util.TableType;
 
+import static io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig.TIMESTAMP_GRANULARITY_LONG_NANOS;
+import static io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig.TIMESTAMP_GRANULARITY_STRING_NANOS;
+import static io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig.TIMESTAMP_GRANULARITY_STRING_ISO_DATETIME;
+
 /**
  * A {@link DatabaseDialect} implementation that provides functionality based upon JDBC and SQL.
  *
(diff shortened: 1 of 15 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -152,6 +156,7 @@ public DatabaseDialect create(AbstractConfig config) {
   private volatile JdbcDriverInfo jdbcDriverInfo;
   private final int batchMaxRows;
   private final TimeZone timeZone;
+  private final String tsGranularity;
```
ncliang, line 159: Any reason this is a string instead of an enum? See `NumericMapping` for an example of using enum in config.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1385,7 +1405,19 @@ protected ColumnConverter columnConverterFor(
 
       // Timestamp is a date + time
       case Types.TIMESTAMP: {
-        return rs -> rs.getTimestamp(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
+        return rs -> {
+          Timestamp timestamp = rs.getTimestamp(col, DateTimeUtils.getTimeZoneCalendar(timeZone));
+          switch (tsGranularity) {
+            case TIMESTAMP_GRANULARITY_LONG_NANOS:
+              return DateTimeUtils.toEpochNanos(timestamp);
```
ncliang, line 1412: If we did this in an enum, we could just have a method on the enum to convert the timestamp, and have this just be a call into `tsGranularity.convert(timestamp)` . WDYT?

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingCriteria.java
```diff
@@ -218,15 +224,40 @@ public TimestampIncrementingOffset extractValues(
...
+                    + e.getMessage());
+          }
+          break;
+        case TIMESTAMP_GRANULARITY_STRING_ISO_DATETIME:
+          ts = DateTimeUtils.toTimestampFromIsoDateTime(
+              (String) record.get(timestampColumn)
+          );
+          break;
+        default:
+          ts = (Timestamp) record.get(timestampColumn);
```
ncliang, line 258: This switch block could also be handled with a `timestampGranularity.convert` if we had the granularity exposed as enum, and a `convert` method on the enum.

ncliang (commented):
Thanks for the change @ddasarathan . I have a suggestion on using enums to simplify the logic a bit so that we wouldn't have to repeatedly do switches based on the string value.

aakashnshah:
@ddasarathan I second @ncliang's suggestion to use an enum. I think it follows our current code pattern/standards, and makes it a little more readable by distinguishing the types of granularity vs what an actual configuration is. One high level question I had - what happens if a user configures a higher level of granularity than the DB they are connecting to provides?

ddasarathan:
Closing in favor of https://github.com/confluentinc/kafka-connect-jdbc/pull/1152

ddasarathan:
@aakashnshah I made the changes to enum, the results can be found in https://github.com/confluentinc/kafka-connect-jdbc/pull/1152

> what happens if a user configures a higher level of granularity than the DB they are connecting to provides?
For eg., if the user configured `nanos-string` and the db has a precision of micros, then the output would look like `1641945720427176000` the nanos component will have 0s in it.

PR #1152, opened 2022-01-11
CCDB-4340 - Add support for the new Timestamp granularity features in the jdbc source connector

ddasarathan:
## Problem
Background
Oracle DB stores a timestamp with microseconds granularity. The JDBC connector produces events that have the same timestamps when the connector rounds up timestamp to milliseconds, causing downstream applications not to sort events properly and not to be able to compare records with higher precision. 

## Solution
Adding micros and millis format negatively impacts the rollout as the offset management becomes backward-incompatible due to loss of precision. So we will have three new config types,

- `connect_logical` (default)
- `nanos_long` (nanos is epoch as Int64)
- `nanos_string` (nanos is epoch as String)
- `nanos_iso_datetime_string` (`yyyy-MM-dd'T'HH:mm:ss.n` for eg., `2022-01-05T12:07:56.702041000`)

We also discussed not allowing the customers to define the iso string format as it might risk losing the column's precision, leading to incompatibility during rollouts. 

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -187,6 +187,21 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
       + "Use -1 to use the current time. If not specified, all data will be retrieved.";
   public static final String TIMESTAMP_INITIAL_DISPLAY = "Unix time value of initial timestamp";
 
+  public static final String TIMESTAMP_GRANULARITY_CONNECT_LOGICAL = "connect-logical";
+  public static final String TIMESTAMP_GRANULARITY_LONG_NANOS = "nanos-long";
+  public static final String TIMESTAMP_GRANULARITY_STRING_NANOS = "nanos-string";
+  public static final String TIMESTAMP_GRANULARITY_STRING_ISO_DATETIME = "iso-datetime-string";
+  public static final String TIMESTAMP_GRANULARITY_CONFIG = "timestamp.granularity";
+  public static final String TIMESTAMP_GRANULARITY_DOC =
+      "Define the granularity of the Timestamp column. Options include: \n"
+          + "  * connect-logical (default): represents timestamp values using Kafka Connect's "
+          + "built-in representations "
+          + "  * nanos-long: represents timestamp values as nanos since epoch"
+          + "  * nanos-string: represents timestamp values as nanos since epoch in string"
+          + "  * iso-datetime-string: uses the iso format 'yyyy-MM-dd'T'HH:mm:ss.n'";
+  public static final String TIMESTAMP_GRANULARITY_DISPLAY = "Timestamp granularity for "
+      + "timestamp columns";
+
   public static final String TABLE_POLL_INTERVAL_MS_CONFIG = "table.poll.interval.ms";
   private static final String TABLE_POLL_INTERVAL_MS_DOC =
       "Frequency in ms to poll for new or removed tables, which may result in updated task "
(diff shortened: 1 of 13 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -786,6 +815,53 @@ public static NumericMapping get(JdbcSourceConnectorConfig config) {
...
+        timestamp -> String.valueOf(DateTimeUtils.toEpochNanos(timestamp)),
+        timestamp -> DateTimeUtils.toTimestamp((String) timestamp)),
+
+    NANOS_ISO_DATETIME_STRING(optional -> optional
+        ? Schema.OPTIONAL_STRING_SCHEMA : Schema.STRING_SCHEMA,
+        DateTimeUtils::toIsoDateTimeString,
+        timestamp -> DateTimeUtils.toTimestampFromIsoDateTime((String) timestamp));
+
+    public final Function<Boolean, Schema> schemaFunction;
+    public final Function<Timestamp, Object> convert;
```
ncliang, line 839: nit: `fromTimestamp` would match better with the `toTimestamp` function below.

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingCriteria.java
```diff
@@ -186,16 +186,18 @@ protected void setQueryParametersTimestamp(
    * @param schema the record's schema; never null
    * @param record the record's struct; never null
    * @param previousOffset a previous timestamp offset if the table has timestamp columns
+   * @param timestampGranularity defines the configured grrityanualrity of the timestamp field
```
ncliang, line 189: nit: typo
```suggestion
   * @param timestampGranularity defines the configured granularity of the timestamp field
```

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingCriteria.java
```diff
@@ -218,17 +220,26 @@ public TimestampIncrementingOffset extractValues(
...
-      Timestamp ts = (Timestamp) record.get(timestampColumn);
-      if (ts != null) {
-        return ts;
+      try {
+        Timestamp ts = timestampGranularity.toTimestamp.apply(record.get(timestampColumn));
+        if (ts != null) {
+          return ts;
+        }
+      } catch (NumberFormatException  e) {
+        throw new ConnectException(
```
ncliang, line 239: Can this be caught and handled only in the enum `toTimestamp` method that would throw the exception?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -643,7 +660,19 @@ public void ensureValid(final String name, final Object value) {
...
         ++orderInGroup,
         Width.MEDIUM,
-        DB_TIMEZONE_CONFIG_DISPLAY);
+        DB_TIMEZONE_CONFIG_DISPLAY
+    ).define(
+        TIMESTAMP_GRANULARITY_CONFIG,
+        Type.STRING,
+        TimestampGranularity.DEFAULT,
+        TIMESTAMP_GRANULARITY_RECOMMENDER,
+        Importance.MEDIUM,
```
ncliang, line 669: I think this should be low importance. The default `connect_logical` should suffice in most cases unless finer granularity is desired.

ncliang (approved):
Thanks @ddasarathan . LGTM with some nits.

PR #1157, opened 2022-01-19
CCDB-4354: preflight cross-fields validation for configuration PK_MODE

fzhong-connect:
## Problem
Misconfigured `pk.mode` and `delete.enabled` combination should not be surfaced by throwing an exception when the connector or task starts. Instead, the connector should notify the user that it is configured incorrectly via the Connector::validate method.

## Solution

Do cross-fields configuration validation in `Connector::validate` method, utilizing existing logic from [PrimaryKeyModeRecommender](https://github.com/confluentinc/kafka-connect-jdbc/blob/348e49f4a93261dd7b1365f469af426fa9ee8c6c/src/main/java/io/confluent/connect/jdbc/util/PrimaryKeyModeRecommender.java#L41).

Ref:
1. [How to Write a Connector for Kafka Connect](https://www.confluent.io/blog/write-a-kafka-connect-connector-with-configuration-handling/)
2. [kafka-connect-elasticsearch](https://github.com/confluentinc/kafka-connect-elasticsearch/blob/6f1a02586748a549b9c09178a6506fd13dde2b77/src/main/java/io/confluent/connect/elasticsearch/Validator.java#L97)
3. [Connect Configuration Validation](https://kafka.apache.org/documentation/#connect_configs)

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
@@ -80,8 +80,8 @@ public class JdbcSinkConnector extends SinkConnector {
          */
         .ifPresent(pkMode -> {
           if (!pkMode.recommendedValues().contains(pkMode.value())) {
-            pkMode.addErrorMessage("'" + pkMode.name() + "' can only have value in list "
-                + pkMode.recommendedValues());
+            pkMode.addErrorMessage("'" + pkMode.value() + "' is invalid:"
+                + " Deletes are only supported for pk.mode 'record_key'");
           }
         });
 
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
@@ -83,7 +83,8 @@ public class JdbcSinkConnector extends SinkConnector {
         .filter(deleteEnabled -> Boolean.TRUE.equals(deleteEnabled.value()))
         .ifPresent(deleteEnabled -> configValue(config, PK_MODE)
             .ifPresent(pkMode -> {
-              if (!RECORD_KEY.name().toLowerCase(Locale.ROOT).equals(pkMode.value())) {
+              if (!RECORD_KEY.name().toLowerCase(Locale.ROOT).equals(pkMode.value())
+              && !RECORD_KEY.name().toUpperCase(Locale.ROOT).equals(pkMode.value())) {
                 String conflictMsg = "Deletes are only supported for pk.mode record_key";
                 pkMode.addErrorMessage(conflictMsg);
                 deleteEnabled.addErrorMessage(conflictMsg);
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
@@ -84,7 +84,7 @@ public class JdbcSinkConnector extends SinkConnector {
         .ifPresent(deleteEnabled -> configValue(config, PK_MODE)
             .ifPresent(pkMode -> {
               if (!RECORD_KEY.name().toLowerCase(Locale.ROOT).equals(pkMode.value())
-              && !RECORD_KEY.name().toUpperCase(Locale.ROOT).equals(pkMode.value())) {
+                  && !RECORD_KEY.name().toUpperCase(Locale.ROOT).equals(pkMode.value())) {
                 String conflictMsg = "Deletes are only supported for pk.mode record_key";
                 pkMode.addErrorMessage(conflictMsg);
                 deleteEnabled.addErrorMessage(conflictMsg);
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
@@ -15,6 +15,8 @@
 
 package io.confluent.connect.jdbc;
 
+import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.PK_MODE;
+
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.connector.Task;
@@ -65,8 +67,21 @@ public class JdbcSinkConnector extends SinkConnector {
 
   @Override
   public Config validate(Map<String, String> connectorConfigs) {
-    // TODO cross-fields validation here: pkFields against the pkMode
-    return super.validate(connectorConfigs);
+    /** get configuration parsed and validated individually */
+    Config config = super.validate(connectorConfigs);
+
+    config.configValues()
+        .stream()
+        /** use recommended values for configuration validation
+         * {@link PrimaryKeyModeRecommender#validValues}
+         */
+        .filter(cfg -> PK_MODE.equals(cfg.name())
+            && !cfg.recommendedValues().contains(cfg.value()))
+        .findFirst()
+        .ifPresent(cfg -> cfg.addErrorMessage("'" + cfg.value() + "' is not in valid values "
+            + cfg.recommendedValues()));
+
+    return config;
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
@@ -72,13 +72,18 @@ public class JdbcSinkConnector extends SinkConnector {
 
     config.configValues()
         .stream()
-        /** use recommended values for configuration validation
+        .filter(cfg -> PK_MODE.equals(cfg.name()))
+        .findFirst()
+        /**
+         * use recommended values for configuration validation
          * {@link PrimaryKeyModeRecommender#validValues}
          */
-        .filter(cfg -> PK_MODE.equals(cfg.name())
-            && !cfg.recommendedValues().contains(cfg.value()))
-        .findFirst()
-        .ifPresent(cfg -> cfg.addErrorMessage("'" + cfg.value() + "' is not valid"));
+        .ifPresent(pkMode -> {
+          if (!pkMode.recommendedValues().contains(pkMode.value())) {
+            pkMode.addErrorMessage("'" + pkMode.name() + "' can only have value in list "
+                + pkMode.recommendedValues());
+          }
+        });
 
     return config;
   }
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
@@ -17,7 +17,9 @@ package io.confluent.connect.jdbc;
 
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.DELETE_ENABLED;
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.PK_MODE;
+import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.PrimaryKeyMode.RECORD_KEY;
 
+import java.util.Locale;
 import java.util.Optional;
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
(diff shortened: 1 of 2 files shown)


src/test/java/io/confluent/connect/jdbc/JdbcSinkConnectorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    pkModeError = connector.validate(connConfig)
+        .configValues()
+        .stream()
+        .filter(cfg -> PK_MODE.equals(cfg.name()))
+        .map(ConfigValue::errorMessages)
+        .findFirst()
+        .orElse(Collections.emptyList());
+
+    assertTrue(pkModeError.isEmpty());
```
C0urante, line 48: Nit: I prefer to do a comparison for tests like this since it gives a more informative message if the assertion fails:
```suggestion
    assertEquals(Optional.empty(), pkModeError);
```

src/test/java/io/confluent/connect/jdbc/JdbcSinkConnectorTest.java
```diff
@@ -0,0 +1,70 @@
...
+    Config config = connector.validate(connConfig);
+
+    List<String> pkModeError = config.configValues()
+        .stream()
+        .filter(cfg -> PK_MODE.equals(cfg.name()))
+        .map(ConfigValue::errorMessages)
+        .findFirst()
+        .orElse(Collections.emptyList());
+
+    assertTrue(pkModeError.isEmpty());
```
C0urante, line 68: (Same nit RE: comparison for assertion)
```suggestion
    assertEquals(Optional.empty(), pkModeError);
```

src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
```diff
@@ -65,8 +67,20 @@ public ConfigDef config() {
...
+
+    config.configValues()
+        .stream()
+        /** use recommended values for configuration validation
+         * {@link PrimaryKeyModeRecommender#validValues}
+         */
+        .filter(cfg -> PK_MODE.equals(cfg.name())
+            && !cfg.recommendedValues().contains(cfg.value()))
+        .findFirst()
+        .ifPresent(cfg -> cfg.addErrorMessage("'" + cfg.value() + "' is not valid"));
```
C0urante, line 81: This is a really interesting, lightweight approach that allows us to implement the cross-property logic for acceptable primary key modes in a single place and reuse it for both programmatic UIs (which leverage the "recommended values" for a property) and now preflight validation.

The one thing that makes me a little hesitant is that the error message we're giving to the user isn't very informative. We could say something like "please only use a recommended value for this property", but for users querying the REST API directly without some kind of programmatic UI in front, they might not even know how to retrieve that information (by hitting the `PUT /connector-plugins/{connectorType}/config/validate` endpoint), and even if they do, it'd probably be a little painful to have to make that extra request instead of being given the relevant information directly in the error message.

Do you think there's a way we can isolate the shared logic for validation and our implementation of the `Recommender` interface for this property in a way that allows us to define the relationship between the `pk.mode` and `delete.enabled` properties once, but also generate user-actionable error messages during validation? I haven't thought too much about how we might structure this but am happy to talk here or offline if you'd like to iterate on possibilities.

src/test/java/io/confluent/connect/jdbc/JdbcSinkConnectorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    JdbcSinkConnector connector = new JdbcSinkConnector();
+    Config config = connector.validate(connConfig);
+
+    List<String> pkModeError = config.configValues()
+        .stream()
+        .filter(cfg -> PK_MODE.equals(cfg.name()))
+        .map(ConfigValue::errorMessages)
+        .findFirst()
+        .orElse(Collections.emptyList());
```
C0urante, line 66: Nit: up to you, but it might be worth it to extract the logic for finding errors with a specific config property into a separate method so that we don't have to duplicate this logic.

src/test/java/io/confluent/connect/jdbc/JdbcSinkConnectorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    assertTrue(pkModeError.isEmpty());
+  }
+
+  @Test
+  public void testValidationWhenDeleteNotEnabled() {
+
+    Map<String, String> connConfig = new HashMap<>();
+    connConfig.put("connector.class", "io.confluent.connect.jdbc.JdbcSinkConnector");
+    connConfig.put("pk.mode", "none");
```
C0urante, line 56: Nit: may want to explicitly disable deletes here (makes the test a little easier to read for people who don't know the default off the top of their head, and makes it resilient to potential changes to that default)

```suggestion
    connConfig.put("pk.mode", "none");
    connConfig.put("delete.enabled", "false");
```

C0urante (commented):
Thanks @fzhong-connect. I've left a few comments; anything beginning with "nit: " can be implemented or ignored at your discretion. The only significant one has to do with providing clear error messages to users; let me know what you think.

Also, what's the reasoning for targeting 10.2.x with this change? Based on https://docs.confluent.io/platform/current/installation/versions-interoperability.html#cp-and-apache-ak-compatibility, we're still providing support as far back as 5.2.x; if it's not too difficult with merge conflicts and the like, could we consider targeting that branch instead?

src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
fzhong-connect: the error message is changed to ```
'pk.mode' can only have value in list [record_key]
```

> Do you think there's a way we can isolate the shared logic for validation and our implementation of the Recommender interface for this property in a way that allows us to define the relationship between the pk.mode and delete.enabled properties once, but also generate user-actionable error messages during validation? 

that is great idea. I am thinking maybe something like ```

interface CrossFieldsValidator extends ConfigDef.Recommender {
 // some method to report more sensible error message
}
```

I need give it more thought, but can we not block this PR before we can settle on a solution?

src/test/java/io/confluent/connect/jdbc/JdbcSinkConnectorTest.java
fzhong-connect: good point.

fzhong-connect:
> Thanks @fzhong-connect. I've left a few comments; anything beginning with "nit: " can be implemented or ignored at your discretion. The only significant one has to do with providing clear error messages to users; let me know what you think.
> 
> Also, what's the reasoning for targeting 10.2.x with this change? Based on https://docs.confluent.io/platform/current/installation/versions-interoperability.html#cp-and-apache-ak-compatibility, we're still providing support as far back as 5.2.x; if it's not too difficult with merge conflicts and the like, could we consider targeting that branch instead?

Thank you for taking time @C0urante .

The reason it's targeting 10.2.x is Nathan believes it mostly benefits user of the fully-managed connector on CCloud, he asked to start from the latest version on CC.

src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
fzhong-connect: the reason I want the error message to be as generic as possible is I'm thinking we may extend validation on other properties as well.

C0urante: The changes you made are definitely an improvement, but one thing that gives me pause is that we're still ultimately giving users a less-informative error message than the one that they currently see when there's an issue with the combination of `pk.mode`/`delete.mode` that they've used: https://github.com/confluentinc/kafka-connect-jdbc/blob/9811eec6a876ff3541cd415d83672cd527094abf/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java#L333

This message provides all of the information that would be given in the current PR, plus it explains the relationship between the `pk.mode` property and the `delete.enabled` property. With the current approach, users might see that the only `pk.mode` that they could configure the connector with is `record_key`, but they wouldn't know why, or that they could also disable deletes by setting `delete.enabled` to `false` instead of changing the `pk.mode`.

I don't know if there really is an easy way to codify these types of relationships in a generic, reusable fashion. So then the question becomes one of whether it's worth it to retain clear error messages at the cost of higher complexity in the code base. I think I'd prefer clearer error messages but would be interested in thoughts from you and others.

FWIW, we added preflight validation for the BigQuery connector recently and you can see an example of the pattern we used for that (although nothing included leveraging recommended values for validation) with these snippets:
- https://github.com/confluentinc/kafka-connect-bigquery/blob/b090f8566ce0cae74795b49337f6d804747e2af7/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkConnector.java#L55-L65
- https://github.com/confluentinc/kafka-connect-bigquery/blob/b090f8566ce0cae74795b49337f6d804747e2af7/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java#L737-L776
- https://github.com/confluentinc/kafka-connect-bigquery/blob/b090f8566ce0cae74795b49337f6d804747e2af7/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/MultiPropertyValidator.java
- https://github.com/confluentinc/kafka-connect-bigquery/blob/b090f8566ce0cae74795b49337f6d804747e2af7/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/PartitioningTypeValidator.java

(Funnily enough, we probably should have actually added `Recommender` logic for this exact example since it's enforcing a very similar relationship to the one in this PR: if your connector config meets some specific condition (in that case, automatic table creation is enabled and the connector should write to tables using BigQuery decorator syntax), then the set of permitted values for a specific property (in this case, the `timePartitioningType`) are restricted, and if that condition is not met, then the set of permitted values is not restricted. But the end result that all values are recommended at the moment, but you get a clear error message when you use one that is not permitted with the current connector config indicating what's wrong, why, and what to change, still provides a decent UX overall.)

fzhong-connect: > The changes you made are definitely an improvement, but one thing that gives me pause is that we're still ultimately giving users a less-informative error message than the one that they currently see when there's an issue with the combination of `pk.mode`/`delete.mode` that they've used:
> 
> https://github.com/confluentinc/kafka-connect-jdbc/blob/9811eec6a876ff3541cd415d83672cd527094abf/src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java#L333

Thank you for the link, I missed existing error message for this particular error.

fzhong-connect: > I don't know if there really is an easy way to codify these types of relationships in a generic, reusable fashion. So then the question becomes one of whether it's worth it to retain clear error messages at the cost of higher complexity in the code base. I think I'd prefer clearer error messages but would be interested in thoughts from you and others.

I agree, there is some balancing need be kept. 

General speaking, I prefer simple/reusable code logic even the cost is less clear error messages, which I believe can be made up by better documentation and run-books. 

I believe your idea is sound to combine recommendation and cross-fields validation into an reusable abstraction. We can explore it further for sure.

Thank you again for spending time and giving a lot of thought on this PR.

src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
```diff
@@ -65,8 +67,25 @@ public ConfigDef config() {
...
+        .findFirst()
+        /**
+         * use recommended values for configuration validation
+         * {@link PrimaryKeyModeRecommender#validValues}
+         */
+        .ifPresent(pkMode -> {
+          if (!pkMode.recommendedValues().contains(pkMode.value())) {
+            pkMode.addErrorMessage("'" + pkMode.value() + "' is invalid:"
+                + " Deletes are only supported for pk.mode 'record_key'");
+          }
```
C0urante, line 85: Is this susceptible to false positives if the user gives a value for the `pk.mode` property that is invalid not because of conflict with `delete.enabled` but just because it doesn't correspond to a recognized primary key mode? For example, something like `pk.mode = not_a_real_mode`.

fzhong-connect: that is great catch! thank you.

src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
```diff
@@ -65,8 +70,38 @@ public ConfigDef config() {
...
+                .map(deleteEnabled -> (Boolean) deleteEnabled.value())
+                .orElse(Boolean.FALSE)) {
+              error += ": Deletes are only supported for pk.mode 'record_key'";
+            }
+
+            pkMode.addErrorMessage(error);
+          }
+        });
+
+    return config;
```
C0urante, line 97: I'm sorry, but I'm still not convinced that we need to use the recommended values list here. It looks like we originally used that approach because it seemed reusable and could be leveraged in the future with little effort, but we've now added in special case logic just for the `pk.mode`/`delete.enabled` combination that basically renders that check redundant. Worse, it actually double-dings users by generating the same error message twice if they give a completely nonsensical primary key mode like `this_mode_does_not_exist` since they should already see a validation error from the [EnumValidator](https://github.com/confluentinc/kafka-connect-jdbc/blob/86d9377bc654603d88c1006970aa5dfa2630d434/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java#L587-L592) that we [use for the `pk.mode` property](https://github.com/confluentinc/kafka-connect-jdbc/blob/86d9377bc654603d88c1006970aa5dfa2630d434/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java#L397).

I don't think we should try to optimize for reusability here if we're already adding special case logic (which is, IMO, warranted in order to preserve clear error messages to users who many be just getting started with this connector).

I think we should either add a more sophisticated approach that shares the logic for valid `pk.mode`/`delete.enabled` combinations in a single place for both validation and recommended values, or acknowledge that we're only implementing preflight validation for one case right now and not try to extend the changes we make here beyond that single case.

fzhong-connect: @C0urante thank you.

I agree. the whole thing now is very `pk.mode` focused, the half-way reusing `recommendedValues` becomes moot.

src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
```diff
@@ -73,34 +75,29 @@ public Config validate(Map<String, String> connectorConfigs) {
...
-        });
+    return validateDeleteEnabledPkMode(config);
+  }
 
+  private Config validateDeleteEnabledPkMode(Config config) {
+    configValue(config, DELETE_ENABLED)
+        .filter(deleteEnabled -> Boolean.TRUE.equals(deleteEnabled.value()))
+        .ifPresent(deleteEnabled -> configValue(config, PK_MODE)
+            .ifPresent(pkMode -> {
+              if (!RECORD_KEY.name().toLowerCase(Locale.ROOT).equals(pkMode.value())) {
```
C0urante, line 86: It looks like we allow the `pk.mode` to be either uppercase or lowercase: https://github.com/confluentinc/kafka-connect-jdbc/blob/3de78db9bfada36d575dc82c23aa8abe356fcda3/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java#L581-L582

Would this report an error to the user if they have deletes enabled but specify a `pk.mode` of `RECORD_KEY` instead of `record_key`?

C0urante (approved):
LGTM, thanks Frank!

src/main/java/io/confluent/connect/jdbc/JdbcSinkConnector.java
fzhong-connect: that is quick. thank you!

you are right. should accept both RECORD_KEY and record_key.

PR #1159, opened 2022-01-21
CCDB-4340 Fix the NPE on timestamp columns that are null

ddasarathan:
## Problem
For timestamp columns that are optional, the values appear as `null`. When the `timestamp.granularity` config is selected to be `nanos_long` or `nanos_string` or `nanos_iso_datetime_string` this is currently not handled properly and results in NPE

## Solution
This is now handled by the connector 

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


ncliang (approved):
LGTM. Thanks @ddasarathan

ddasarathan:
Thanks Nigel

PR #1169, opened 2022-02-15
CCDB-4586: Request to add a `Prefix` or a `Snapshot` mode for JDBC Source Connector to execute the query in snapshot mode

SajanaW:

## Problem
Customers want the ability to set transaction isolation for queries.  Ansi-sql supports `READ_UNCOMMITTED `, `READ_COMMITTED`, `REPEATABLE_READ`, `SERIALIZABLE`.  We want to set the infrastructure to support custom db specific isolation modes as well. For version 1, we're supporting `SQL_SERVER_SNAPSHOT_ISOLATION` for Sql Server as well.

## Solution
Provide a config to set an isolation mode. We implement the isolation mode setting in the dialect classes to allow specific dbs to override the generic implementation and set custom isolation modes.

Note that we still gate keep which isolation modes are supported by limiting the `TransactionIsolationMode` enum (internal to the config class). However we are able to expand on this model and set isolation modes for more dbs. 

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -633,23 +633,6 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     }
   }
 
-  protected static int mapToAnsiSqlTransactionIsolationMode(
-          JdbcSourceConnectorConfig.TransactionIsolationMode isolationMode
-  ) {
-    switch (isolationMode) {
-      case READ_UNCOMMITTED:
-        return Connection.TRANSACTION_READ_UNCOMMITTED;
-      case READ_COMMITTED:
-        return Connection.TRANSACTION_READ_COMMITTED;
-      case REPEATABLE_READ:
-        return Connection.TRANSACTION_REPEATABLE_READ;
-      case SERIALIZABLE:
-        return Connection.TRANSACTION_SERIALIZABLE;
-      default:
-        return -1;
-    }
-  }
-
   protected String displayableTableTypes(String[] types, String delim) {
     return Arrays.stream(types).sorted().collect(Collectors.joining(delim));
   }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -606,16 +606,18 @@ public class GenericDatabaseDialect implements DatabaseDialect {
           Connection connection,
           TransactionIsolationMode transactionIsolationMode
   ) {
-    int isolationMode =
-            mapToAnsiSqlTransactionIsolationMode(
-                    transactionIsolationMode
-            );
-    if (isolationMode == -1) {
-      // not an ANSI-SQL isolation mode.
-      log.warn("Unable to set transaction.isolation.mode: "
-              +  transactionIsolationMode.name()
-              +  ". This mode is not supported by the database."
-              + "No transaction isolation mode will be set for the queries");
+    if (transactionIsolationMode
+            == TransactionIsolationMode.DEFAULT) {
+      return;
+    }
+    int isolationMode = TransactionIsolationMode.get(
+            transactionIsolationMode
+    );
+    if (transactionIsolationMode
+            == TransactionIsolationMode.SQL_SERVER_SNAPSHOT_ISOLATION
+            && !name().equals(SqlServerDatabaseDialect.class.getSimpleName())) {
+      log.warn("Unable to set transaction.isolation.mode: " +  transactionIsolationMode.name()
+              +  ". This is only valid for a Connector with Sql Server dialect. ");
       return;
     }
     try {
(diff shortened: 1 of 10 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -872,6 +905,11 @@ public static TimestampGranularity get(JdbcSourceConnectorConfig config) {
     }
   }
 
+  public enum TransactionIsolationMode {
+    DEFAULT, READ_UNCOMMITTED, READ_COMMITTED,
```
ncliang, line 909: Also, consider implementing this similar to how `numeric.mapping` is implemented so that you don't need to keep doing `TransactionIsolationMode.valueOf` to go from the specified value back to the enum.

ncliang (commented):
Thanks @SajanaW . Few comments and questions.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
SajanaW: Pushing back on this because after the changes that were suggested was added, we only call `valueOf` once and the enum extension seems like overkill for it

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -872,6 +905,11 @@ public static TimestampGranularity get(JdbcSourceConnectorConfig config) {
     }
   }
 
+  public enum TransactionIsolationMode {
```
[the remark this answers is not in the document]
SajanaW: I implemented a multi-config validator to check this. It doesn't just work OOTB via an enum validator since we won't have access to all the config.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -330,6 +336,29 @@
...
 
+  public static final String TRANSACTION_ISOLATION_MODE_DEFAULT =
+          TransactionIsolationMode.DEFAULT.name();
+  public static final String TRANSACTION_ISOLATION_MODE_CONFIG = "transaction.isolation.mode";
+  private static final String TRANSACTION_ISOLATION_MODE_DOC =
+          "Mode to control which transaction isolation level is used when running queries. "
+                  + "against the database. By default no explicit transaction isolation"
+                  + "mode is set. SQL_SERVER_SNAPSHOT_ISOLATION_MODE will only work"
+                  + "against a connector configured to write to Sql Server. "
+                  + "The Microsoft Sql Server driver is a requirement to use this feature. "
```
ncliang, line 347: I don't think we need to explicitly state anything about the SQL server driver requirement. Those using the connector against non-SQL server database would not expect `SQL_SERVER_SNAPSHOT_ISOLATION_MODE` to work and would not be required to have the SQL server driver to use any other isolation modes.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -599,6 +602,37 @@ public boolean tableExists(
...
+            == TransactionIsolationMode.DEFAULT) {
+      return;
+    }
+    int isolationMode = TransactionIsolationMode.get(
+            transactionIsolationMode
+    );
+    if (transactionIsolationMode
+            == TransactionIsolationMode.SQL_SERVER_SNAPSHOT_ISOLATION
+            && !name().equals(SqlServerDatabaseDialect.class.getSimpleName())) {
+      log.warn("Unable to set transaction.isolation.mode: " +  transactionIsolationMode.name()
```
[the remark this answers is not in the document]
SajanaW: Yeah I think that's fairly unlikely. I think I added this earlier when I was opting for a runtime check than a validation. Removing it

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -330,6 +336,28 @@
...
+                  + "against the database. By default no explicit transaction isolation"
+                  + "mode is set. SQL_SERVER_SNAPSHOT_ISOLATION_MODE will only work"
+                  + "against a connector configured to write to Sql Server. "
+                  + " Options include:\n"
+                  + "  * DEFAULT\n "
+                  + "  * READ_UNCOMMITED\n"
+                  + "  * READ_COMMITED\n"
+                  + "  * REPEATABLE_READ\n"
+                  + "  * SERIALIZABLE\n"
+                  + "  * SQL_SERVER_SNAPSHOT_ISOLATION_MODE\n";
```
ncliang, line 353: nit: maybe rename this to `SQL_SERVER_SNAPSHOT` to match other isolation modes that do not have the isolation mode postfix

ncliang (approved):
Thanks @SajanaW . LGTM with minor nit.

ncliang (approved):
LGTM. Thanks @SajanaW

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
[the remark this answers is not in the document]
SajanaW: Yeah there's a commit that did just that.

PR #1176, opened 2022-03-03
Remove the assertion of executeUpdate count and check for EXECUTE_FAILED

mageshn:
## Problem

Some DB return a mix if SUCCES_NO_INFO and actual count resulting in the assertion failing. 

## Solution
Remove the code that does the assertion and also fail if any of the responses were EXECUTE_FAILED to ensure we fail. The driver spec is throw a `BatchUpdateException` but we are additionally checking for `EXECUTE_FAILED` in the response 

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -219,6 +220,9 @@ public BufferedRecords(
   private Optional<Long> executeUpdates() throws SQLException {
     Optional<Long> count = Optional.empty();
     for (int updateCount : updatePreparedStatement.executeBatch()) {
+      if (updateCount == Statement.EXECUTE_FAILED) {
```
gharris1727, line 223: I assume that if this error code is returned that there is no further context (like as an exception) to be retrieved from the driver. I don't see any relevant details to include in the error message nearby, except for the number of records or records themselves, so this might already be the best error message possible.

gharris1727 (commented):
I would have preferred moving the condition itself into the dialect instead of adding a boolean flag, but I can't find precedence for either strategy in the dialect. Moving the condition requires too much context (the whole list of records, for instance) and would require baking all of this code into the dialect which is not really viable.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
mageshn: That's right. I will explore a bit more to see if there is anything that the driver could give back. The driver might just throw an exception in many cases or return this error code. We were not handling this scenario at all and we were  relied on the count reconcile to throw an exception. If we don't do a reconcile, we need this explicit check.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -191,8 +191,9 @@ public BufferedRecords(
     log.trace("{} records:{} resulting in totalUpdateCount:{} totalDeleteCount:{}",
         config.insertMode, records.size(), totalUpdateCount, totalDeleteCount
     );
-    if (totalUpdateCount.filter(total -> total != expectedCount).isPresent()
-        && config.insertMode == INSERT) {
+    if (dbDialect.assertUpdateCount()
+            && totalUpdateCount.filter(total -> total != expectedCount).isPresent()
+            && config.insertMode == INSERT) {
```
ncliang, line 196: I think I agree with @gharris1727 that I would rather we move this logic into the dialect and use inheritance to have dialects that should skip this check override. What if we change the `assertUpdateCount` signature to take as input the `totalUpdateCount` and `expectedCount` and maybe insert mode and just called into the dialect here
```
dbDialect.assertUpdateCount(expectedCount, totalUpdateCount, INSERT);
```

mageshn: I can't think of any different ay to check this equality by dialect. I see only two variants compare or don't compare. In reality the comparison is kind of irrelevant if we check EXECUTE_FAILED in the loop and fail.

mageshn:
@ncliang @gharris1727  Thoughts on what kind of abstraction we should have in the dialect? I don't think it is viable to move the complete `executeUpdate` as it would lead to duplication in other dialects and the actual execution itself doesn't change between dialects. The real difference is going to be on how we treat the response of `executeBatch`. Over there, the only diff is whether we reconcile the count or not. We don't need this if we decide not to do the reconcile .

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
ncliang: If the comparison is going to be irrelevant, I don't see why we need to expose a method on the dialect just for it. Maybe we should just remove this logic in favor of catching this earlier in `executeUpdates`

mageshn: I would definitely prefer that if we all agree that there is no value from the actual reconcile

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -232,6 +236,9 @@ private long executeDeletes() throws SQLException {
     long totalDeleteCount = 0;
```
ddasarathan, line 236: nit (same as above): signature needs to include throws SQLException, ConnectException

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
ddasarathan:
```
If one of the commands in a batch update fails to execute properly, this method throws a BatchUpdateException, and a JDBC driver may or may not continue to process the remaining commands in the batch. However, the driver's behavior must be consistent with a particular DBMS, either always continuing to process commands or never continuing to process commands. If the driver continues processing after a failure, the array returned by the method BatchUpdateException.getUpdateCounts will contain as many elements as there are commands in the batch, and at least one of the elements will be the following:
A value of EXECUTE_FAILED -- indicates that the command failed to execute successfully and occurs only if a driver continues to process commands after a command fails
```
Trying to understand the wording here. If one of the command fails, then the `executeBatch` method throws a `BatchUpdateException` and `BatchUpdateException.getUpdateCounts` returns an array of INT which contains `EXECUTE_FAILED` for the commands in batch that failed. 

So from my understanding this statement will never be true.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
ddasarathan: I like this approach better. It keeps the check as is for the existing dialects while allowing new dialects to bypass this (which I assume will happen for a few drivers)

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -232,6 +236,9 @@ private long executeDeletes() throws SQLException {
...
     if (nonNull(deletePreparedStatement)) {
       for (int updateCount : deletePreparedStatement.executeBatch()) {
+        if (updateCount == Statement.EXECUTE_FAILED) {
```
ddasarathan, line 239: same as described in `executeUpdates`

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
mageshn: maybe the right thing to do would be to get the array from `BatchUpdateException` if it fails or rely on the response from the actual method for success cases and for drivers that don't do this per the spec.

ddasarathan: By not doing anything here, I think `JdbcSinkTask` attempts to retry a few times and fails the task after retry expiry, logging the update counts and `EXECUTE_FAILED` values.

ddasarathan: I am wrong. I don't see the updateCounts explicitly included in the Exception message. It would be better to log it, rethrow the same exception and not throw a `ConnectException` to retain the existing retry behavior.

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
mageshn: The one thing we cold potentially do is to pass the `insertMode` into this and use it to expand this a bit further. and default impl would return true if config.insertMode == INSERT

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
mageshn: changed to throw BatchUpdateException

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -219,6 +220,9 @@ public BufferedRecords(
   private Optional<Long> executeUpdates() throws SQLException {
```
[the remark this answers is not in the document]
mageshn: changed to throw BatchUpdateException

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -218,7 +216,11 @@ public BufferedRecords(
    */
   private Optional<Long> executeUpdates() throws SQLException {
     Optional<Long> count = Optional.empty();
+
     for (int updateCount : updatePreparedStatement.executeBatch()) {
+      if (updateCount == Statement.EXECUTE_FAILED) {
+        throw new BatchUpdateException("Execution failed for part of the batch update", new int[]{Statement.EXECUTE_FAILED});
```
ddasarathan, line 222: can we just include `updatePreparedStatement.executeBatch()` for the updateCount array?

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -232,6 +234,9 @@ private long executeDeletes() throws SQLException {
     long totalDeleteCount = 0;
     if (nonNull(deletePreparedStatement)) {
       for (int updateCount : deletePreparedStatement.executeBatch()) {
+        if (updateCount == Statement.EXECUTE_FAILED) {
+          throw new BatchUpdateException("Execution failed for part of the batch delete", new int[]{Statement.EXECUTE_FAILED});
```
ddasarathan, line 238: same comment as above. Requesting to pass through `deletePreparedStatement.executeBatch()` instead of creating a new int array with `Statement.EXECUTE_FAILED`

ddasarathan (approved):
Requested a change in the `BatchUpdateException` constructor, but otherwise LGTM.

mageshn:
@ncliang can you please take a look as well? I want to make sure everyone in the thread is aligned on the abstraction.

ncliang (commented):
Thanks @mageshn . I think based on this discussion, I would prefer if we removed the count comparison check in `flush()` , and along with it the need to introduce a one-off method on the Dialect interface. Instead, we can rely on the checks in `executeUpdates` which were added.

https://github.com/confluentinc/kafka-connect-jdbc/pull/1176/files#r822198440

ddasarathan:
@ncliang @mageshn 
I think all of this discussion depends on how the driver implements the batch updates right. We now think it can,
- return an array of ints equal to the updated row for each command
- return an array of ints not equal to the updated row for each command (for some drivers)
- return an array of ints with some items containing an error code
- throw a BatchUpdateException which may or may not contain an array of ints with record counts / error codes
- throw any SqlException

IMO this is a big list, and its better to leave it specific for each driver. I am assuming the count comparison check is probably the only way to detect a failure for some driver implementation. So I'd prefer the current changes that @mageshn has and suggest not to remove the count comparison check.

mageshn:
@ddasarathan are you suggesting that it would be a safety net when driver's don't report EXECUTE_FAILED or throw a BatchUpdateException but return 0 for non updated records?

ncliang:
My only concern is the addition of the `assert` method on the dialect interface. From previous discussion or indeed from the PR implementation itself, it is not entirely obvious that this state needs to be exposed on the interface. If we can get by without exposing this method, I would prefer if we did not.

From the discussion, it sounds like we might be able to make changes to make changes to do the assertions in `executeUpdates` to take into account all the aforementioned possibilities, and then not have to rely on the conditional assert in `flush` . If this is possible, I would really prefer this approach and not needlessly complicate the dialect interface.

mageshn:
> My only concern is the addition of the `assert` method on the dialect interface. From previous discussion or indeed from the PR implementation itself, it is not entirely obvious that this state needs to be exposed on the interface. If we can get by without exposing this method, I would prefer if we did not.
> 
> From the discussion, it sounds like we might be able to make changes to make changes to do the assertions in `executeUpdates` to take into account all the aforementioned possibilities, and then not have to rely on the conditional assert in `flush` . If this is possible, I would really prefer this approach and not needlessly complicate the dialect interface.

We can remove it only if we can safely assume that all the driver would either return EXECUTE_FAILED or would throw an Exception in case of failures. I think it is a reasonable expectation. @ddasarathan Thoughts?

ncliang (approved):
Thanks @mageshn , LGTM.

PR #1178, opened 2022-03-10
Spawn Tasks even when there's no tables that match the include/exclude list.

SajanaW:

## Problem
If there's no table found in the TableMonitorThread, we currently just don't spawn any tasks. While this is reasonable behavior, it makes debugging these issues a little bit harder since we see a connector running but no tasks. 

## Solution
We should aim to instead spawn tasks regardless of the table compatibility but immediately fail them to represent this misconfiguration.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -154,7 +154,7 @@ public class JdbcSourceConnector extends SourceConnector {
                 + "the list of tables from the database yet"
         );
       } else if (currentTables.isEmpty()) {
-        taskConfigs = new ArrayList<>(1));
+        taskConfigs = new ArrayList<>(1);
         log.warn("No tables were found so there's no work to be done.");
         Map<String, String> taskProps = new HashMap<>(configProperties);
         taskProps.put(JdbcSourceTaskConfig.TABLES_CONFIG, "[]");
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -154,8 +154,11 @@ public class JdbcSourceConnector extends SourceConnector {
                 + "the list of tables from the database yet"
         );
       } else if (currentTables.isEmpty()) {
-        taskConfigs = Collections.emptyList();
-        log.warn("No tasks will be run because no tables were found");
+        taskConfigs = new ArrayList<>(1));
+        log.warn("No tables were found so there's no work to be done.");
+        Map<String, String> taskProps = new HashMap<>(configProperties);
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -92,6 +92,18 @@ public void start(Map<String, String> properties) {
       throw new ConnectException("Couldn't start JdbcSourceTask due to configuration error", e);
     }
 
+    List<String> tables = config.getList(JdbcSourceTaskConfig.TABLES_CONFIG);
+    String query = config.getString(JdbcSourceTaskConfig.QUERY_CONFIG);
+
+    if ((tables.isEmpty() && query.isEmpty())) {
+      throw new ConnectException("Task is being killed because it was not assigned a table nor a query to execute."
```
mageshn, line 99: Can we add details on potential scenarios when we would not find tables?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -92,6 +92,25 @@ public void start(Map<String, String> properties) {
...
+      throw new ConnectException("Task is being killed because"
+              + " it was not assigned a table nor a query to execute."
+              + " If run in table mode please make sure that the tables"
+              + " are exist on the database. If the table does exist on"
+              + " the database, we recommend using the fully qualified"
+              + " table name.");
+    }
+
+    if ((!tables.isEmpty() && !query.isEmpty())) {
+      throw new ConnectException("Invalid configuration: each JdbcSourceTask"
```
mageshn, line 108: The message looks a bit mis-leading

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -154,8 +154,11 @@ protected CachedConnectionProvider connectionProvider(int maxConnAttempts, long
                 + "the list of tables from the database yet"
         );
       } else if (currentTables.isEmpty()) {
-        taskConfigs = Collections.emptyList();
-        log.warn("No tasks will be run because no tables were found");
+        taskConfigs = new ArrayList<>(1);
```
mageshn, line 157: do we need a similar behavior in the currentTables == null code branch as well?

SajanaW: No I don't think so. If it's null it means that at the time configureTasks was called, the query was still blocking (even after @ncliang 's change).  We have to expect the query to conclude eventually and a request to reconfigure tasks to be called.

mageshn (approved):
LGTM

PR #1179, opened 2022-03-15
CCDB-4632: fix a race condition in table monitoring thraed startup

ncliang:
## Problem
In https://github.com/confluentinc/kafka-connect-jdbc/pull/1145 , `JdbcSourceConnector::taskConfigs` was made to be entirely non-blocking. This introduced a race condition between the table monitoring thread initialization and connector thread calling `JdbcSourceConnector::taskConfigs` . When the table monitoring thread is not yet done initializing, the call to `TableMonitoringThread::tables` will return null and cause `taskConfigs` to not generate any tasks. The expectation is that when the table monitoring thread is done with initialization, it will call `ConnectorContext::requestTaskReconfiguration` to request reconfiguration and trigger another call to `taskConfigs` .

While this works in the normal path, the reconfiguration is predicated on the connector being in [running](https://github.com/apache/kafka/blob/7795ca0fee650db181cd5e83c99d4cab15b8d37e/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L1455-L1458) state. In rare scenarios where the connector was paused before the table monitoring thread has finished its initial table read, the connector ends up in a state with a paused connector and no tasks. Since the connector was paused when the table monitoring thread finished initialization, the `reconfiguration` request was ignored. Furthermore, since the tables list has not changed, there will be no more attempts to reconfigure the connector to get it back to a good state on resume. The only way to get out of this state is to first resume the connector, then restart it.

## Solution
Re-introduce a small amount of blocking time to the `taskConfigs` request. This gives the table monitoring thread some time to finish the initial table load request. Introduce a new config parameter `table.monitoring.startup.polling.limit.ms` to control the maximum amount of time to wait on table monitoring thread initial read.

- [ ] yes
- [x] no

## Test Strategy
Existing unit test was updated. A new unit test was added.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Land in `5.3.x` and `pint merge` back to master.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
@@ -19,6 +19,9 @@ import io.confluent.connect.jdbc.dialect.DatabaseDialect;
 import io.confluent.connect.jdbc.util.ConnectionProvider;
 import io.confluent.connect.jdbc.util.QuoteMethod;
 import io.confluent.connect.jdbc.util.TableId;
+
+import org.apache.kafka.common.errors.TimeoutException;
+import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.connect.connector.ConnectorContext;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
@@ -46,26 +49,32 @@ public class TableMonitorThread extends Thread {
   private final ConnectionProvider connectionProvider;
   private final ConnectorContext context;
   private final CountDownLatch shutdownLatch;
+  private final long startupMs;
   private final long pollMs;
   private final Set<String> whitelist;
   private final Set<String> blacklist;
   private final AtomicReference<List<TableId>> tables;
+  private final Time time;
 
   public TableMonitorThread(DatabaseDialect dialect,
       ConnectionProvider connectionProvider,
       ConnectorContext context,
+      long startupMs,
       long pollMs,
       Set<String> whitelist,
-      Set<String> blacklist
+      Set<String> blacklist,
+      Time time
   ) {
     this.dialect = dialect;
     this.connectionProvider = connectionProvider;
     this.context = context;
     this.shutdownLatch = new CountDownLatch(1);
+    this.startupMs = startupMs;
     this.pollMs = pollMs;
     this.whitelist = whitelist;
     this.blacklist = blacklist;
     this.tables = new AtomicReference<>();
+    this.time = time;
   }
 
   @Override
@@ -98,6 +107,7 @@ public class TableMonitorThread extends Thread {
    *         successfully yet
    */
   public List<TableId> tables() {
+    awaitTablesReady(startupMs);
     List<TableId> tablesSnapshot = tables.get();
     if (tablesSnapshot == null) {
       return null;
@@ -145,6 +155,15 @@ public class TableMonitorThread extends Thread {
     return tablesSnapshot;
   }
 
+  private void awaitTablesReady(long timeoutMs) {
+    try {
+      time.waitObject(tables, () -> tables.get() != null, time.milliseconds() + timeoutMs);
+    } catch (InterruptedException | TimeoutException e) {
+      log.info("Timed out or interrupted while awaiting for tables being read.");
+      return;
+    }
+  }
+
   public void shutdown() {
     log.info("Shutting down thread monitoring tables.");
     shutdownLatch.countDown();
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -46,26 +49,32 @@
...
       Set<String> whitelist,
-      Set<String> blacklist
+      Set<String> blacklist,
+      Time time
   ) {
     this.dialect = dialect;
     this.connectionProvider = connectionProvider;
     this.context = context;
     this.shutdownLatch = new CountDownLatch(1);
+    this.startupMs = startupMs;
```
mageshn, line 72: can we rely on existing pollms config for this?

SajanaW:
> Furthermore, since the tables list has not changed, there will be no more attempts to reconfigure the connector to get it back to a good state on resume. The only way to get out of this state is to first resume the connector, then restart it. 

So the race on the side of resume looks like this:

1. REST end point resume is hit
2. connector.start(config) is called by the herder 
3. start will spawn object `TableMonitorThread` and start it's run method without blocking
4. [we try to set the connector's state to running on the herder](https://github.com/confluentinc/ce-kafka/blob/master/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerConnector.java#L186)

The only chance of a race here is between 3 and 4, the the table monitor thread makes a network call, identifies the tables on the Db, and requests a task reconfiguration.  The task reconfiguration actually runs with [retries](https://github.com/confluentinc/ce-kafka/blob/master/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L1520-L1545). Further more, isn't the herder single threaded?  So shouldn't the [doTransitionTo](https://github.com/confluentinc/ce-kafka/blob/master/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerConnector.java#L343-L355) block?

All in all, I'm skeptical that this race condition could happen.



Therefore, I'm skeptical whether a block with retries can solve this problem? Or even if it did - whether this is indeed the root cause of the issues we see 

 -> 
start will

mageshn:
Should we start with a fixed block instead of a config to begin with?

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -145,6 +155,15 @@ public void run() {
     return tablesSnapshot;
   }
 
+  private void awaitTablesReady(long timeoutMs) {
+    try {
+      time.waitObject(tables, () -> tables.get() != null, time.milliseconds() + timeoutMs);
+    } catch (InterruptedException | TimeoutException e) {
+      log.info("Timed out or interrupted while awaiting for tables being read.");
```
ddasarathan, line 162: Can we make this a warn?

src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
mageshn: or use a constant for now before we see compelling reasons to expose a configuration

ncliang: I made a conscious decision to separate the two as the time spent here waiting on startup is blocking the herder thread. The polling interval is specified by `table.poll.interval.ms` , which defaults to a minute. I don't think many users would change this. And, I did not want to be blocking the herder thread for up to a minute in the default config. Conceptually too, these are very different time intervals, so it makes sense you might want to adjust them separately.

ncliang:
Thanks for review @mageshn , @SajanaW , and @ddasarathan 

@SajanaW . As you say, the race can occur between the table monitor thread starting and the herder thread transitioning the connector to RUNNING state. The missing piece in your analysis above I think is that the reconfiguration logic while it does retry on error, will not attempt the retry on a [non-running](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L1552) connector. We have verified this by finding these log lines in Kibana for the affected connector.

> Should we start with a fixed block instead of a config to begin with?

@mageshn any reason to use a hardcoded interval instead of user configurable one? I can see the aversion to exposing yet another user config, but wanted to see if you had concerns other than further complicating the configuration.

mageshn:
My only concern is complicating the user configuration further.

ncliang:
Thanks for review @mageshn . I have made the config internal to avoid further complicating the user configuration.

mageshn (approved):
LGTM

ncliang:
Jenkins run is green, merging
https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/kafka-connect-jdbc/view/change-requests/job/PR-1179/

PR #1180, opened 2022-03-16
CCDB-4450:Limit Retries on SQLExceptions on JdbcSource 

SajanaW:

## Problem
SQL Exceptions are retried indefinitely

## Solution
Retry SQL Exceptions a fix number of times

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -305,6 +305,12 @@
       "Suffix to append at the end of the generated query.";
   public static final String QUERY_SUFFIX_DISPLAY = "Query suffix";
 
+  public static final String QUERY_RETRIES_CONFIG = "query.retry.attempts";
```
mageshn, line 308: @NathanNam thoughts on the config name?

mageshn:
LGTM in general. Can we add a test with mock?

mageshn:
Should we target this against master?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -70,6 +70,9 @@
   private final AtomicBoolean running = new AtomicBoolean(false);
   private final AtomicLong taskThreadId = new AtomicLong(0);
 
+  private Map<TableQuerier, Integer> retriesAttemptedPerTableQuerier;
```
gharris1727, line 73: I see that we're using a similar strategy for the counting the consecutiveEmptyResults, but that's a local map only used within a single poll.

WDYT about making this a mutable state field inside the TableQuerier instead of a map? This sort of "external member field" seems to break some OOP rules.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskLifecycleTest.java
```diff
@@ -308,6 +319,66 @@ public void testNonTransientSQLExceptionThrows() throws Exception {
...
+    Field cachedConnectionProvider = mockedTask.getClass().getDeclaredField("cachedConnectionProvider");
+    Field retriesAttemptedPerTableQuerier = mockedTask.getClass().getDeclaredField("retriesAttemptedPerTableQuerier");
+    Field maxRetriesPerQuerier = mockedTask.getClass().getDeclaredField("maxRetriesPerQuerier");
+
+
+    tableQueue.setAccessible(true);
+    running.setAccessible(true);
+    cachedConnectionProvider.setAccessible(true);
+    retriesAttemptedPerTableQuerier.setAccessible(true);
+    maxRetriesPerQuerier.setAccessible(true);
```
gharris1727, line 371: Is there a way to test this not involving reflection?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -409,6 +416,7 @@ protected void closeResources() {
           }
         } else {
           consecutiveEmptyResults.put(querier, 0);
+          retriesAttemptedPerTableQuerier.put(querier, 0);
```
mageshn, line 419: Why are we resetting only if there are results? Even no results is a success, right?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
SajanaW: yeah sounds good to me.

mageshn: I don't think you need that if the state is mutable via a method that signals the exception and asks the querier to  increment the retry counter. @gharris1727 is that what you had in mind?

SajanaW:
Hey folks,

I took a stab at the comments left. I tried hard to use an IT instead of mocks - however most of the common JDBC configuration issues are surfaced as `NonTransientSqlExceptions`. Only connection related issues are packaged as `SqlExceptions`.  However these are thrown as ConnectExceptions from the Connection Provider. 


Instead of the reflections, I chose to use package-private/ package-local so that I can replace the querier with a mock and the connection with a mock! 


Sorry about the delay here

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -421,6 +425,7 @@ protected void closeResources() {
 
         if (results.isEmpty()) {
           consecutiveEmptyResults.compute(querier, (k, v) -> v + 1);
+          querier.resetRetryCount();
```
ncliang, line 428: Rather than reseting the retry count in both if and else clauses, does it make sense to reset the count before entering this if block?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -447,6 +453,12 @@ protected void closeResources() {
       } catch (SQLException sqle) {
         log.error("SQL exception while running query for table: {}", querier, sqle);
```
ncliang, line 454: Maybe add something in logging to indicate we are doing attempt x of y retries.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskLifecycleTest.java
```diff
@@ -26,22 +26,26 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.PriorityQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.confluent.connect.jdbc.util.CachedConnectionProvider;
 
+import static org.easymock.EasyMock.*;
```
ncliang, line 40: Please don't use star imports.

ncliang (commented):
Thanks @SajanaW . Mostly LGTM with a few nits

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -66,11 +66,13 @@
   private Time time;
   private JdbcSourceTaskConfig config;
   private DatabaseDialect dialect;
-  private CachedConnectionProvider cachedConnectionProvider;
-  private PriorityQueue<TableQuerier> tableQueue = new PriorityQueue<TableQuerier>();
+  CachedConnectionProvider cachedConnectionProvider;
```
ncliang, line 69: nit: add a comment stating these are made `//VisibleForTesting`

ncliang (approved):
Thanks @SajanaW . LGTM

ncliang:
Build is green. Merging
https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/kafka-connect-jdbc/view/change-requests/job/PR-1180/

PR #1195, opened 2022-05-26
CCDB-4777: Change JDBC auto.create for STRING types from CLOB to VARC…

ManasjyotiSharma:
…HAR.

## Problem
For Oracle JDBC Connector, String fields were getting mapped to CLOB. This was causing performance degradation.

## Solution
For OracleDialect of JDBC Connector, mapping String fields to VARCHAR2(4000). If this cannot handle some specific/corner case, then the workaround is that end user has to upfront create the table/column manually with the needed data type.

- [ ] yes
- [x ] no

##### Testing done:
- [x ] Unit tests
- [x ] Integration tests
- [ ] System tests
- [x ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -200,7 +200,7 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
       case BOOLEAN:
         return "NUMBER(1,0)";
       case STRING:
-        return "CLOB";
+        return "VARCHAR2(4000)";
       case BYTES:
         return "BLOB";
       default:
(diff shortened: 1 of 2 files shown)


sagarrao12:
Thanks @ManasjyotiSharma . I am assuming this PR is applicable for tables created via the connector. I would suggest writing an IT as well to check the datatype of the table created(if that's possible in this codebase)

ManasjyotiSharma:
> Thanks @ManasjyotiSharma . I am assuming this PR is applicable for tables created via the connector. I would suggest writing an IT as well to check the datatype of the table created(if that's possible in this codebase)

Thanks @sagarrao12 for taking a look at the PR and sharing your feedback. As discussed with you and the team during today's stand-up, for the current issue and code changes, the UTs seem to suffice and cover the cases. Please let me know if you have any other comments/feedbacks. Thanks!

sagarrao12 (approved):
LGTM

PR #1214, opened 2022-07-07
Add Redshift dialect implementation

jirufik:
## Problem
The current implementation of kafka-connect-jdbc does not properly support Redshift. Adding multiple columns does not work. Redshift converts the TEXT type to VARCHAR(256) by default. For many data, this is a short length. Converts the date from Debezium to TEXT.

## Solution
Creating a dialect for Redshift based on the modified PostgreSQL dialect.

- [x] yes
- [ ] no

##### If yes, where?
In the datamesh functionality of the EBAC organization https://ebaconline.com.br/

## Test Strategy
Added several modules to check the Redshift language. Manual testing was carried out on the EBAC developer platform.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ derby.log
 __test_database_*
 test.sqlite
 *.db-journal
+
+kafka-connect-jdbc.jar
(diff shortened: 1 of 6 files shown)


sagarrao12:
Thanks for the review @jirufik .. At this point we aren't looking at new dialects. @NathanNam , can you plz confirm if this something we want to add? Thanks!

Issue #1198, opened 2022-05-31
JDBC Source Connector connection.url error


dipdeb:
Hi
I have installed the confluent local version - 7.1.1. After installing the kafka-connect-jdbc getting this error
`An unexpected error occurred: org.apache.kafka.common.config.ConfigException: Missing required configuration "connection.url" which has no default value.`
on clicking the JdbcSourceConnector tile on the control center.

How to solve this problem?

OneCricketeer:
According to the error, you didn't supply the listed value, which is required

dipdeb:
@OneCricketeer thanks for the idea. It's working now. The problem is in the latest version of kafka-connect-jdbc (10.5.0). I had installed kafka-connect-jdbc couple months back in another machine and it is working, so I checked the version of jdbc connector in the old machine, using 
`curl localhost:8083/connector-plugins | jq`
and tried installing the same version in the new machine instead of using "latest" in 
`confluent-hub install --no-prompt confluentinc/kafka-connect-jdbc:10.3.3`

OneCricketeer:
Glad to hear it. Feel free to close the issue

OneCricketeer:
> exception occurs and the usual config screen is unable to display

But you still have the ability to add the url under Additional Properties. 

This issue itself is not with the JDBC connector. It should be brought up elsewhere for Control Center.

miramar-labs:
> But you still have the ability to add the url under Additional Properties.

but even if you do that, you've then lost all the other properties that need to be set ... the page doesn't 'reload' ... 

why not just add a default value of "" or null in JdbcSourceConnectorConfig.java .. around line 426 .. that would keep the UI happy I'm sure...

bjosloff:
Any update on this issue.  I'm see the same problem on CP 7.2.1 control center.  
Version 10.4.x and above are throwing this error. 
An unexpected error occurred: org.apache.kafka.common.config.ConfigException: Missing required configuration "connection.url" which has no default value.

miramar-labs:
while you wait for a resolution, a workaround is to toss the web GUI altogether and use scripts/REST calls to manage your connectors - see [here](https://docs.confluent.io/platform/current/connect/references/restapi.html#connectors)

mikementzmaersk:
Same issue here, with a variety of different kafka connect UI's (akhq, kafka-ui etc...) - as well as failures provisioning via api port 8083 directly.   10.3.3 is the only version this works on, all versions since have this issue.

PR #1222, opened 2022-07-29
isolate sql transaction mode in connector validation

alozano3:
## Problem
The PR https://github.com/confluentinc/kafka-connect-jdbc/pull/1169 introduced a validation that is provoking a failure in the [connector plugin validation endpoint](https://github.com/confluentinc/kafka-connect-jdbc/pull/1169). More specifically, if you call the validate endpoint for the JDBC Source Connector with only the _connector.class_ parameter, it returns the error below:

`An unexpected error occurred: org.apache.kafka.common.config.ConfigException: Missing required configuration "connection.url" which has no default value.`

The reason is the validation instantiates the class JdbcSourceConnectorConfig and additional validations are made. The key _**connection.url**_  has no default value, so it is failing when checking its value. For parameters that have default values is OK.

This is the same error mentioned in https://github.com/confluentinc/kafka-connect-jdbc/issues/1198

## Solution
Only instantiate the class JdbcSourceConnectorConfig when the transaction mode is SQL_SERVER_SNAPSHOT. The error will continue happening for that transaction mode, but at least not for the other configurations. I didn't want to modify the database dialects creation and modify the https://github.com/confluentinc/kafka-connect-jdbc/pull/1169 behaviour.

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -15,6 +15,9 @@
 
 package io.confluent.connect.jdbc;
 
+import static io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig.TRANSACTION_ISOLATION_MODE_CONFIG;
+
+import java.util.Optional;
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
@@ -27,7 +30,13 @@ import org.apache.kafka.connect.util.ConnectorUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import io.confluent.connect.jdbc.dialect.DatabaseDialect;
 import io.confluent.connect.jdbc.dialect.DatabaseDialects;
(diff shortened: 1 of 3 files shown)


alozano3:
I tested the code in PR https://github.com/confluentinc/kafka-connect-jdbc/pull/1227 to check if solved the problem. 

With that code, if you call the validate endpoint or directly the validate method without the parameter connection.url, he error is different because of the default value "". Now the error is ` _org.apache.kafka.connect.errors.ConnectException: Not a valid JDBC URL:_ `  when tries to create the dialect in the _validateMultiConfigs_ method.

aniketshrimal:
@alozano3 I want to check - Are you trying to fix https://github.com/confluentinc/kafka-connect-jdbc/issues/1198 with this PR? 

> I tested the code in PR https://github.com/confluentinc/kafka-connect-jdbc/pull/1227 without applying the code from this PR to check if solved the problem.
With that code, if you call the validate endpoint or directly the validate method without the parameter connection.url, he error is different because of the default value "". Now the error is _org.apache.kafka.connect.errors.ConnectException: Not a valid JDBC URL:_ when tries to create the dialect in the validateMultiConfigs method.

I have not tried the fix(https://github.com/confluentinc/kafka-connect-jdbc/pull/1227) with latest branch. Let me do the manual testing on CP and get back to you on this.

alozano3:
@aniketshrimal yes, I am trying to fix #1198 . If PR #1227 gets merged, I'll modify the code of this PR because the parameter would have a default value.
I put this in draft and I'll push again when the other PR is merged.

alozano3:
@aniketshrimal Yes you are right, the validation call will be successful. It will only fail if the `connection.url` is not sent and the `transaction.isolation.mode`  value is "SQL_SERVER_SNAPSHOT", but that is another issue. 

Thank you. Closing this PR

PR #1230, opened 2022-08-08
Added reload4j version

rajdangwal:
## Problem
The branches above 10.0.x don't get the reload4j version from the common parent.

## Solution
Manually fixed the version.
Will pint merge to further branches.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,7 @@
         <postgresql.version>42.3.2</postgresql.version>
         <jtds.driver.version>1.3.1</jtds.driver.version>
         <slf4j.version>1.7.36</slf4j.version>
+        <reload4j.version>1.2.19</reload4j.version>
         <licenses.name>Confluent Community License</licenses.name>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.package.home>target/${project.artifactId}-${project.version}-package</project.package.home>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -171,6 +171,7 @@
         <dependency>
             <groupId>ch.qos.reload4j</groupId>
             <artifactId>reload4j</artifactId>
+            <version>1.2.19</version>
```
aniketshrimal, line 174: Could we pin this version in properties section similar to how we have done for slf4j - https://github.com/confluentinc/kafka-connect-jdbc/blob/10.0.x/pom.xml#L64? 

Also is this dependency version not present in commons version?

aniketshrimal (approved):
LGTM @rajdangwal! Just one minor comment. Feel free to merge after u resolve that comment. Thanks!

pom.xml
rajdangwal: Thanks @aniketshrimal 
I have pinned the version as a property.

Yes, the common-parent (common-parent-6.0.0) does not define the reload4j.version property. Hence, we need to add the version separately.
The property was defined for common-parent of 5.4.x and 5.5.x branches. 
`<reload4j.version>1.2.19</reload4j.version>`

PR #1234, opened 2022-08-28
RCCA 7749 Use national character methods for unicode strings

rajdangwal:
## Problem
The JDBC sink connector when used with SqlServer, observes performance degradation due to passing of string query parameters in unicode format. 

## Solution
The recommended approach is to use `setStringParametersAsUnicode=false` when using non-unicode data types such as CHAR, VARCHAR, LONGVARCHAR JDBC types.
However, if unicode data types such as NCHAR, NVARCHAR, LONGNVARCHAR are used, the application should use national character methods such as setNString, setNCharacterStream etc. Otherwise, it could lead to data loss.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
@@ -19,12 +19,14 @@ import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
 import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Schema.Type;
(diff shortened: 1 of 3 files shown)


sagarrao12:
@rajdangwal , I looked at the PR and PR looks good. Just one clarifying question, the data loss that you are talking about, is it reproducible via tests? Other than that I am +1

rajdangwal:
Thank you @sagarrao12 
I have added an integration test to show that the unicode characters are stored correctly even when `setStringParametersAsUnicode` is set to `false`. This happens because we used the national character method (`setNString`) for unicode data types.

However, it is hard to show data loss through tests with my existing changes. But if we try to run above test without my changes, we can see that instead of unicode characters the DB will store question marks (`?`), and the assertion will fail the test case.

sagarrao12 (approved):
LGTM

PR #1239, opened 2022-09-13
CCDB-5069: Fix data loss in timestamp mode due to in-mem offsets

mukkachaitanya:
https://confluentinc.atlassian.net/browse/CCDB-5069

## Problem

#1093 provided a fix to avoid data loss, in Timestamp Query mode, when we have multiple rows with the same timestamp in the DB if the connector is shut down after reading at least one such row but before reading all of them. This fix however failed to capture the case in-memory offsets that are used when the connector faces a connectivity error and goes into a retry when the DB has a network partition.

Going back to the example used in #1093 -- if the connector's current timestamp is T0 and it then reads five rows from the current query with timestamps R1(T1), R2(T1), R3(T2), R4(T2), and R5(T3), the records it produces will have timestamps of T0, T1, T1, T2, and T3, respectively. In this case, if the connector sees a transient `SQLException` while seeking, records, say while trying to read R4(T2). We would've flushed an offset of `T1` but we store `T2` as the committed offset in memory. This leads to us missing R4(T2), as we start the next iteration from T3 hence reading R5(T3).

## Solution

The solution here was straightforward, we make sure to keep the offsets in-memory sync with the offsets we have flushed. Note, this would surely lead to data duplication as we re-pross the records having timestamp >T1, i.e. we see R3(T1) twice, but still, adhere to at-least-once semantics. All the other limitations discussed with timestamp mode still hold.

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

TODO: Add a test case to replicate the issue.

## Release Plan

 
Target this to `5.0.x` branch and forward merge to all the branches till master

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
@@ -128,9 +128,9 @@ public class TimestampTableQuerier extends TimestampIncrementingTableQuerier {
     if (nextRecord == null
         || canCommitTimestamp(currentRecord.timestamp(), nextRecord.timestamp())) {
       latestCommittableTimestamp = currentRecord.timestamp();
+      // set current in memory offset to the latestCommittableTimestamp
+      this.offset = new TimestampIncrementingOffset(latestCommittableTimestamp, null);
     }
-    // set current in memory offset to the latestCommittableTimestamp
-    this.offset = new TimestampIncrementingOffset(latestCommittableTimestamp, null);
     return currentRecord.record(offset);
   }
 
--- a/src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
@@ -129,6 +129,8 @@ public class TimestampTableQuerier extends TimestampIncrementingTableQuerier {
         || canCommitTimestamp(currentRecord.timestamp(), nextRecord.timestamp())) {
       latestCommittableTimestamp = currentRecord.timestamp();
     }
+    // set current in memory offset to the latestCommittableTimestamp
+    this.offset = new TimestampIncrementingOffset(latestCommittableTimestamp, null);
     return currentRecord.record(latestCommittableTimestamp);
   }
 
@@ -145,8 +147,13 @@ public class TimestampTableQuerier extends TimestampIncrementingTableQuerier {
         throw new DataException(e);
       }
     }
-    this.offset = criteria.extractValues(schemaMapping.schema(), record, offset);
-    Timestamp timestamp = offset.hasTimestampOffset() ? offset.getTimestampOffset() : null;
+    // Use the extracted timestamp as the record's timestamp
+    TimestampIncrementingOffset timestampOffset = criteria.extractValues(
+        schemaMapping.schema(),
+        record,
+        offset
+    );
+    Timestamp timestamp = timestampOffset.hasTimestampOffset() ? timestampOffset.getTimestampOffset() : null;
     return new PendingRecord(partition, timestamp, topic, record.schema(), record);
   }
 
--- a/src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
@@ -131,7 +131,7 @@ public class TimestampTableQuerier extends TimestampIncrementingTableQuerier {
     }
     // set current in memory offset to the latestCommittableTimestamp
     this.offset = new TimestampIncrementingOffset(latestCommittableTimestamp, null);
-    return currentRecord.record(latestCommittableTimestamp);
+    return currentRecord.record(offset);
   }
 
   private PendingRecord doExtractRecord() {
@@ -153,7 +153,9 @@ public class TimestampTableQuerier extends TimestampIncrementingTableQuerier {
         record,
         offset
     );
-    Timestamp timestamp = timestampOffset.hasTimestampOffset() ? timestampOffset.getTimestampOffset() : null;
+    Timestamp timestamp = timestampOffset.hasTimestampOffset()
+                          ? timestampOffset.getTimestampOffset()
+                          : null;
     return new PendingRecord(partition, timestamp, topic, record.schema(), record);
   }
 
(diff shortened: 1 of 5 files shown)


devinay:
Could we have a test for the infinite loop case:
- There are a large number of records with the same timestamp
- The timestamp is set to the current one
- The same rows are returned.

My concern is that with the `T1` stored, we' enter an infinite loop if there are a large number of records with just `T1`.

mukkachaitanya:
> Could we have a test for the infinite loop case:
> 
>     * There are a large number of records with the same timestamp
> 
>     * The timestamp is set to the current one
> 
>     * The same rows are returned.
> 
> 
> My concern is that with the `T1` stored, we' enter an infinite loop if there are a large number of records with just `T1`.

Thanks, for the comment @devinay! I just want to begin by saying that we aren't changing any existing functionality but rather fixing a bug/miss from the design approved as part of #1093. Do let me know if you feel that's not the case.

For the infinite loop concern; Let me know if this makes sense or if I am missing anything here. The only case this can happen is we have a lot of records with say T1 timestamp. In this case: 
- in an ideal scenario without any network issues/exceptions from the connector, we seek through and read all the records, and the last record with T1 timestamp has T1 offset and all the previous ones have T0 as the offset. Once that is committed we don't have any possibilities of an infinite loop with timestamp T1.
- other cases we have here are when we see a transient SQLException (note, for non-transient ones we fail connector). This PR fixes a bug for that exact case. If there are any such exceptions while processing data with timestamp T1, we enter a retry loop. 
	- With this PR, we restart our processing by starting a new query for records `>T0` and surely reprocess the data that has been already written to Kafka before causing duplicates. 
	- The only case we might go into an infinite loop here is if we continuously get errors on each iteration which IMO is very rare. This is also cause we retry infinitely. The newer versions of the connector have a parameter `query.retry.attempts` which makes sure we don't get into this infinite loop.
	
Therefore, IMO, here it feels like the tradeoffs are b/w providing at least once guarantee or potential data loss. That said, without this change, we surely have a bug in the connector i.e we are producing offset `T0` but storing `T1` in memory. 

I think we should rather document this case of duplicates and the fact that if we have continuous disconnects and a lot of the same timestamp data we can go into a reprocessing loop (again probability is very low here); hence users should also limit their retries. With the right configuration, we should be avoiding the issue for sure. Do let me know if there is still value for a test case here, and what kind of parameters are we looking the test to run agaisnt.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -1072,7 +1163,11 @@ private void assertTimestampOffsets(List<SourceRecord> records) {
     for(SourceRecord record : records) {
       Timestamp timestampValue = (Timestamp) ((Struct)record.value()).get("modified");
       Timestamp offsetValue = TimestampIncrementingOffset.fromMap(record.sourceOffset()).getTimestampOffset();
-      assertEquals(timestampValue, offsetValue);
+      assertTrue(
+          String.format("Invalid timestamp: {} and offset{} combination.", timestampValue,
```
ManasjyotiSharma, line 1167:
```minor suggestion
          String.format("Invalid timestamp {} and offset {} combination.", timestampValue,
```

src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
```diff
@@ -132,7 +132,9 @@ public SourceRecord extractRecord() {
         || canCommitTimestamp(currentRecord.timestamp(), nextRecord.timestamp())) {
       latestCommittableTimestamp = currentRecord.timestamp();
     }
-    return currentRecord.record(latestCommittableTimestamp);
+    // set current in memory offset to the latestCommittableTimestamp
+    this.offset = new TimestampIncrementingOffset(latestCommittableTimestamp, null);
```
ManasjyotiSharma, line 136: Shouldn't this line be sitting under the if block above i.e. just below the line where we mutate latestCommittableTimestamp?

mukkachaitanya: >  Existing one would work too but seems we are creating the same object again and again unnecessarily when latestCommittableTimestamp does not change.

I agree, this is to keep things as they were already. Before this change as well we were updating offset each time - https://github.com/confluentinc/kafka-connect-jdbc/pull/1239/files#diff-304c376e74a018800a82cae4e42d9e58c60d6bb753c492ee3ed7e5297ba18910L151-L152 

Both are schematically equivalent from what I understand. I don't have a strong inclination toward either. I can move this to the if block above as well.

src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
```diff
@@ -863,6 +866,81 @@ public void testCustomQueryWithTimestamp() throws Exception {
     PowerMock.verifyAll();
   }
 
+  @Test
+  public void testCustomQueryMultipleRecordsWithDBPartition() throws Exception {
```
mukkachaitanya, line 870: @ManasjyotiSharma - I see that this test doesn't actually cover the case we discuss in the PR description. I'll rework it to make sure it's the same/similar.

src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
ManasjyotiSharma: @mukkachaitanya Grouping together related muations under one block helps understand the intent when someone is reading the code. However I don't know why previously the code was creating offset each time; just with a cursory peek at the code I din't find a reason for the same. But I have not checked if offset is thread-safe, is concurrent update fine etc. Do we have enough safety nests in terms of UTs/ITs/SystemTests for this connector? Is so moving the line inside the if and running the tests should be quicker; otherwise I am fine to leave it as it as because the suggestion is a minor one and doesn't weigh against regressions should safety nets are not robust for this Connector.

mukkachaitanya: Done!

ManasjyotiSharma (approved):
Thanks for fixing this @mukkachaitanya! Interesting (and certainly non-trivial to debug) issue. LGTM!

mukkachaitanya:
I have rebased this to 10.0.x and awaiting the build. The change will be merged all the way to master. The rebase to 5.0.x has been un-successfully due to a lot of conflicts and some logical changes, I'll work on fixing it on older branches via a separate PR. cc @ManasjyotiSharma

PR #1241, opened 2022-09-16
support regex replacement for sink table naming

linfan:
## Problem

Currently this jdbc connect can only sink data to a table with fixed name or use the whole topic name as part of the table name, but in many cases, people may need a more flexible way to named their target tables.

For example, while using debezium source connect, the kafka topic would be named in a format of `<server-id>.<source-database>.<source-table>`, which is not suitable to be use as part of the sink table name.

## Solution

One of the most flexible way for user to specify a table naming rule is **regular expression replacement**, which mean take the kafka topic name and modify it via a regular expression before put it into the final sink table name.

In this PR, an expression of ${topic/**Pattern**/**Replacement**/} is supported for advanced sink table naming. This expression format is inspired by `sed` and `bash`.

- [x] yes
- [ ] no

##### If yes, where?

`bash` support a very similar variable value replacement format as ${var/Pattern/Replacement}, and `sed` support a more regular expression friendly replacement format as '/Pattern/Replacement/', while the char `/` can be substitute by any other symbol such as '#'.

## Test Strategy

Unit test case is included in this PR.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
@@ -30,6 +30,8 @@ import io.confluent.connect.jdbc.util.TableId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static io.confluent.connect.jdbc.util.NameUtils.toTableName;
+
 public class JdbcDbWriter {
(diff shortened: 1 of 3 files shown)


OneCricketeer:
Maybe I'm wrong, but RegexRouter transform is already able to accomplish this?

PR #1297, opened 2023-01-19
Adds config to conditionally mask SQL statements

b-goyal:
## Problem
Reverts changes done in [CCDB-5222](https://confluentinc.atlassian.net/browse/CCDB-5222)
Find more details on why this is being done [here] (https://confluent.slack.com/archives/C04JV6U2HU2)
- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 

[CCDB-5222]: https://confluentinc.atlassian.net/browse/CCDB-5222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -254,7 +254,7 @@ public class JdbcSinkConfig extends AbstractConfig {
       + "view definition does not match the records' schemas (regardless of ``"
       + AUTO_EVOLVE + "``).";
 
-  public static final String TRIM_SENSITIVE_LOG_ENABLED = "trim.sensitive.log.enabled";
+  public static final String TRIM_SENSITIVE_LOG_ENABLED = "trim.sensitive.log";
   private static final String TRIM_SENSITIVE_LOG_ENABLED_DEFAULT = "false";
   private static final EnumRecommender QUOTE_METHOD_RECOMMENDER =
       EnumRecommender.in(QuoteMethod.values());
@@ -498,7 +498,7 @@ public class JdbcSinkConfig extends AbstractConfig {
             TRIM_SENSITIVE_LOG_ENABLED,
             ConfigDef.Type.BOOLEAN,
             TRIM_SENSITIVE_LOG_ENABLED_DEFAULT,
-            ConfigDef.Importance.HIGH
+            ConfigDef.Importance.LOW
         );
 
   public final String connectorName;
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -254,6 +254,8 @@ public class JdbcSinkConfig extends AbstractConfig {
       + "view definition does not match the records' schemas (regardless of ``"
       + AUTO_EVOLVE + "``).";
 
+  public static final String TRIM_SENSITIVE_LOG_ENABLED = "trim.sensitive.log.enabled";
+  private static final String TRIM_SENSITIVE_LOG_ENABLED_DEFAULT = "false";
   private static final EnumRecommender QUOTE_METHOD_RECOMMENDER =
       EnumRecommender.in(QuoteMethod.values());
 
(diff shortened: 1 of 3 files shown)


mukkachaitanya (commented):
@b-goyal instead of reverting the PR. I prefer we add the code in a feature toggle (say, and internal config def for now), could be set to false by default. We can use this config to change the mode when necessary to add supportability.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -254,6 +254,8 @@ public enum PrimaryKeyMode {
       + "view definition does not match the records' schemas (regardless of ``"
       + AUTO_EVOLVE + "``).";
 
+  public static final String TRIM_SENSITIVE_LOG_ENABLED = "trim.sensitive.log.enabled";
```
mukkachaitanya, line 257:
```suggestion
  public static final String TRIM_SENSITIVE_LOG_ENABLED = "trim.sensitive.log";
```

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -491,6 +493,12 @@ public enum PrimaryKeyMode {
             2,
             ConfigDef.Width.SHORT,
             RETRY_BACKOFF_MS_DISPLAY
+        )
+        .defineInternal(
+            TRIM_SENSITIVE_LOG_ENABLED,
+            ConfigDef.Type.BOOLEAN,
+            TRIM_SENSITIVE_LOG_ENABLED_DEFAULT,
+            ConfigDef.Importance.HIGH
```
mukkachaitanya, line 501:
```suggestion
            ConfigDef.Importance.LOW
```

src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkTaskTest.java
```diff
@@ -474,7 +477,40 @@ void initWriter() {
...
+  public void testGetAllMessagesExceptionWithoutTrim() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+    JdbcSinkTask task = new JdbcSinkTask();
+    SQLException exception = new BatchUpdateException("Batch entry 0 INSERT INTO \"abc\" (\"c1\",\"c2\",\"c3\",\"c4\") " +
+            "VALUES ('1','2','3',NULL) was aborted: ERROR: null value in column \"c4\" violates not-null constraint\n" +
+            "  Detail: Failing row contains (1, 2, 3, null).  Call getNextException to see other errors in the batch.",
+            new int[0]);
+    String exceptedExceptionMessage = "Exception chain:" + System.lineSeparator() + exception + System.lineSeparator();
+    Method privateMethod = JdbcSinkTask.class.getDeclaredMethod("getAllMessagesException", SQLException.class);
+    task.shouldTrimSensitiveLogs = false;
+    privateMethod.setAccessible(true);
```
mukkachaitanya, line 490: @b-goyal - typically we prefer using mocks to test the exceptions. I would prefer we don't do reflections. Can you re-work the tests? I am gonna approve and merge to pull the feature in for now

mukkachaitanya (approved):
LG

src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkTaskTest.java
b-goyal: Mockito does not provide ways to test private methods, that's why I used reflections here.

b-goyal: Not using private method any more: https://github.com/confluentinc/kafka-connect-jdbc/pull/1299

ManasjyotiSharma:
@mukkachaitanya @b-goyal I see that the PR is already merged, nonetheless leaving following comments: 

(1) Default for TRIM_SENSITIVE_LOG_ENABLED_DEFAULT is set to "false". In other words we are saying we won't trim sensitive info by default unless the low level config is overridden to true via LD flag. Is this understanding correct? If so then would we not be opening up the flood gates again for simialr/original sev-1 issue of logging sensitive info? I guess we should rather do it the opposite i.e. by default set it to true i.e. trim sensitive data and only on a need basis to help debug specific issues for specific LCC; for a short window via LD overrides should disable it.

(2) Are we not following the log redactor path as discussed [here](https://confluent.slack.com/archives/C04JV6U2HU2/p1673952132068659?thread_ts=1673861664.988979&cid=C04JV6U2HU2)?

b-goyal:
Hey @ManasjyotiSharma, sorry for the confusion here. On cloud the plan is to keep it true by default. 
Had this discussion with @mukkachaitanya offline - 
_Yeah. In connector default the redaction to false, and on cloud we can have it true by default until we figure out a balance._
1) For cloud, this will be true
2) @mukkachaitanya had [this](https://github.com/confluentinc/kafka-connect-jdbc/pull/1297#pullrequestreview-1261673962) concern and we added config for toggling instead of reverting the PR.

PR #1303, opened 2023-01-29
Rename method releaseLocksQuietly to commitQuietly

rajdangwal:
## Problem
The method name was confusing. We are only calling the commit method inside releaseLocksQuietly. Releasing the acquired locks is a side-effect of commit call.

## Solution
Renamed the method to commitQuietly.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


ManasjyotiSharma (approved):
LGTM!

PR #1284, opened 2022-12-21
CCDB-5077: Fix closin CacheConnectionProvider to close on lifecycle events

mukkachaitanya:
## Problem

Fixes #1205. Follows up on #1206 to fix the issue on non life-cycle events
While CacheConenctionProvider is trying to get a connection, it is impossible to stop.

## Solution

Add a flag to not go on with connection attempt when a close request has been issued.

##### Does this solution apply anywhere else?
- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

merge to 5.0.x and further

PR #1308, opened 2023-02-06
[Minor] Fix upstream tests failing due to #1284

mukkachaitanya:
JdbcSourceTaskLifecycleTests were failing due to a change to API introduced as part of #1284

### DIFF
--- a/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskLifecycleTest.java
+++ b/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskLifecycleTest.java
@@ -78,7 +78,7 @@ public class JdbcSourceTaskLifecycleTest extends JdbcSourceTaskTestBase {
 
     // Should request a connection, then should close it on stop()
     EasyMock.expect(mockCachedConnectionProvider.getConnection()).andReturn(db.getConnection()).anyTimes();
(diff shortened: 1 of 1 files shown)


src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskLifecycleTest.java
```diff
@@ -78,7 +78,7 @@ protected CachedConnectionProvider connectionProvider(
 
     // Should request a connection, then should close it on stop()
     EasyMock.expect(mockCachedConnectionProvider.getConnection()).andReturn(db.getConnection()).anyTimes();
-    mockCachedConnectionProvider.close();
+    mockCachedConnectionProvider.close(true);
```
ManasjyotiSharma, line 81: @mukkachaitanya what does passing "true" mean?

mukkachaitanya: True implies that the connector is stopping and we want to close the ConnectionProvider entirely. False means that the connector is closing the connection when it faces some transient errors. This was introduced as part of https://github.com/confluentinc/kafka-connect-jdbc/pull/1284

ManasjyotiSharma (approved):
LGTM!

PR #1329, opened 2023-04-14
CC-19550 CC-19551 Standard logs jdbc

srishti-saraswat:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
@@ -50,7 +50,9 @@ public class JdbcSinkTask extends SinkTask {
   public void start(final Map<String, String> props) {
     log.info("Starting JDBC Sink task");
     config = new JdbcSinkConfig(props);
+    log.info("Initializing JDBC writer");
     initWriter();
+    log.info("JDBC writer initialized");
     remainingRetries = config.maxRetries;
     shouldTrimSensitiveLogs = config.trimSensitiveLogsEnabled;
     try {
@@ -90,6 +92,7 @@ public class JdbcSinkTask extends SinkTask {
       if (reporter != null) {
         unrollAndRetry(records);
       } else {
+        log.error(tace.toString());
         throw tace;
       }
     } catch (SQLException sqle) {
@@ -111,6 +114,7 @@ public class JdbcSinkTask extends SinkTask {
         initWriter();
         remainingRetries--;
         context.timeout(config.retryBackoffMs);
+        log.debug(sqlAllMessagesException.toString());
         throw new RetriableException(sqlAllMessagesException);
       } else {
         if (reporter != null) {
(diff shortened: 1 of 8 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -169,7 +170,7 @@ public void start(Map<String, String> properties) {
                                                   JdbcSourceConnectorConstants.QUERY_NAME_VALUE));
           break;
         default:
-          throw new ConnectException("Unknown query mode: " + queryMode);
+          throw new ConfigException("Unknown query mode: " + queryMode);
```
Tanish0019, line 173: Same change needed in line 213

src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java
```diff
@@ -57,8 +58,10 @@ public synchronized Connection getConnection() {
         newConnection();
       }
     } catch (SQLException sqle) {
+      log.debug("Could not establish connection with database.");
```
Tanish0019, line 61: log level should be debug or error?

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -50,7 +50,9 @@ public class JdbcSinkTask extends SinkTask {
   public void start(final Map<String, String> props) {
     log.info("Starting JDBC Sink task");
     config = new JdbcSinkConfig(props);
+    log.info("Initializing JDBC writer");
```
Tanish0019, line 53: Move initialising and initialised logs inside the `initWriter` function? As it is created again in case of a retry as well.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkTask.java
```diff
@@ -139,9 +143,11 @@ private void unrollAndRetry(Collection<SinkRecord> records) {
       try {
         writer.write(Collections.singletonList(record));
       } catch (TableAlterOrCreateException tace) {
+        log.debug(tace.toString());
         reporter.report(record, tace);
         writer.closeQuietly();
       } catch (SQLException sqle) {
+        log.debug(sqle.toString());
```
Tanish0019, line 150: Instead of putting debug here, put debug on sqlAllMessagesException in the next line as it does some sensitive log trimming

Tanish0019 (changes requested):

src/main/java/io/confluent/connect/jdbc/util/CachedConnectionProvider.java
srishti-saraswat: Kept this as debug since this method is called multiple times. Could lead to frequent logging at error level.

Tanish0019 (approved):

PR #1332, opened 2023-04-20
[CC-19887] flattened oracle related dependencies to remove pom file

Tanish0019:
[CC-19887](https://confluentinc.atlassian.net/browse/CC-19887)
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 

[CC-19887]: https://confluentinc.atlassian.net/browse/CC-19887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,20 +85,6 @@
         </pluginRepository>
     </pluginRepositories>
 
-    <!-- Adding BOM instead of directly using ojdbc8-production dependency to get rid of .pom file in final package  -->
-    <!-- Add all the dependencies mentioned in this BOM during a version upgrade   -->
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>com.oracle.database.jdbc</groupId>
-                <artifactId>ojdbc-bom</artifactId>
-                <version>${oracle.jdbc.driver.version}</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-        </dependencies>
-    </dependencyManagement>
-
     <dependencies>
         <dependency>
             <groupId>org.apache.kafka</groupId>
@@ -124,7 +110,8 @@
             <version>${postgresql.version}</version>
             <scope>runtime</scope>
         </dependency>
-        <!-- Oracle related drivers. Version is explicitly mentioned, so they get picked up by Fossa correctly -->
+        <!-- All packages added by ojdbc8-production. Do not use ojdbc8-production directly to avoid .pom file in final package -->
+        <!-- See https://www.oracle.com/database/technologies/maven-central-guide.html -->
         <dependency>
             <groupId>com.oracle.database.jdbc</groupId>
             <artifactId>ojdbc8</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -111,14 +123,63 @@
...
             <groupId>com.oracle.database.jdbc</groupId>
-            <artifactId>ojdbc8-production</artifactId>
+            <artifactId>ojdbc8</artifactId>
+            <version>${oracle.jdbc.driver.version}</version>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.oracle.database.jdbc</groupId>
+            <artifactId>ucp</artifactId>
+            <version>${oracle.jdbc.driver.version}</version>
```
anupamaggarwal, line 134: nit: we can avoid the version here? (applies to all dependent runtime libs brought by ojdbc-bom)

pom.xml
```diff
@@ -85,6 +85,18 @@
         </pluginRepository>
     </pluginRepositories>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.oracle.database.jdbc</groupId>
+                <artifactId>ojdbc-bom</artifactId>
```
anupamaggarwal, line 92: nit: maybe add a comment here asking user to explicitly mention all dependent libs brought in by this bom whenever a version upgrade happens?

pom.xml
Tanish0019: @anupamaggarwal Adding version here for Fossa because it was picking up latest package if version was not present. I'll comment this too

anupamaggarwal: @Tanish0019  didn't understand this fully , are you saying fossa requires an explicit version (and is not happy if it picks the versions from the bom in the dependency mgmt section)?

Tanish0019: If I don't supply a version fossa will pick up the latest version in its scan. I have currently manually set these dependencies to be ignored by fossa scan and given the link to their license in the comment there. I am not sure if that ignore will work if version is changed so added explicitly over here

anupamaggarwal: @Tanish0019  - I se, in that case we can probably remove the dependency management section? (where we have the bom imported in pom scope), it would be redundant since we are pinning the versions explicitly.

Tanish0019: @anupamaggarwal updated

pom.xml
```diff
@@ -110,15 +110,66 @@
             <version>${postgresql.version}</version>
             <scope>runtime</scope>
         </dependency>
+        <!-- All packages added by ojdbc8-production. Do not use ojdbc8-production directly to avoid .pom file in final package -->
```
anupamaggarwal, line 113: nit : - please add all packages imported by ojdbc8-production explicitly to pom when bumping up driver version?

anupamaggarwal (approved):
lgtm (pls feel free to incorporate the nit as per your judgement)

Issue #1287, opened 2023-01-04
 Insert explicit value for identity column mssql database


sgv0007:
Hi Team,

I am getting error while inserting records to ms sql database table. Can you please let me know if there is any solution or workaround for this problem.

java.sql.BatchUpdateException: Cannot
        insert explicit value for identity column in table 'Clients' when IDENTITY_INSERT
        is set to OFF.

similar to issue:
https://github.com/confluentinc/kafka-connect-jdbc/issues/837

Thanks,
Shiva

Issue #837, opened 2020-04-16
IDENTITY_INSERT


McPatate:
Hello,

I was wondering if it were possible to set IDENTITY_INSERT on Microsoft SQL Server Databases before importing bulks from a kafka topic.

Currently I am getting an error such as:
```
java.sql.SQLException: java.sql.BatchUpdateException: Cannot insert explicit value for identity column in table 'tbl_trameCityBox' when IDENTITY_INSERT is set to OFF.
```

It'd be great to be able to do this, I feel the only other option I have is to write my own sink connector 😓.

Best regards,

Luc

mubeen:
I am experiencing the same issue, are there any guidelines or possible solutions that are available to us to overcome this issue.

harunkucuk5:
same problem, I have not found any solution for this

deshpandeamrut:
Same problem, any solution for this?

ajeesh-s:
Same problem, any solution for this?

alam157:
I have same problem, any solution for this?

hernan812:
Anyone found a solution for this? I have debezium-debezium-connector-sqlserver-2.5.4 installed and used the new property [dialect.sqlserver.identity.insert] still error persist. 
Any help would be appreciated.
Thanks

PR #1343, opened 2023-06-01
Option for not inserting primary keys into Sql Server tables 

miller45:
## Problem
Inserting values into IDENTITY columns on SQL Server (MS) will result in SQL Error: 
<code>insert explicit value for identity column in table 'tablename' when IDENTITY_INSERT is set to OFF</code>
The are valid scenarios where you don´t want to (or cannot) alter the target tables of database just to "make kafka-connect happy".
But you might need to include the primary key in you kafka-connect config because you have to include it for the upsert mode/UPDATE case.
Related to #1287 and #837

## Solution
Added option mssql.insert.primary.keys which defaults to true. 
If set to false the INSERT will omit the primary key columns. This allows allows the database to create its own primary keys, if they are configured with IDENTITY(n,n) for autoincremental keys. This works for mode=insert and mode=upsert

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [x] System tests
- [x] Manual tests

## Release Plan
This feature is non-breaking by design. Given there is no bug, it will not change any existing behavior. So it can be released without any migration effort.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -1,0 +1,95 @@
+# This file is managed by ServiceBot plugin - Semaphore. The content in this file is created using a common
+# template and configurations in service.yml.
+# Any modifications made to ths file will be overwritten by the generated content in nightly runs.
+# For more information, please refer to the page:
+# https://confluentinc.atlassian.net/wiki/spaces/Foundations/pages/2871296194/Add+SemaphoreCI
+version: v1.0
+name: build-test-release
+agent:
+  machine:
+    type: s1-prod-ubuntu24-04-amd64-1
+
+fail_fast:
+  cancel:
+    when: "true"
+
+execution_time_limit:
+  hours: 1
+
+queue:
+  - when: "branch != 'master' and branch !~ '[0-9]+\\.[0-9]+\\.x'"
+    processing: parallel
+
+global_job_config:
+  prologue:
+    commands:
+      - checkout
+      - sem-version java 8
+      - . cache-maven restore
+
+blocks:
+  - name: Test
+    dependencies: []
+    run:
+      # don't run the tests on non-functional changes...
+      when: "change_in('/', {exclude: ['/.deployed-versions/', '.github/']})"
+    task:
+      jobs:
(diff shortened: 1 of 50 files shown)


miller45:
> ## Problem
> Inserting values into IDENTITY columns on SQL Server (MS) will result in SQL Error: `insert explicit value for identity column in table 'tablename' when IDENTITY_INSERT is set to OFF` There are valid scenarios where you don´t want to (or cannot) alter the target tables of database just to "make kafka-connect happy". But you might need to include the primary key in you kafka-connect config because you have to include it for the upsert mode/UPDATE case. Related to #1287 and #837
> 
> ## Solution
> Added option mssql.insert.primary.keys which defaults to true. If set to false the INSERT will omit the primary key columns. This allows allows the database to create its own primary keys, if they are configured with IDENTITY(n,n) for autoincremental keys. This works for mode=insert and mode=upsert
> 
> ##### Does this solution apply anywhere else?
> * [ ]  yes
> * [x]  no
> 
> ##### If yes, where?
> ## Test Strategy
> ##### Testing done:
> * [x]  Unit tests
> * [ ]  Integration tests
> * [x]  System tests
> * [x]  Manual tests
> 
> ## Release Plan
> This feature is non-breaking by design. Given there is no bug, it will not change any existing behavior. So it can be released without any migration effort.

Tanish0019:
Hi @AnthonyWhitaker, yes looks like a valid issue. I will try to get to reviewing it soon.

hernan812:
@AnthonyWhitaker @Tanish0019 
After installing debezium-debezium-connector-sqlserver-2.5.4, and added the [dialect.sqlserver.identity.insert] = true as property in the sink connector I still getting:
\njava.sql.BatchUpdateException: Cannot insert explicit value for identity column in table 'Banks' when IDENTITY_INSERT is set to OFF

Is there any other solution, advise on dealing with IDENTITY columns in SQL server for the SINK connector? 
Thanks

miller45:
@Tanish0019  signed the new CLA and merged all conflicts

PR #1348, opened 2023-06-29
[CC-21208] Fixed connector pause sending connector into provisioning status on CC

Tanish0019:
## Problem
Currently, JDBC Source connector can get stuck in provisioning state in confluent cloud if a pause and restart are triggered simultaneously. Although rare, we have seen it happen a couple of time for some customers. This happens due to how taskConfig method is defined in the connector. When a connector is paused, the task target state is set to PAUSED  and `stop` method of the connector is [called](https://github.com/confluentinc/ce-kafka/blob/88507b33de3e90afbebace923e40d9afb85c51f4/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerConnector.java#L235). If this happens to right around a restart which can happen due to a config change then:
1. connector is stopped (due to restart)
2. Started again (due to restart)
3. Stopped again (due to pause)

When connector is in between state 2 and 3, taskConfigs method inside the connector is called which will return 0 tasks from [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/345036e84afd38d015e0e36081cfab970004f21b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java#L166) . No tasks means connector is in provisioning. Even if you restart the connector now, the target state of the connector will remain paused and new tasks won't initialize as they are blocked in this [section](https://github.com/confluentinc/ce-kafka/blob/88507b33de3e90afbebace923e40d9afb85c51f4/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java#L221)

The only way to get it back to working is to resume the connector but as connector goes into provisioning, the resume button is not available to the customer.

## Solution
A simple solution would be to return 1 task with empty TABLE_CONFIG, however this will lead to [task failure](https://github.com/confluentinc/kafka-connect-jdbc/blob/345036e84afd38d015e0e36081cfab970004f21b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java#L102) each time the connector starts till all tables are fetched from the DB. This call is quite slow and can take upto a minute so connector will stay in failed state which is a pretty bad experience.

SourceTaskContext can also not be used because update needs to happen in the connector class inside the `taskReconfig` method where we don't have access to it. 

This PR addresses this by adding a new config for tasks - TABLES_FETCHED. This signifies if the first call to fetch tables has finished or not. Even when call is not finished we will create a single task but that task will not do anything in the start and poll method. When the tables are fetched from the database the task will get reconfiged and start working as usual. 

- [ ] yes
- [X] no

## Test Strategy
Need to test on CC once.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
+++ b/src/test/java/io/confluent/connect/jdbc/source/JdbcSourceTaskUpdateTest.java
@@ -887,6 +887,17 @@ public class JdbcSourceTaskUpdateTest extends JdbcSourceTaskTestBase {
   }
 
   @Test
+  public void testTaskCreatedWhileWaitingToFetchTables() throws InterruptedException {
+    initializeTask();
+    Map<String, String> props = new HashMap<>();
+    props.put(JdbcSourceTaskConfig.TABLES_CONFIG, "");
+    props.put(JdbcSourceTaskConfig.TABLES_FETCHED, "false");
+    task.start(props);
+    List<SourceRecord> records = task.poll();
+    assertNull(records);
+  }
+
+  @Test
   public void testCustomQueryMultipleRecordsWithDBPartition() throws Exception {
     expectInitializeNoOffsets(Arrays.asList(JOIN_QUERY_PARTITION));
 


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -97,15 +97,25 @@ public void start(Map<String, String> properties) {
...
+    Boolean tablesFetched = config.getBoolean(JdbcSourceTaskConfig.TABLES_FETCHED);
     String query = config.getString(JdbcSourceTaskConfig.QUERY_CONFIG);
 
     if ((tables.isEmpty() && query.isEmpty())) {
+      // We are still waiting for the tables call to complete.
+      // Start task but do nothing.
+      if (!tablesFetched) {
+        taskThreadId.set(Thread.currentThread().getId());
+        log.info("Started JDBC source task. No work will be done, waiting to fetch tables.");
+        return;
```
anupamaggarwal, line 109: If we return from here (as opposed to failing) won't this lead to some state not being initialised and throw NPEs when poll is called... So in effect the task might still fail during poll ?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTaskConfig.java
```diff
@@ -29,9 +29,12 @@ public class JdbcSourceTaskConfig extends JdbcSourceConnectorConfig {
 
   public static final String TABLES_CONFIG = "tables";
   private static final String TABLES_DOC = "List of tables for this task to watch for changes.";
-
+  public static final String TABLES_FETCHED = "tables.fetched";
```
sagarrao12, line 32: hmm not sure this is a user level config? What values are they going to provide in this? If the purpose of this config is to check if the tables have been fetched or not, it can be done by a variable as well. Not sure I follow this..

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -97,15 +97,25 @@ public void start(Map<String, String> properties) {
...
-              + " it was not assigned a table nor a query to execute."
-              + " If run in table mode please make sure that the tables"
-              + " exist on the database. If the table does exist on"
-              + " the database, we recommend using the fully qualified"
-              + " table name.");
+          + " it was not assigned a table nor a query to execute."
+          + " If run in table mode please make sure that the tables"
+          + " exist on the database. If the table does exist on"
+          + " the database, we recommend using the fully qualified"
+          + " table name.");
```
sagarrao12, line 118: this looks like an unnecessary diff due to tab spaces.

sagarrao12 (changes requested):

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTaskConfig.java
Tanish0019: This is not user level config. This is internal config of the task itself. Can't use variable for this status of query is fetched from Connector class whereas this has to be used in the task class.

sagarrao12: Ok. Having it exposed via config is  still not the right way to go. You can look at the interface `SourceTaskContext` which is exposed by the framework. I think it's designed for such cases. Also, IIRC we had also discussed a case where the query failed and the. You can even haver an AtomicBoolean object instantiated in connector and updated by the Task or TableMonitoringThread.

Tanish0019: Resolved offline. We can't use SourceTaskContext here and the config we have exposed is internal config and not user config. This pattern is already being used in this Connector with the TABLES config in the task.

ashwinpankaj: please add details in the PR description on why SourceTaskContext cannot be used.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTaskConfig.java
```diff
@@ -29,9 +29,11 @@ public class JdbcSourceTaskConfig extends JdbcSourceConnectorConfig {
 
   public static final String TABLES_CONFIG = "tables";
   private static final String TABLES_DOC = "List of tables for this task to watch for changes.";
+  public static final String TABLES_FETCHED = "tables.fetched";
 
   static ConfigDef config = baseConfigDef()
-      .define(TABLES_CONFIG, Type.LIST, Importance.HIGH, TABLES_DOC);
+      .define(TABLES_CONFIG, Type.LIST, Importance.HIGH, TABLES_DOC)
```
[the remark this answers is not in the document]
Tanish0019: This is internal task config not connector config, so it will not show up.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -389,6 +399,18 @@ protected void closeResources() {
   public List<SourceRecord> poll() throws InterruptedException {
     log.trace("Polling for new data");
 
+    // If the call to get tables has not completed we will not do anything.
+    // This is only valid in in table mode.
+    Boolean tablesFetched = config.getBoolean(JdbcSourceTaskConfig.TABLES_FETCHED);
+    String query = config.getString(JdbcSourceTaskConfig.QUERY_CONFIG);
+    if (query.isEmpty() && !tablesFetched) {
+      final long sleepMs = config.getInt(JdbcSourceTaskConfig.POLL_INTERVAL_MS_CONFIG);
```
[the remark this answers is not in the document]
Tanish0019: Following same pattern as used in poll currently. Also in case POLL_INTERVAL_MS is very small it will print too many logs

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -163,17 +163,18 @@ public List<Map<String, String>> taskConfigs(int maxTasks) {
...
-            "No tasks will be run because the connector has not been able to read "
-                + "the list of tables from the database yet"
-        );
-      } else if (currentTables.isEmpty()) {
+      if (currentTables == null || currentTables.isEmpty()) {
         taskConfigs = new ArrayList<>(1);
-        log.warn("No tables were found so there's no work to be done.");
         Map<String, String> taskProps = new HashMap<>(configProperties);
-        taskProps.put(JdbcSourceTaskConfig.TABLES_CONFIG, "[]");
+        taskProps.put(JdbcSourceTaskConfig.TABLES_CONFIG, "");
```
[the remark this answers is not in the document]
Tanish0019: Added comment explaining reasoning

src/test/java/io/confluent/connect/jdbc/source/integration/BaseOOMIntegrationTest.java
```diff
@@ -47,6 +47,7 @@ public void stopTask() {
   @Test
   public void testStreamingReads() throws InterruptedException {
```
anupamaggarwal, line 48: @Tanish0019  - can you pls add one test here for successful start if tables have not been fetched also...otherwise LGTM!

anupamaggarwal (approved):

sagarrao12 (approved):
LGTM

PR #1394, opened 2024-01-30
[CC-24497] Added validation for column check

Tanish0019:
## Problem
Ref: [CC-24497](https://confluentinc.atlassian.net/browse/CC-24497)

## Solution
Added a check that sees if column names supplied as part of configuration do exist in at least one table. This prevents misuse of these configuration which can be used for SQL injection.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 

[CC-24497]: https://confluentinc.atlassian.net/browse/CC-24497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -323,22 +323,27 @@ public class JdbcSourceTask extends SourceTask {
         Map<ColumnId, ColumnDefinition> defnsById = dialect.describeColumns(conn, table, null);
         Set<String> columnNames = defnsById.keySet().stream().map(ColumnId::name)
             .map(String::toLowerCase).collect(Collectors.toSet());
+
         if ((mode.equals(JdbcSourceTaskConfig.MODE_INCREMENTING)
             || mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP_INCREMENTING))
-            && !incrementingColumn.isEmpty()) {
-          if (!columnNames.contains(incrementingColumn.toLowerCase(Locale.getDefault()))) {
-            throw new ConnectException("Incrementing column " + incrementingColumn
-                + " does not exist");
-          }
+            && !incrementingColumn.isEmpty()
+            && !columnNames.contains(incrementingColumn.toLowerCase(Locale.getDefault()))) {
+          throw new ConfigException("Incrementing column: " + incrementingColumn
+              + " does not exist.");
         }
 
         if ((mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP)
             || mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP_INCREMENTING))
             && !timestampColumns.isEmpty()) {
-          for (String timestampColumn : timestampColumns) {
-            if (!columnNames.contains(timestampColumn.toLowerCase(Locale.getDefault()))) {
-              throw new ConnectException("Timestamp column " + timestampColumn + " does not exist");
-            }
+
+          Set<String> missingTsColumns = timestampColumns.stream()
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -309,6 +313,43 @@ public void start(Map<String, String> properties) {
...
+            throw new ConnectException("Incrementing column " + incrementingColumn
+                + " does not exist");
+          }
+        }
+
+        if ((mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP)
+            || mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP_INCREMENTING))
+            && !timestampColumns.isEmpty()) {
+          for (String timestampColumn : timestampColumns) {
+            if (!columnNames.contains(timestampColumn.toLowerCase(Locale.getDefault()))) {
```
ashoke-cube, line 339: We are returning at the first failure. Would it be better to check all column names at once and then throw an exception with all the missing columns.

You could form a `set` of timestamp columns as well and then find the missing ones as below.

```
        Set<String> lowercaseTsColumns = timestampColumns.stream()
            .map(column -> column.toLowerCase(Locale.getDefault()))
            .collect(Collectors.toSet());

        Set<String> missingTsColumns = lowercaseTsColumns.stream()
            .filter(column -> !columnNames.contains(column))
            .collect(Collectors.toSet());
```

Tanish0019: > Do you need add any new test cases for the method?
Yeah, I'll add one.

> Again wouldn't that be another security issue, that any malicious user can simply turn it off?
Oh yeah, I was thinking of an internal config but I forgot this is open source so it can be viewed.

Tanish0019:
@ashoke-cube 

> Do you need add any new test cases for the method?

I'll add one

> Again wouldn't that be another security issue, that any malicious user can simply turn it off?

Oh yeah. I was thinking of an internal config but forgot this is open source and the config can just be viewed

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
ashoke-cube: I realised we might lose the case sensitivity in column names if we convert to lowerCase, so better to do below instead.

```
        Set<String> missingTsColumns = timestampColumns.stream()
            .filter(column -> !columnNames.contains(column.toLowerCase(Locale.getDefault())))
            .collect(Collectors.toSet());
```

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -309,6 +313,43 @@ public void start(Map<String, String> properties) {
...
+      String mode, String incrementingColumn, List<String> timestampColumns, String table) {
+    try {
+      final Connection conn = cachedConnectionProvider.getConnection();
+      boolean autoCommit = conn.getAutoCommit();
+      try {
+        conn.setAutoCommit(true);
+        Map<ColumnId, ColumnDefinition> defnsById = dialect.describeColumns(conn, table, null);
+        Set<String> columnNames = defnsById.keySet().stream().map(ColumnId::name)
+            .map(String::toLowerCase).collect(Collectors.toSet());
+        if ((mode.equals(JdbcSourceTaskConfig.MODE_INCREMENTING)
```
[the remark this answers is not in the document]
Tanish0019: updated

PR #1410, opened 2024-04-15
CUP-1584 Stored Procedure support

pablodc00:
## Problem
The connector does not support reading from Stored Procedures

## Solution
Added support for reading from Stored Procedure

- [x] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -179,6 +186,10 @@ public void start(Map<String, String> properties) {
           partitions.add(Collections.singletonMap(JdbcSourceConnectorConstants.QUERY_NAME_KEY,
                                                   JdbcSourceConnectorConstants.QUERY_NAME_VALUE));
           break;
+        case STORED_PROCEDURE:
+          log.trace("Starting in STORED PROCEDURE mode");
+          partitions.add(Collections.singletonMap(JdbcSourceConnectorConstants.STORED_PROCEDURE,
+                                                  JdbcSourceConnectorConstants.STORED_PROCEDURE));
```
anastasiya-yakushina, line 192: `break;` is missed here, maybe that could be the reason of the error `"message": "Invalid value storedprocedure for configuration mode: String must be one of: , bulk, timestamp, incrementing, timestamp+incrementing"`

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -179,6 +186,10 @@ public void start(Map<String, String> properties) {
           partitions.add(Collections.singletonMap(JdbcSourceConnectorConstants.QUERY_NAME_KEY,
                                                   JdbcSourceConnectorConstants.QUERY_NAME_VALUE));
           break;
+        case STORED_PROCEDURE:
```
anastasiya-yakushina, line 189: The upper condition is 
`if (mode.equals(JdbcSourceTaskConfig.MODE_INCREMENTING)
        || mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP)
        || mode.equals(JdbcSourceTaskConfig.MODE_TIMESTAMP_INCREMENTING)) {`
        
 does it mean that we will not get here in any case?

anastasiya-yakushina:
I guess there is a little confusion in terms of modes. Mu suggestion is to have the existing display mode, which are "bulk, timestamp, incrementing and timestamp+incrementing" and add a new querier mode to have "query, table and storedprocedure". In our case bulk+storedprocedure will work fine

PR #1432, opened 2024-09-09
add replace.null.with.default for JdbcSink

SML0127:
## Problem
When used with Kafka connect, there are cases where null values ​in the record ​are changed to default values.
In kafka-connect library v3.5, [`replace.null.with.default` option was added ](https://issues.apache.org/jira/browse/KAFKA-8713)to prevent null values ​​from always being replaced with default values ​​when converting json.
But, the column constraints related to default values ​​still remain in the record, so null values ​​are entered as default values ​​when building insert/update queries in JdbcSink.

## Solution
Add  `replace.null.with.default` option to choose whether null values should be changed to default values or not.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
@@ -42,6 +42,8 @@ public class PreparedStatementBinder implements StatementBinder {
   private final JdbcSinkConfig.InsertMode insertMode;
   private final DatabaseDialect dialect;
   private final TableDefinition tabDef;
+  private final boolean replaceNullWithDefault;
+
 
   @Deprecated
   public PreparedStatementBinder(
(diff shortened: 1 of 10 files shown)


src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -169,7 +175,11 @@ protected int bindNonKeyFields(
   ) throws SQLException {
     for (final String fieldName : fieldsMetadata.nonKeyFieldNames) {
       final Field field = record.valueSchema().field(fieldName);
-      bindField(index++, field.schema(), valueStruct.get(field), fieldName);
+      if (this.replaceNullWithDefault) {
+        bindField(index++, field.schema(), valueStruct.get(field), fieldName);
+      } else {
+        bindField(index++, field.schema(), valueStruct.getWithoutDefault(field.name()), fieldName);
+      }
```
SML0127, line 182: Major changes

SML0127:
I don't think there is a problem with the current query build logic where null values ​​are replaced with default values ​​if there is a default value constraint. 
However, since this connector used with kafka-connect and kafka-connect supports `replace.null.with.default`, so I think it is necessary to support this option.

PR #1433, opened 2024-09-10
ADD: add replace.null.with.default for JdbcSink

SML0127:
## Problem
When used with Kafka connect, there are cases where null values ​in the record ​are changed to default values.
In kafka-connect library v3.5, [`replace.null.with.default` option was added ](https://issues.apache.org/jira/browse/KAFKA-8713)to prevent null values ​​from always being replaced with default values ​​when converting json.
But, the column constraints related to default values ​​still remain in the record, so null values ​​are entered as default values ​​when building insert/update queries in JdbcSink.

I don't think there is a problem with the current query build logic where null values ​​are replaced with default values ​​if there is a default value constraint. However, since this connector used with kafka-connect, and kafka-connect supports replace.null.with.default. So I think it is necessary to support this option.

## Solution
Add  `replace.null.with.default` option to choose whether null values should be changed to default values or not.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
@@ -42,6 +42,8 @@ public class PreparedStatementBinder implements StatementBinder {
   private final JdbcSinkConfig.InsertMode insertMode;
   private final DatabaseDialect dialect;
   private final TableDefinition tabDef;
+  private final boolean replaceNullWithDefault;
+
 
   @Deprecated
   public PreparedStatementBinder(
@@ -50,7 +52,8 @@ public class PreparedStatementBinder implements StatementBinder {
       JdbcSinkConfig.PrimaryKeyMode pkMode,
       SchemaPair schemaPair,
       FieldsMetadata fieldsMetadata,
-      JdbcSinkConfig.InsertMode insertMode
+      JdbcSinkConfig.InsertMode insertMode,
+      boolean replaceNullWithDefault
   ) {
     this(
         dialect,
@@ -59,7 +62,8 @@ public class PreparedStatementBinder implements StatementBinder {
         schemaPair,
         fieldsMetadata,
         null,
-        insertMode
+        insertMode,
+        replaceNullWithDefault
     );
   }
 
@@ -70,7 +74,8 @@ public class PreparedStatementBinder implements StatementBinder {
       SchemaPair schemaPair,
       FieldsMetadata fieldsMetadata,
       TableDefinition tabDef,
-      JdbcSinkConfig.InsertMode insertMode
+      JdbcSinkConfig.InsertMode insertMode,
+      boolean replaceNullWithDefault
   ) {
     this.dialect = dialect;
     this.pkMode = pkMode;
@@ -79,6 +84,7 @@ public class PreparedStatementBinder implements StatementBinder {
     this.fieldsMetadata = fieldsMetadata;
     this.insertMode = insertMode;
     this.tabDef = tabDef;
+    this.replaceNullWithDefault = replaceNullWithDefault;
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -169,7 +175,11 @@ protected int bindNonKeyFields(
   ) throws SQLException {
     for (final String fieldName : fieldsMetadata.nonKeyFieldNames) {
       final Field field = record.valueSchema().field(fieldName);
-      bindField(index++, field.schema(), valueStruct.get(field), fieldName);
+      if (this.replaceNullWithDefault) {
+        bindField(index++, field.schema(), valueStruct.get(field), fieldName);
+      } else {
+        bindField(index++, field.schema(), valueStruct.getWithoutDefault(field.name()), fieldName);
+      }
```
SML0127, line 182: Major changes

sangeet259:
Hi @SML0127, thanks for the PR. Will review and get it merged this week.

sangeet259 (approved):
LGTM! Thanks for the PR.

sangeet259:
Some tests are failing. Checking.

SML0127:
> Some tests are failing. Checking.

@sangeet259 
if there is anything to do (such as rebase), feel free to ask me.

sangeet259:
Sure, for now this is failing: `mvn test -Dtest=JdbcDbWriterTest#verifyConnectionRollbackFailed`. Can you check this ?

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
sangeet259: Small suggestion. Can we extract out the logic of getting the ObjectValue and use ternary operator there.
Then use that value in a single call to bindField ?

src/test/java/io/confluent/connect/jdbc/sink/JdbcDbWriterTest.java
```diff
@@ -137,7 +138,7 @@ private SQLException verifyConnectionRollback(boolean succeedOnRollBack) throws
     PreparedStatement mockStatement = mock(PreparedStatement.class);
     when(dialect.parseTableIdentifier(any())).thenReturn(mock(TableId.class));
     when(dialect.createPreparedStatement(any(), any())).thenReturn(mockStatement);
-    when(dialect.statementBinder(any(), any(), any(), any(), any(), any()))
+    when(dialect.statementBinder(any(), any(), any(), any(), any(), any(), eq(true)))
```
SML0127, line 141: `replaceNullWithBoolean` is type `boolean`. So, we have to use matcher `eq()`, not `any())

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -175,11 +175,10 @@ protected int bindNonKeyFields(
...
       final Field field = record.valueSchema().field(fieldName);
-      if (this.replaceNullWithDefault) {
-        bindField(index++, field.schema(), valueStruct.get(field), fieldName);
-      } else {
-        bindField(index++, field.schema(), valueStruct.getWithoutDefault(field.name()), fieldName);
-      }
+      Object objectValue = this.replaceNullWithDefault
+              ? valueStruct.get(field)
+              : valueStruct.getWithoutDefault(field.name());
+      bindField(index++, field.schema(), objectValue, fieldName);
```
SML0127, line 181: Now we use `objectValue` and call `bindField` once.

SML0127:
all tests passed
image

sangeet259:
Great thanks, I think mockito was getting confused between two versions [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/10eed32023d5b5bd826a2abd97ca9f91786765a1/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java#L534-L542) and [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/10eed32023d5b5bd826a2abd97ca9f91786765a1/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java#L560-L568).

sangeet259 (approved):
LGTM after the tests.

sangeet259:
Merged to master, need to cherry pick to lower versions.

PR #1434, opened 2024-09-11
CC-18911 - Added INFO logs for better debugging

KGaneshDatta:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/util/JdbcDriverInfo.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/JdbcDriverInfo.java
@@ -15,10 +15,16 @@
 
 package io.confluent.connect.jdbc.util;
 
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * A summary of the version information about a JDBC driver and the database.
  */
 public class JdbcDriverInfo {
+
+  private static final Logger log = LoggerFactory.getLogger(JdbcDriverInfo.class);
 
   private final int jdbcMajorVersion;
   private final int jdbcMinorVersion;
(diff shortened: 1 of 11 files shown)


b-goyal (approved):

src/main/java/io/confluent/connect/jdbc/util/JdbcDriverInfo.java
```diff
@@ -108,9 +114,13 @@ public boolean jdbcVersionAtLeast(
       int jdbcMinorVersion
   ) {
     if (this.jdbcMajorVersion() > jdbcMajorVersion) {
+      log.info("JDBC driver version {} is newer than required version {}.{}",
+               this.jdbcMajorVersion(), jdbcMajorVersion, jdbcMinorVersion);
       return true;
     }
     if (jdbcMajorVersion == jdbcMajorVersion() && jdbcMinorVersion() >= jdbcMinorVersion) {
+      log.info("JDBC driver version {} is newer than required version {}.{}",
```
Shiva-Confluent, line 122: @KGaneshDatta  we should log the major and minor version both here.
Otherwise, the log would look weird. for example, the current version is 4.2 and and we are comparing it with 4.0 then the log would say 
`JDBC driver version 2 is newer than required version 4.0`

Shiva-Confluent: should this log be an info log at all?
for CP customers this seems like info they wouldn't care much.
it may confuse them.
Also, how do we know the required version is 4.0 the wording of the log seems off here.
it seems like we need to be on the 4.0 version. where as it is just to use `blob.free()` 
functionality to free up b/clobs/nclobs proactively.

PR #1436, opened 2024-09-26
Add support for Custom Credentials Provider Class in JDBC connectors

aniketshrimal:
## Problem
- Add support to configure Custom Credentials Provider Class in the connector to fetch database credentials. 
- Currently to update db credentials, users needs to manually update the credential configs which leads to connector restart or incase they are using ConfigProviders -> require a restart to pull in new credentials. 

## Solution
- Introducing the following interfaces in connector which users can use to implement their custom class. 
```
public interface JdbcCredentialsProvider {

  /**
   * @return JdbcCredentials which caller can use for authentication purpose
   */
  JdbcCredentials getJdbcCredentials();

  /**
   * Instructs the provider to refresh or renew credentials.
   * Default behavior is no-op.
   */
  default void refresh() {
    // no need to refresh anything by default
  }
}
```
```
public interface JdbcCredentials {

  /**
   * @return Username to use for authentication to database
   */
  String getUsername();

  /**
   *
   * @return Password/Token to use for authentication to database
   */
  String getPassword();

}
```
- A new config - `jdbc.credentials.provider.class` is added which users can use to configure their custom class in the connector. Default value for this config is `io.confluent.connect.jdbc.util.DefaultJdbcCredentialsProvider`.
- Class `DefaultJdbcCredentialsProvider` and `BasicJdbcCredentials` are added as part of this PR to provide basic implementation of these interface and to use as default value for above config to ensure backward compatibility with existing credentials flow. 
- Connector uses `getJdbcCredentials()` method to fetch the database credentials to create a connection object. 
- `refresh` method in CredentialsProvider interface is not called within connector life-cycle. This method is part of interface to concretise that credentials refresh could be triggered within custom class logic.

- [ ] yes
- [ ] no

## Test Strategy
- Added relevant UTs to test the credentials provider functionality and ensure that behaviour is backward compatible. 
- Manual testing is done with AWS RDS instance for MySQL and PostgreSQL db where password refresh functionality was tested for AWS IAM user.  

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
Plan is to merge this change to master and cut a minor release branch for this backward-compatible feature.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -18,6 +18,7 @@ package io.confluent.connect.jdbc.dialect;
 import java.time.ZoneOffset;
 import java.util.TimeZone;
 
+import org.apache.kafka.common.Configurable;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.types.Password;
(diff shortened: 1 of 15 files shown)


KGaneshDatta (approved):
LGTM

PR #1451, opened 2025-01-09
Enhanced INFO Logging to check for Race Condition

KGaneshDatta:
## Problem
This PR enhances the logging in order to debug a potential race condition while updating the task configs. 

[CC-30123](https://confluentinc.atlassian.net/browse/CC-30123) - Update the misleading log line for the JDBC Driver Version.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 

[CC-30123]: https://confluentinc.atlassian.net/browse/CC-30123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
+++ b/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
@@ -159,6 +159,7 @@ public class JdbcSourceConnector extends SourceConnector {
 
   @Override
   public List<Map<String, String>> taskConfigs(int maxTasks) {
+    log.info("Starting with the task Configuration method.");
     String query = config.getString(JdbcSourceConnectorConfig.QUERY_CONFIG);
     List<Map<String, String>> taskConfigs;
     if (!query.isEmpty()) {
@@ -205,6 +206,7 @@ public class JdbcSourceConnector extends SourceConnector {
           taskProps.put(JdbcSourceTaskConfig.TABLES_FETCHED, "true");
           taskConfigs.add(taskProps);
         }
+        log.info("Current Tables size: {}", currentTables.size());
         log.trace(
             "Producing task configs with no custom query for tables: {}",
             currentTables.toArray()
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -208,6 +209,7 @@ private boolean updateTables() {
       filteredTables.addAll(allTables);
     }
 
+    log.info("Filtered tables size: {}", filteredTables);
```
rpindikura, line 212: would this cause too many log msgs ? if yes - maybe, log only if the number-of-tables has changed ?

nmala01 (approved):
LGTM

src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java
```diff
@@ -159,6 +159,7 @@ public Config validate(Map<String, String> connectorConfigs) {
 
   @Override
   public List<Map<String, String>> taskConfigs(int maxTasks) {
+    log.info("Starting with the task Configuration method.");
```
venkatteki, line 162: qq - shouldn't this be debug level if the goal is check for Race condition

nmala01: @venkatteki This is important log line to chk when this was invoked, since race condition can happen anytime, we wan to keep it at info level, so that we don need any additional config to get this log line

KGaneshDatta: Thanks for the review @venkatteki. We intended to keep it in debug but this condition is happening unexpectedly and we wanted to have logs during that time. Hence, we kept this one as INFO log.

nmala01 (approved):
LGTM

PR #1478, opened 2025-02-25
Microsecond Precision timestamp support for JDBC Connectors

KGaneshDatta:
## Problem
CC-31131 : Microsecond Precision timestamp support for JDBC Connectors.

Currently Microsecond Precision is not supported in the JDBC Connectors without the nanos_long config in timestamp granularity. 
REF - https://confluentinc.atlassian.net/wiki/spaces/~63a167967cde7bff9d76a990/pages/4128737414/Micro+Nanoprecision+Time+support+in+JDBC+Connectors

## Solution
This PR introduces a new mode i.e. MICROS_LONG in the timestamp.granularity config which when configured only the microsecond precision timestamp will be produced in kafka records.

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests
Test cases have been included in this doc - https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/4243293217/Microprecision+Large+Dates+support+in+JDBC+Connectors

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -967,7 +967,7 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
 
     MICROS_LONG(optional -> optional ? Schema.OPTIONAL_INT64_SCHEMA : Schema.INT64_SCHEMA,
         (timestamp, tz) -> DateTimeUtils.toEpochMicros(timestamp),
-        (epochMicros, tz) -> DateTimeUtils.toTimestamp((Long) epochMicros)),
+        (epochMicros, tz) -> DateTimeUtils.toMicrosTimestamp((Long) epochMicros)),
 
     NANOS_LONG(optional -> optional ? Schema.OPTIONAL_INT64_SCHEMA : Schema.INT64_SCHEMA,
         (timestamp, tz) -> DateTimeUtils.toEpochNanos(timestamp),
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -210,6 +210,7 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
       "Define the granularity of the Timestamp column. Options include: \n"
           + "  * connect_logical (default): represents timestamp values using Kafka Connect's "
           + "built-in representations \n"
+          + "  * micros_long: represents timestamp values as micros since epoch\n"
           + "  * nanos_long: represents timestamp values as nanos since epoch\n"
           + "  * nanos_string: represents timestamp values as nanos since epoch in string\n"
           + "  * nanos_iso_datetime_string: uses iso format 'yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS'\n";
@@ -963,6 +964,10 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
         : org.apache.kafka.connect.data.Timestamp.builder().build(),
         (timestamp, tz) -> timestamp,
         (timestamp, tz) -> (Timestamp) timestamp),
+
+    MICROS_LONG(optional -> optional ? Schema.OPTIONAL_INT64_SCHEMA : Schema.INT64_SCHEMA,
+     (timestamp, tz) -> DateTimeUtils.toEpochMicros(timestamp),
+     (epochMicros, tz) -> DateTimeUtils.toTimestamp((Long) epochMicros)),
 
     NANOS_LONG(optional -> optional ? Schema.OPTIONAL_INT64_SCHEMA : Schema.INT64_SCHEMA,
         (timestamp, tz) -> DateTimeUtils.toEpochNanos(timestamp),
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
@@ -211,6 +211,8 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
           + "  * connect_logical (default): represents timestamp values using Kafka Connect's "
           + "built-in representations \n"
           + "  * micros_long: represents timestamp values as micros since epoch\n"
+          + "  * micros_string: represents timestamp values as micros since epoch in string\n"
+          + "  * micros_iso_datetime_string: uses iso format 'yyyy-MM-dd'T'HH:mm:ss.SSSSSS'\n"
           + "  * nanos_long: represents timestamp values as nanos since epoch\n"
           + "  * nanos_string: represents timestamp values as nanos since epoch in string\n"
           + "  * nanos_iso_datetime_string: uses iso format 'yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS'\n";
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -77,6 +79,24 @@ public static String formatTimestamp(Date date, TimeZone timeZone) {
     }).format(date);
   }
 
+  private static Long convertToEpochMicros(Timestamp t) {
+    Long epochMillis = TimeUnit.SECONDS.toMicros(t.getTime() / MILLISECONDS_PER_SECOND);
+    Long microsInSecond = TimeUnit.MICROSECONDS.toMicros(t.getNanos() / MILLISECONDS_PER_SECOND);
+    return epochMillis + microsInSecond;
```
rajdangwal, line 85: nit: Better readability.
```suggestion
    Long microsComponent = t.getNanos() / NANOSECONDS_PER_MICROSECOND;
    return epochMillis + microsComponent;
```

Where NANOSECONDS_PER_MICROSECONDS is a constant defined as:
`static final long NANOSECONDS_PER_MICROSECOND = TimeUnit.MICROSECONDS.toNanos(1);`

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -123,6 +143,24 @@ public static String toIsoDateTimeString(Timestamp timestamp, TimeZone tz) {
...
+  /**
+   * Get {@link Timestamp} from epoch with micro precision
+   *
+   * @param micros epoch micro in long
+   * @return the equivalent java sql Timestamp
+   */
+  public static Timestamp toMicrosTimestamp(Long micros) {
+    return Optional.ofNullable(micros)
+        .map(
+            n -> {
```
rajdangwal, line 155: nit:
```suggestion
            m -> {
```

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -123,6 +143,24 @@ public static String toIsoDateTimeString(Timestamp timestamp, TimeZone tz) {
...
+              Timestamp ts = new Timestamp(micros / MICROSECONDS_PER_MILLISECOND);
+              ts.setNanos(
+                  (int) ((micros % MICROSECONDS_PER_SECOND) * MICROSECONDS_PER_MILLISECOND));
```
rajdangwal, line 158: I think the calculation might be off here. Can you please confirm once?

To calculate the nanoseconds portion, the following should be done:
```
// First get the remainder microseconds after dividing by milliseconds
long remainderMicros = micros % MICROSECONDS_PER_MILLISECOND;
// Convert microseconds to nanoseconds for the setNanos method
int nanos = (int)(remainderMicros * NANOSECONDS_PER_MICROSECOND);
ts.setNanos(nanos);
```

rajdangwal (changes requested):
Thanks @KGaneshDatta 
I have taken the first pass and left some comments.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -210,6 +210,7 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
       "Define the granularity of the Timestamp column. Options include: \n"
           + "  * connect_logical (default): represents timestamp values using Kafka Connect's "
           + "built-in representations \n"
+          + "  * micros_long: represents timestamp values as micros since epoch\n"
```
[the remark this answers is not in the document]
poojakuntalcflt: @rajdangwal @KGaneshDatta : Yes, microprecision should also provide `micros_string` and `micros_iso_datetime_string` same as nanos.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
KGaneshDatta: I have tested the calculations and it is working fine for all the test cases. IMO, the current approach and suggested approach produce the same results as 

- Current approach:  
Takes remainder of microseconds in a second (% MICROSECONDS_PER_SECOND)
Converts to nanoseconds by multiplying by 1000 (MICROSECONDS_PER_MILLISECOND)

- Suggested approach:  
Takes remainder of microseconds in a millisecond (% MICROSECONDS_PER_MILLISECOND)
Converts to nanoseconds by multiplying by 1000 (NANOSECONDS_PER_MICROSECOND)

But I agree the suggested approach is more direct and clear. Thanks for suggestion @rajdangwal. 
Updated it 🚀

rajdangwal: @KGaneshDatta Thanks for considering the suggestion. I think it makes the code much clearer to read. But looks like I made a small mistake in the suggested approach.

The `ts.setNanos()` method in `Timestamp` class doesn't just set the sub-millisecond or sub-nanoseconds portion, it sets the entire sub-second field, which includes the millisecond portion as well.
Since, you were computing the reminder portion with `(micros % MICROSECONDS_PER_SECOND)` which included the milliseconds part, you were getting the correct output for the converted timestamp.

I was wrong when considering the `setNanos` method to only set the microseconds and nanoseconds portion. So the above proposed alternative is incomplete as I did not consider the milliseconds part in my approach.

When setting the nanos portion lets also add the existing milliseconds portion which can be obtained from `ts.getNanos()` after we have initialized the timestamp in milliseconds:
```
ts.setNanos(ts.getNanos() + (int)(remainderMicros * NANOSECONDS_PER_MICROSECOND));
```
Thanks.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -159,15 +244,28 @@ public static Timestamp toTimestamp(String nanos) throws NumberFormatException {
    * @param tz the timezone of the source database
    * @return the equivalent java sql Timestamp
    */
-  public static Timestamp toTimestampFromIsoDateTime(String isoDT, TimeZone tz) {
+  public static Timestamp toTimestampFromIsoDateMicrosTime(String isoDT, TimeZone tz) {
```
rajdangwal, line 247: nit: Fix the javadoc
```
/**
* Get {@link Timestamp} from epoch with micro precision
*
* @param isoDT iso dateTime format "yyyy-MM-dd'T'HH:mm:ss.SSSSSS"
```

rajdangwal (commented):
Thanks @KGaneshDatta 
Left some minor comments.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
KGaneshDatta: Ack. Thanks for the suggestion @rajdangwal. Updated as suggested.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
rajdangwal: We need to have 6 `S` in the format. Looks like there are 3 additional currently.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -167,7 +265,4 @@ public static Timestamp toTimestampFromIsoDateTime(String isoDT, TimeZone tz) {
         .map(Timestamp::from)
         .orElse(null);
   }
-
-  private DateTimeUtils() {
-  }
```
rajdangwal, line 172: Ohh, looks like you had moved it from bottom to some other place earlier and I thought it was redundant.
Please add this back.

KGaneshDatta: Ack. With the intention of not being used/referred, I have removed this constructor. But it is good to have I have added it back.

rajdangwal (dismissed):
Thanks @KGaneshDatta 
LGTM.

rajdangwal (approved):
LGTM.

PR #1479, opened 2025-03-03
add compaction for batch by primary key fields

typik89:
## Problem
I'm using jdbc sink task to save data to Postgresql. I wanted to use configuration property of jdbc driver reWriteBatchedInserts=true to optimize saving data. 
This property toggles converting my separate upsert queries to one multi-value upsert. But there is problem when such multi-value query contains updates for same primary key.

## Solution
I think it would be useful to have a feature that, before forming batches of database queries, reduces batches of records by keeping only the most recent record for a primary key.

record(primaryKey,value): (1,1),(2,2),(1,3) -> (2,2),(1,3)

It might be practical not only for my case and potentially could reduce number of queries when there are updates by same primary key.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/BatchPKCompaction.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/BatchPKCompaction.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2018 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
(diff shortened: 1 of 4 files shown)


typik89:
> ❌ Error getting contributor login(s). Please ensure the email address associated with this commit is [added to your Github account](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user).

done

Issue #921, opened 2020-09-17
Support preferred postgres data type of timestamp with time zone


jfinzel:
Our biggest problem with the JDBC Sink Connector for Postgres is that it does not support `timestamptz` or `timestamp with time zone`, which is the standard and preferred method of storing time in postgres - in fact we don't even allow `timestamp without time zone` anymore.  There are serious issues with `timestamp without time zone`, elaborated in the Postgres wiki: https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29

It is possible to create a table manually using `timestamp with time zone`, but the JDBC Sink Connector interprets the timestamps always as UTC offset, with apparently no option to configure it differently. The db.timezone property does not make any difference.  This is evident in looking at the logs where the hardcoded value for each timestamp shows offset +00 added to the end, indicating a UTC offset.  I have db_timezone set to America/New_York in my config.

This is using version 5.5.0.  I would be grateful for any insight or workaround, most especially for supporting this data type natively!

alisator:
Hello I encoutered the same issue, we use timestamptz.
Error message below.

```
Caused by: org.postgresql.util.PSQLException: ERROR: column "df_created" is of type timestamp with time zone but expression is of type character varying
  Hint: You will need to rewrite or cast the expression.
```

I found, there is a custom [debezium convertor](https://github.com/oryanmoshe/debezium-timestamp-converter), which has limits in precision, but it might work.

marcotollini:
While we wait for an official solution, a temporary fix is as follows:

`CREATE OR REPLACE FUNCTION varchar_to_timestamp (varchar) RETURNS timestamptz AS $$ SELECT to_timestamp($1, 'YYYY-MM-DDTHH24:MI:SS.USZ') $$ LANGUAGE SQL;
CREATE CAST (varchar as timestamptz ) WITH FUNCTION varchar_to_timestamp (varchar) AS IMPLICIT;`

It is far from perfect, but it works.

jfinzel:
@marcotollini thanks for the input.  This is similar to our own workaround which is to do no conversions at all on raw data but only through views using postgres built-in functions, which can also be indexed.  For example we don't even convert bigint epochs in Kafka but land them as-is and provide conversions through views.

alisator:
> While we wait for an official solution, a temporary fix is as follows:
> 
[...]

PR #1480, opened 2025-03-04
Add support for timestamptz/timestamp with PostgresqlDialect

mdedetrich:
## Problem

This PR is an attempt to solve https://github.com/confluentinc/kafka-connect-jdbc/issues/921, specifically the issue at https://github.com/confluentinc/kafka-connect-jdbc/issues/921#issuecomment-742590395 where even if the database column is `timestamp with timezone`/`timestamp`, the JDBC Sink Connector still cannot insert data into the database because you get the error 

```
Caused by: org.postgresql.util.PSQLException: ERROR: column "df_created" is of type timestamp with time zone but expression is of type character varying
  Hint: You will need to rewrite or cast the expression.
```

Note that this PR doesn't update the create table statements to use `timestamp with timezone`/`timestamp` as I don't think its possible to do this comprehensively. I am not sure about avro, but I use json with json schema and from what I can tell the connect API doesn't have access to the `format` field in the JSON RFC Schema spec which would tell it to create a column with the correct types, i.e. 

```json
"meta_ingested_at" : {
    "type" : "string",
    "format" : "datetime"
}
```

(in this case the `datetime` specifically tells us that you would need to create a `timestamp with timezone` column, where as if its `datetime-local` it would be `timestamp`).

## Solution

The solution to this issue is to cast the expressions to `timestamptz`/`timestamp` using Postgres's JDBC drivers native support with `OffsetDateTime`/`LocalDateTime` (see https://tada.github.io/pljava/use/datetime.html).

Note that there is a workaround, i.e. using `?stringtype=unspecified` from https://github.com/confluentinc/kafka-connect-jdbc/issues/921#issuecomment-811341564 however this has the following issues

* It doesn't work with supported connectors on the confluent platform as there is no way to configure the JDBC connection string
* It may have unintended consequences for other column types or in other words the reason why this is a workaround is because your telling the postgres server to assume the type and automatically cast it to what it thinks is correct. The more principled solution is the statements being sent to postgres having the correct type in the first place.

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
This feature should be safe to release as it only fixes currently broken behaviour and doesn't have an impact otherwise. If you  happen to be using the `?stringtype=unspecified` workaround then it will stop applying after this fix as the workaround only effects types with no casts (and the point of this PR is to add a cast).

I can handle back-porting to older versions of connectors if necessary

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -44,12 +44,18 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+
+import static java.time.format.DateTimeFormatter.ISO_DATE_TIME;
+import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
 
 /**
  * A {@link DatabaseDialect} for PostgreSQL.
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -280,6 +312,13 @@ protected ColumnConverter columnConverterFor(
         if (UUID.class.getName().equals(columnDefn.classNameForType())) {
           return rs -> rs.getString(col);
         }
+
+        if (OffsetDateTime.class.getName().equals(columnDefn.classNameForType())) {
+          return rs -> rs.getObject(col, OffsetDateTime.class).format(ISO_OFFSET_DATE_TIME);
```
mdedetrich, line 317: Using `format` here with explicit formatters is strictly not necessary as you could use `toString` instead (the value output is the same), but the usage of an explicit formatter is better as there is no risk that it would change in the future (where as `.toString` representations can)

PR #1482, opened 2025-03-09
Removed usage of kafka utils mkSet

CodeUnravel:
## Problem
Utils.mkSet is removed from apache kafka. Utilities functions are meant for the repository and should not be used by clients. 

## Solution
Removed the usage of kafka utility function Utils.mkSet


KGaneshDatta (approved):
LGTM

PR #1483, opened 2025-03-10
Support Large Dates with Nanos Precision

KGaneshDatta:
## Problem
[CC-31133](https://confluentinc.atlassian.net/browse/CC-31133)
Currently, nanosecond-precision timestamps are capped at the year 2262 due to the limitations of the Long data type's range which is a constraint to support maximum dates up to 9999-12-31.

## Solution
As a solution, we configure the timestamp conversion as a BigInt to accommodate extended date ranges. This approach is currently applied to the nanos_string configuration, which simply converts the nanos_long value to a string. However, for dates beyond the year 2262, their nanosecond representation exceeds the maximum value of a Java Long. Therefore, we are updating nanos_string to correctly represent these extended values.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests
Test cases have been included in this doc - https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/4243293217/Microprecision+Large+Dates+support+in+JDBC+Connectors

## Release Plan

 

[CC-31133]: https://confluentinc.atlassian.net/browse/CC-31133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
@@ -150,9 +150,9 @@ public class DateTimeUtils {
    */
   public static String toEpochNanosString(Timestamp timestamp) {
     return Optional.ofNullable(timestamp)
-            .map(DateTimeUtils::convertToEpochNanosBigInt)
-            .map(String::valueOf)
-            .orElse(null);
+        .map(DateTimeUtils::convertToEpochNanosBigInt)
+        .map(String::valueOf)
+        .orElse(null);
   }
 
   /**
@@ -235,12 +235,11 @@ public class DateTimeUtils {
   /**
    * Get {@link Timestamp} from epoch with nano precision
    *
-   * @param nanos epoch nanos in string
+   * @param nanos epoch nanos BigInteger
    * @return the equivalent java sql Timestamp
    */
-  public static Timestamp toTimestamp(String nanos) {
+  public static Timestamp toTimestamp(BigInteger nanos) {
     return Optional.ofNullable(nanos)
-        .map(BigInteger::new)
         .map(
             n -> {
               long milliseconds =
@@ -251,6 +250,19 @@ public class DateTimeUtils {
               return ts;
             })
         .orElse(null);
+  }
+
+  /**
+   * Get {@link Timestamp} from epoch with nano precision
+   *
+   * @param nanos epoch nanos in string
+   * @return the equivalent java sql Timestamp
+   */
+  public static Timestamp toTimestamp(String nanos) {
+    return Optional.ofNullable(nanos)
+            .map(BigInteger::new)
+            .map(DateTimeUtils::toTimestamp)
+            .orElse(null);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -141,10 +142,13 @@ public static String toEpochMicrosString(Timestamp timestamp) {
    * @return the epoch nanoseconds string
    */
   public static String toEpochNanosString(Timestamp timestamp) {
-    return Optional.ofNullable(timestamp)
-        .map(DateTimeUtils::convertToEpochNanos)
-        .map(String::valueOf)
-        .orElse(null);
+    if (timestamp == null) {
```
rajdangwal, line 145: nit: We can keep the method chaining style as it was before.
```
public static String toEpochNanosString(Timestamp timestamp) {
    return Optional.ofNullable(timestamp)
        .map(DateTimeUtils::convertToEpochNanosBig)
        .map(BigInteger::toString)
        .orElse(null);
}
```
And keep the computation inside `convertToEpochNanosBig` method in DateTimeUtils.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -230,11 +234,17 @@ public static Timestamp toTimestamp(Long nanos) {
    * @param nanos epoch nanos in string
    * @return the equivalent java sql Timestamp
    */
-  public static Timestamp toTimestamp(String nanos) throws NumberFormatException {
-    return Optional.ofNullable(nanos)
-        .map(Long::parseLong)
-        .map(DateTimeUtils::toTimestamp)
-        .orElse(null);
+  public static Timestamp toTimestamp(String nanos) {
```
rajdangwal, line 237: nit: Similar to previous comment we can keep the existing coding style.

rajdangwal (commented):
Thanks @KGaneshDatta 
Left some minor nits.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
KGaneshDatta: Ack.
Updated the both methods with the existing code framework only.

src/main/java/io/confluent/connect/jdbc/util/DateTimeUtils.java
```diff
@@ -230,10 +238,18 @@ public static Timestamp toTimestamp(Long nanos) {
...
+        .map(BigInteger::new)
+        .map(
+            n -> {
+              long milliseconds =
+                  n.divide(BigInteger.valueOf(NANOSECONDS_PER_MILLISECOND)).longValue();
+              int nanoseconds = n.mod(BigInteger.valueOf(NANOSECONDS_PER_SECOND)).intValue();
+              Timestamp ts = new Timestamp(milliseconds);
+              ts.setNanos(nanoseconds);
+              return ts;
+            })
```
rajdangwal, line 252: nit: Lets create an overridden method `toTimestamp(BigInteger nanos)` to do this conversion.

KGaneshDatta: Added the new overload method toTimestamp(Bigint) to abstract the conversion.

rajdangwal (approved):
Thanks @KGaneshDatta 
LGTM.

PR #1486, opened 2025-03-16
feat: set JDBC credentials from environmental variables

pvannierop:
## Problem
At present, database credentials cannot be set via the environment from a container. This method of secret provisioning is widely used on containers running an application. For instance, when database passwords are not yet available when the ConfigMap on Kubernetes that holds the database credentials as `database.user` and `database.password` is generated, it is difficult to provide correct database credentials at runtime.

Note: This can be tackled with the `valueFromEnv` mechanism on Kubernetes.

## Solution
This PR adds a small class that acts as `JdbcCredentialsProvider` that reads JDBC credentials from the _CONNECT_JDBC_CONNECTION_USER_ and _CONNECT_JDBC_CONNECTION_PASSWORD_ environmental variables.

Note: The names were chosen somewhat consistent with official env var nomenclature for connect components described [here](https://docs.confluent.io/platform/current/installation/docker/config-reference.html).

- [ ] yes
- [x] no

## Test Strategy
Since this PR adds very little in terms of application logic, I did not add tests for now. When the feature added here is accepted, I would be more that happy to write some tests.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


pvannierop:
I noticed, however, that the expected log messages are not transmitted by this class.

PR #1495, opened 2025-03-30
Timestamp Precision Support for JDBC Sink Connectors

KGaneshDatta:
## Problem
Support the Microsecond and Nanosecond precision in all the JDBC Connectors. Currently, the timestamp fields would only support the millisecond precision when sourcing the records to kafka topic and sinking them to the end databases. 

## Solution
With this PR, the microsecond and nanosecond precision for all the supported databases in JDBC connectors would be supported.
Design Wiki - https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/4315349035/Microsecond+Nanosecond+precision+Timestamps+in+JDBC+Connectors

- [ ] yes
- [ ] no

## Test Strategy
Done the Manual Testing using the Local CP setup.
Source Connector Testing - https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/4316170071/Testing+Microsecond+Nanosecond+precision+Timestamps+in+JDBC+Source+Connectors

Sink Connector Testing - https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/4319838406/Testing+Microsecond+Nanosecond+precision+Timestamps+in+JDBC+Sink+Connectors

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -657,6 +704,8 @@ public JdbcSinkConfig(Map<?, ?> props) {
         DateTimezone.valueOf(getString(DATE_TIMEZONE_CONFIG).toUpperCase());
     dateTimeZone = dateTimezoneConfig.equals(DateTimezone.UTC)
         ? TimeZone.getTimeZone(ZoneOffset.UTC) : timeZone;
+    timestampPrecisionMode =
+        TimestampPrecisionMode.valueOf(getString(TIMESTAMP_PRECISION_MODE_CONFIG).toUpperCase());
```
sudeshwasnik, line 708: we are setting the values here but we are not using them anywhere. 
could you please change the Dialect implementations to directly use these values ?

src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
```diff
@@ -93,6 +94,10 @@ protected String getSqlType(SinkRecordField field) {
       case INT32:
         return "INTEGER";
       case INT64:
+        if (config instanceof JdbcSinkConfig
+             && config.getList(JdbcSinkConfig.TIMESTAMP_FIELDS_WHITELIST).contains(field.name())) {
+          return "TIMESTAMP";
+        }
```
sudeshwasnik, line 100: nit : just to make it look simpler -> 
```
    boolean shouldConvertToTimestamp = config instanceof JdbcSinkConfig
            && config.getList(JdbcSinkConfig.TIMESTAMP_FIELDS_WHITELIST).contains(field.name());
    switch (field.schemaType()) {
      case INT8:
        return "SMALLINT";
      case INT16:
        return "SMALLINT";
      case INT32:
        return "INTEGER";
      case INT64:
        return shouldConvertToTimestamp ? "TIMESTAMP" : "BIGINT";
      case FLOAT32:
        return "FLOAT";
      case FLOAT64:
        return "DOUBLE";
      case BOOLEAN:
        return "SMALLINT";
      case STRING:
        return shouldConvertToTimestamp ? "TIMESTAMP" : "VARCHAR(32672)";
      case BYTES:
        return "BLOB(64000)";
      default:
        return super.getSqlType(field);
    }
```

sudeshwasnik (commented):
hi @KGaneshDatta  ! Nice work ! 
Overall logic looks good ! 

I put some suggestions on code-refactoring. 
I see an implementation done for `timestampGranularity` ([ref](https://github.com/confluentinc/kafka-connect-jdbc/blob/9313e1f4c3e6ff564477ddefdf97e85475e5a677/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java#L231-L235)). Lets take inspiration from it and use the same in this PR.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -262,6 +267,23 @@ public enum DateTimezone {
       + "the timezone set for db.timzeone configuration (to maintain backward compatibility). It "
       + "is recommended to set this to UTC to avoid conversion for DATE type values.";
 
+  public static final String TIMESTAMP_FIELDS_WHITELIST = "timestamp.fields.whitelist";
```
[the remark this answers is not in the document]
KGaneshDatta: how abt timestamp.coversion.fields.whitelist ?

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -262,6 +267,23 @@ public enum DateTimezone {
...
+  private static final String TIMESTAMP_FIELDS_WHITELIST_DEFAULT = "";
+  private static final String TIMESTAMP_FIELDS_WHITELIST_DOC =
+      "List of comma-separated record value timestamp field names. If empty, "
+       + "no fields from the record value are utilized, otherwise used to  "
+       + "filter to the desired fields";
+  private static final String TIMESTAMP_FIELDS_WHITELIST_DISPLAY = "Timestamp Fields Whitelist";
```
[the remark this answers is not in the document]
KGaneshDatta: Yeah. Initially thought of updating in the public documentation. But will be added in this config doc also.
As we don't have the validation class in JDBC, we can't have a check for that and also throwing an exception is also not possible with the flow.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -262,6 +267,26 @@ public enum DateTimezone {
...
+
+  public static final String TIMESTAMP_FIELDS_WHITELIST = "timestamp.conversion.fields";
+  private static final String TIMESTAMP_FIELDS_WHITELIST_DEFAULT = "";
+  private static final String TIMESTAMP_FIELDS_WHITELIST_DOC =
+      "List of comma-separated record value timestamp field names that should be converted "
+          + "to timestamps. These fields will be converted based on precision mode specified in "
+          + TIMESTAMP_PRECISION_MODE_CONFIG
+          + "(microseconds or nanoseconds). The timestamp fields included here "
+          + "should be Long or String type and nested fields are not supported.";
+  private static final String TIMESTAMP_FIELDS_WHITELIST_DISPLAY = "Timestamp Fields Whitelist";
```
sudeshwasnik, line 288: lets add that its only applicable to Field of type INT_64 and STRING

pom.xml
```diff
@@ -119,6 +120,12 @@
             <version>${postgresql.version}</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <version>${mysql.version}</version>
+            <scope>runtime</scope>
+        </dependency>
```
KGaneshDatta, line 128: Will remove this

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -262,6 +267,26 @@ public enum DateTimezone {
...
+  public static final String TIMESTAMP_PRECISION_MODE_DEFAULT =
+      TimestampPrecisionMode.MICROSECONDS.toString();
+  private static final String TIMESTAMP_PRECISION_MODE_CONFIG_DISPLAY =
+      "Sink Timestamp Precision mode";
+  private static final String TIMESTAMP_PRECISION_MODE_CONFIG_DOC =
+      "Convert the Timestamp with precision. If set to microseconds, "
+          + "the timestamp will be converted to microsecond precision. If set to "
+          + "nanoseconds the timestamp will be converted to nanoseconds precision.";
+
+  public static final String TIMESTAMP_FIELDS_WHITELIST = "timestamp.conversion.fields";
```
rajdangwal, line 280: nit: Let's also update the variable name.
```suggestion
  public static final String TIMESTAMP_CONVERSION_FIELDS = "timestamp.conversion.fields";
```

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -622,9 +673,11 @@ public enum DateTimezone {
   public final PrimaryKeyMode pkMode;
   public final List<String> pkFields;
   public final Set<String> fieldsWhitelist;
+  public final Set<String> timestampFieldsWhitelist;
```
rajdangwal, line 676: nit: Also update the variable name here.

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -1935,7 +1992,6 @@ protected void formatColumnValue(
       case INT64:
       case FLOAT32:
       case FLOAT64:
-        // no escaping required
```
rajdangwal, line 1938: nit: Non-required change?

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -113,6 +114,10 @@ protected String getSqlType(SinkRecordField field) {
       case INT32:
         return "INT";
       case INT64:
+        if (config instanceof JdbcSinkConfig
+             && config.getList(JdbcSinkConfig.TIMESTAMP_FIELDS_WHITELIST).contains(field.name())) {
+          return "TIMESTAMP(6)";
```
rajdangwal, line 119: Q: Why specifying the precision with LONG, but not with STRING?

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -247,6 +250,10 @@ protected String getSqlType(SinkRecordField field) {
       case INT32:
         return "NUMBER(10,0)";
       case INT64:
+        if (config instanceof JdbcSinkConfig
```
rajdangwal, line 253: Q: Same question, why not using the precision here similar to STRING case?

src/main/java/io/confluent/connect/jdbc/sink/BufferedRecords.java
```diff
@@ -123,17 +123,15 @@ public List<SinkRecord> add(SinkRecord record) throws SQLException, TableAlterOr
           config,
           connection,
           tableId,
-          fieldsMetadata
-      );
+          fieldsMetadata);
```
rajdangwal, line 126: nit: Non-required changes.

rajdangwal (commented):
Thanks @KGaneshDatta. Great work.

Left few comments/questions.
Also, can we please add some ITs for the changes?

pom.xml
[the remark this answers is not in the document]
KGaneshDatta: My bad. I have included this as part of my Local testing and pushed the same changes. Will remove it.

src/main/java/io/confluent/connect/jdbc/dialect/Db2DatabaseDialect.java
```diff
@@ -93,6 +94,17 @@ protected String getSqlType(SinkRecordField field) {
       case INT32:
         return "INTEGER";
       case INT64:
+        if (config instanceof JdbcSinkConfig
+             && config.getList(JdbcSinkConfig.TIMESTAMP_FIELDS_WHITELIST).contains(field.name())) {
+          if (((JdbcSinkConfig) config).timestampPrecisionMode
+               == JdbcSinkConfig.TimestampPrecisionMode.MICROSECONDS) {
+            return "TIMESTAMP";
+          }
```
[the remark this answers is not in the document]
KGaneshDatta: We have checked with Postgres, MSSQL Server, MySQL & Oracle Databases. 
All the databases would have the same conversion logic, the only difference for each database would be the timestamp column attribute while creating the new table if not exists and that should be having the appropriate Datatype as supported by the database.
I have checked the official documentation for the other DBs and updated the data types accordingly.

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
KGaneshDatta: Assumed that default is microsecond, have kept it. But to have nanoseconds it should have been there. Added it.

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
KGaneshDatta: Somehow this change got reverted. This has been in place while testing and all the test cases have been done using this. Now, added it.

rajdangwal (approved):
Thanks @KGaneshDatta 
LGTM.

Please update the field name from `timestamp.conversion.fields` in the [test wiki](https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/4319838406/Testing+Microsecond+Nanosecond+precision+Timestamps+in+JDBC+Sink+Connectors) as well.

Issue #1446, opened 2024-11-21
org.apache.kafka.common.config.ConfigException: Incrementing column: ID does not exist.


c0desurfer:
[This introduced check](https://github.com/confluentinc/kafka-connect-jdbc/pull/1394) seems to work incorrectly. We have the following JDBC connector configuration.

```
{
	"numeric.mapping": "best_fit",
	"table.whitelist": <SNIP>",
	"auto.register.schemas": "false",
	"auth.type": "<SNIP>",
	"tasks.max": "1",
	"transforms.createKey.fields": "<SNIP>",
	"connector.class": "JdbcSourceConnector",
	"transforms.RenameField.renames": "<SNIP>",
	"consumer.override.security.protocol": "<SNIP>",
	"transforms.reroute.type": "org.apache.kafka.connect.transforms.RegexRouter",
	"transforms.RenameField.type": "org.apache.kafka.connect.transforms.ReplaceField$Value",
	"producer.override.sasl.mechanism": "<SNIP>",
	"producer.override.bootstrap.servers": "<SNIP>",
	"transforms.ReplaceField.type": "org.apache.kafka.connect.transforms.ReplaceField$Value",
	"topic.prefix": "<SNIP>",
	"value.converter.schema.registry.url": "<SNIP>",
	"incrementing.column.name": "ID",
	"value.converter.auto.register.schemas": "false",
	"timestamp.column.name": "<SNIP>",
	"validate.non.null": "false",
	"key.converter": "io.confluent.connect.json.JsonSchemaConverter",
	"consumer.override.sasl.jaas.config": "<SNIP>",
	"value.converter": "io.confluent.connect.json.JsonSchemaConverter",
	"connection.attempts": "480",
	"connection.backoff.ms": "30000",
	"key.converter.schema.registry.url": "<SNIP>",
	"connection.url": "jdbc:oracle:thin:@(DESCRIPTION = (ADDRESS_LIST = (FAILOVER = ON)(LOAD_BALANCE = OFF)(ADDRESS = (PROTOCOL = TCP)(HOST = <SNIP>)(PORT = <SNIP>))(ADDRESS = (PROTOCOL = TCP)(HOST = <SNIP>)(PORT = <SNIP>))) (CONNECT_DATA = (SERVICE_NAME = <SNIP>)))",
	"connection.user": "<SNIP>",
	"key.converter.schemas.enable": "true",
	"key.converter.auto.register.schemas": "false",
[...]

PR #1496, opened 2025-03-30
fix: include synonyms on Oracle connections

michherren:
## Problem
Solves: https://github.com/confluentinc/kafka-connect-jdbc/issues/1446

## Solution
This fix sets the includeSynonyms flag to true on the connection using the SYNONYM table type.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -17,6 +17,8 @@ package io.confluent.connect.jdbc.source;
 
 import java.sql.SQLNonTransientException;
 import java.util.TimeZone;
+
+import oracle.jdbc.driver.OracleConnection;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.connect.errors.ConnectException;
@@ -59,6 +61,7 @@ import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig.TransactionIso
 public class JdbcSourceTask extends SourceTask {
   // When no results, periodically return control flow to caller to give it a chance to pause us.
   private static final int CONSECUTIVE_EMPTY_RESULTS_BEFORE_RETURN = 3;
+  private static final String TABLE_TYPE_SYNONYM = "SYNONYM";
 
   private static final Logger log = LoggerFactory.getLogger(JdbcSourceTask.class);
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -196,6 +199,12 @@ public void start(Map<String, String> properties) {
     TimeZone timeZone = config.timeZone();
     String suffix = config.getString(JdbcSourceTaskConfig.QUERY_SUFFIX_CONFIG).trim();
 
+    final List<String> tableTypes = config.getList(JdbcSourceConnectorConfig.TABLE_TYPE_CONFIG);
+    final Connection conn = cachedConnectionProvider.getConnection();
+    if (tableTypes.contains(TABLE_TYPE_SYNONYM) && conn instanceof OracleConnection) {
```
poojakuntalcflt, line 204: @michherren : the `setIncludeSynonyms` method of the `OracleConnection` interface controls whether synonym information is included when retrieving metadata, particularly with the getColumns method. By default, this inclusion is disabled to optimize performance, as incorporating synonym data can significantly slow down metadata retrieval due to the necessity of complex joins. 
Enabling it will reduce the performance drastically.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -196,6 +199,12 @@ public void start(Map<String, String> properties) {
...
+      ((OracleConnection) conn).setIncludeSynonyms(true);
```
poojakuntalcflt, line 205: Also, Please avoid adding DB specific code in common Class. We have set of Dialect classes which can be used in such case.

poojakuntalcflt (changes requested):

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
poojakuntalcflt: @michherren : Alternate approach -> Resolving SYNONYM to its base table and then use the table to fetch its columns - I have created a reference PR - https://github.com/confluentinc/kafka-connect-jdbc/pull/1498

michherren: Hi @poojakuntalcflt 
Thanks a lot for the feedback! Your approach with a more generic solution definitely makes sense. For me it's just important to have a solution :) 
So I will close this PR in favour of the new one.

michherren: Question: Can I contribute some tests? In my PR I added an integration test for Oracle - maybe this can be updated to also include some of the other synonym capable db-engines?

michherren:
Resolved in https://github.com/confluentinc/kafka-connect-jdbc/pull/1498

PR #1498, opened 2025-04-08
Adding validation for table.type:SYNONYMS

poojakuntalcflt:
## Problem
Addresses: https://github.com/confluentinc/kafka-connect-jdbc/issues/1446
Also addressing this [limitation](https://github.com/microsoft/mssql-jdbc/issues/1814) in the connector to resolve Synonyms for the Sql Server JDBC Source Connector. 

## Solution
Added the resolveSynonym() method in respective dialects to resolve the actualTable from the Synonym name.
Few databases doesn't support the Synonym types for which the exceptions have been logged.

- [ ] yes
- [ ] no

## Test Strategy
Manually tested on Local Confluent Platform

**Oracle JDBC Source :-**

image

**SqlServer JDBC Source :-**

image
 

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -328,12 +328,6 @@
             <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.testcontainers</groupId>
-            <artifactId>mssqlserver</artifactId>
-            <version>1.20.0</version>
-            <scope>test</scope>
-        </dependency>
 
         <!-- common pom defines them in dependency management and hence they're pulled in compile scope.
         Explicitly define them as test to prevent from them form being packaged. -->
--- a/pom.xml
+++ b/pom.xml
@@ -328,6 +328,12 @@
             <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>mssqlserver</artifactId>
+            <version>1.20.0</version>
+            <scope>test</scope>
+        </dependency>
 
         <!-- common pom defines them in dependency management and hence they're pulled in compile scope.
         Explicitly define them as test to prevent from them form being packaged. -->
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -396,6 +396,6 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
         return rs.getString("TABLE_NAME");
       }
     }
-    return super.resolveSynonym(connection, synonymName);
+    return null;
   }
 }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -27,7 +27,9 @@ import io.confluent.connect.jdbc.util.TableDefinition;
 import java.io.ByteArrayInputStream;
 import java.io.StringReader;
 import java.nio.ByteBuffer;
+import java.sql.Connection;
 import java.sql.PreparedStatement;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
 import org.apache.kafka.common.config.AbstractConfig;
@@ -379,5 +381,22 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
       );
     }
     return tableId;
+  }
+
+  @Override
+  public String resolveSynonym(Connection connection, String synonymName) throws SQLException {
+    // Oracle-specific implementation using ALL_SYNONYMS view
+    try (PreparedStatement stmt = connection.prepareStatement(
+        "SELECT TABLE_OWNER, TABLE_NAME FROM ALL_SYNONYMS WHERE OWNER = ? AND " +
+        "SYNONYM_NAME = ?")) {
+      stmt.setString(1, connection.getMetaData().getUserName().toUpperCase());
+      stmt.setString(2, synonymName.toUpperCase());
+      ResultSet rs = stmt.executeQuery();
+      if (rs.next()) {
+        return rs.getString("TABLE_NAME");
+      }
+    }
+    // Fall back to generic implementation if ALL_SYNONYMS query fails
+    return super.resolveSynonym(connection, synonymName);
   }
 }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -385,7 +385,6 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
 
   @Override
   public String resolveSynonym(Connection connection, String synonymName) throws SQLException {
-    // Oracle-specific implementation using ALL_SYNONYMS view
     try (PreparedStatement stmt = connection.prepareStatement(
         "SELECT TABLE_OWNER, TABLE_NAME FROM ALL_SYNONYMS WHERE OWNER = ? AND "
         + "SYNONYM_NAME = ?")) {
@@ -397,7 +396,6 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
         return rs.getString("TABLE_NAME");
       }
     }
-    // Fall back to generic implementation if ALL_SYNONYMS query fails
     return super.resolveSynonym(connection, synonymName);
   }
 }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -389,8 +389,9 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
     try (PreparedStatement stmt = connection.prepareStatement(
         "SELECT TABLE_OWNER, TABLE_NAME FROM ALL_SYNONYMS WHERE OWNER = ? AND "
         + "SYNONYM_NAME = ?")) {
+      String tableName = parseTableIdentifier(synonymName).tableName();
       stmt.setString(1, connection.getMetaData().getUserName().toUpperCase());
-      stmt.setString(2, synonymName.toUpperCase());
+      stmt.setString(2, tableName.toUpperCase());
       ResultSet rs = stmt.executeQuery();
       if (rs.next()) {
         return rs.getString("TABLE_NAME");
--- a/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
@@ -387,8 +387,8 @@ public class OracleDatabaseDialect extends GenericDatabaseDialect {
   public String resolveSynonym(Connection connection, String synonymName) throws SQLException {
     // Oracle-specific implementation using ALL_SYNONYMS view
     try (PreparedStatement stmt = connection.prepareStatement(
-        "SELECT TABLE_OWNER, TABLE_NAME FROM ALL_SYNONYMS WHERE OWNER = ? AND " +
-        "SYNONYM_NAME = ?")) {
+        "SELECT TABLE_OWNER, TABLE_NAME FROM ALL_SYNONYMS WHERE OWNER = ? AND "
+        + "SYNONYM_NAME = ?")) {
       stmt.setString(1, connection.getMetaData().getUserName().toUpperCase());
       stmt.setString(2, synonymName.toUpperCase());
       ResultSet rs = stmt.executeQuery();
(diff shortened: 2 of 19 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -380,4 +382,22 @@ public TableId parseTableIdentifier(String fqn) {
     }
     return tableId;
   }
+
+  @Override
+  public String resolveSynonym(Connection connection, String synonymName) throws SQLException {
+    // Oracle-specific implementation using ALL_SYNONYMS view
+    try (PreparedStatement stmt = connection.prepareStatement(
+        "SELECT TABLE_OWNER, TABLE_NAME FROM ALL_SYNONYMS WHERE OWNER = ? AND "
+        + "SYNONYM_NAME = ?")) {
```
rajdangwal, line 391: Does it require additional privileges to query this view, which were not needed earlier?

src/main/java/io/confluent/connect/jdbc/dialect/SapHanaDatabaseDialect.java
```diff
@@ -157,4 +161,26 @@ public String buildUpsertQueryStatement(
     builder.append(" WITH PRIMARY KEY");
     return builder.toString();
   }
+
+  @Override
+  public String resolveSynonym(Connection connection, String synonymName) throws SQLException {
+    // SAP HANA supports synonyms through the SYNONYMS system view
+    try (PreparedStatement stmt = connection.prepareStatement(
+        "SELECT SCHEMA_NAME, TARGET_SCHEMA_NAME, TARGET_OBJECT_NAME "
```
rajdangwal, line 169: nit: We don't need to get the SCHEMA_NAME

src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java
```diff
@@ -349,4 +350,21 @@ private void transformUpdate(ExpressionBuilder builder, ColumnId col) {
            .append("=incoming.")
            .appendColumnName(col.name());
   }
+
+  @Override
+  public String resolveSynonym(Connection connection, String synonymName) throws SQLException {
+    // Sybase-specific implementation using sys.synonyms
+    try (PreparedStatement stmt = connection.prepareStatement(
+        "SELECT base_object_name FROM sys.synonyms WHERE name = ? AND "
+        + "schema_id = SCHEMA_ID()")) {
```
rajdangwal, line 359: Q: Does the `SCHEMA_ID()` resolves to the current schema name for which the connection is established?

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -91,6 +96,8 @@ public class SqlServerDatabaseDialect extends GenericDatabaseDialect {
 
   private boolean verifiedSqlServerTimestamp = false;
 
+  List<String> tableTypeConfig = config.getList(JdbcSourceConnectorConfig.TABLE_TYPE_CONFIG);
+
```
rajdangwal, line 100: We already have a set named `tableTypes` above. It can be reused.

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -568,4 +623,23 @@ protected String sanitizedUrl(String url) {
                 .replaceAll("(?i)(;keyStoreSecret=)[^;]*", "$1****")
                 .replaceAll("(?i)(;gsscredential=)[^;]*", "$1****");
   }
+
+  @Override
+  public String resolveSynonym(Connection connection, String synonymName) throws SQLException {
+    // Oracle-specific implementation using ALL_SYNONYMS view
+    try (PreparedStatement stmt =
+        connection.prepareStatement(
+            "SELECT PARSENAME(s.base_object_name, 2) AS TABLE_OWNER,"
```
rajdangwal, line 632: nit: We don't need the `TABLE_OWNER` or schema name here.

rajdangwal (commented):
Thanks @KGaneshDatta 
Took the initial pass and left few comments. PTAL.

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
KGaneshDatta: No extra privileges are needed. However, if access to the tables is governed by role‑based permissions, it’s better to provide the specific database username.

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
```diff
@@ -380,4 +382,22 @@ public TableId parseTableIdentifier(String fqn) {
...
+      String tableName = parseTableIdentifier(synonymName).tableName();
+      stmt.setString(1, connection.getMetaData().getUserName().toUpperCase());
+      stmt.setString(2, tableName.toUpperCase());
+      ResultSet rs = stmt.executeQuery();
+      if (rs.next()) {
+        return rs.getString("TABLE_NAME");
+      }
+    }
+    // Fall back to generic implementation if ALL_SYNONYMS query fails
+    return super.resolveSynonym(connection, synonymName);
```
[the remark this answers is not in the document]
KGaneshDatta: If the corresponding database dialect method does not find a match, the fallback prevents the method from returning null prematurely, allowing the base class to attempt resolution where this method although requires to be implemented.

src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java
KGaneshDatta: Updated the query as per the system tables supported by Sybase.

src/main/java/io/confluent/connect/jdbc/dialect/SapHanaDatabaseDialect.java
```diff
@@ -157,4 +161,26 @@ public String buildUpsertQueryStatement(
...
+        + "FROM SYS.SYNONYMS WHERE SCHEMA_NAME = ? AND SYNONYM_NAME = ?")) {
+      // Use the current schema if not specified
+      String tableName = parseTableIdentifier(synonymName).tableName();
+      String schema = connection.getMetaData().getUserName();
```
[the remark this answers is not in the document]
KGaneshDatta: We didn't the test Instance to explicitly test it. I have explored online through different resources and added the required query.

pom.xml
```diff
@@ -328,6 +328,12 @@
             <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>mssqlserver</artifactId>
+            <version>1.20.0</version>
+            <scope>test</scope>
+        </dependency>
```
KGaneshDatta, line 336: nit : Will remove it.

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
```diff
@@ -495,6 +499,54 @@ public void validateSpecificColumnTypes(
...
+        String tableName = rs.getString(3);
+        TableId tableId = new TableId(catalogName, schemaName, tableName);
+        if (includeTable(tableId)) {
+          tableIds.add(tableId);
+        }
+      }
+    }
+
+    // Get synonyms from sys.synonyms on a separate statement
+    if (this.tableTypes.contains("SYNONYM")) {
```
rajdangwal, line 524: I don't think it will work if someone specifies `synonym` instead of `SYNONYM` in the `table.types` config.
At line 505 we are getting the tableTypes in uppercase. So, it should use the local variable `tableTypes` here instead of the instance variable.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -98,6 +98,7 @@ public void start(Map<String, String> properties) {
     List<String> tables = config.getList(JdbcSourceTaskConfig.TABLES_CONFIG);
     Boolean tablesFetched = config.getBoolean(JdbcSourceTaskConfig.TABLES_FETCHED);
     String query = config.getString(JdbcSourceTaskConfig.QUERY_CONFIG);
+    List<String> tableType = config.getList(JdbcSourceConnectorConfig.TABLE_TYPE_CONFIG);
```
rajdangwal, line 101: We are getting the tableTypes directly from config and passing it to `validateColumnsExist` method below which is not using the passed value and again getting it via `config.getList(...)`.

It is passing the configured value to `describeColumnsForTables` method and it is performing this check: `tableType.contains("SYNONYM")`. We may need to call `tableTypes(metadata, this.tableTypes)` to get the values in Uppercase before doing the check.

src/test/java/io/confluent/connect/jdbc/source/integration/MSSqlServerTableIT.java
```diff
@@ -0,0 +1,106 @@
...
+          .withEnv("SA_PASSWORD", PASS)
+          .withFixedExposedPort(1433, 1433);
+
+  private final Map<String, String> props = new HashMap<>();
+  private Connection connection;
+
+  @Before
+  public void setup() throws Exception {
+    // Set up JDBC Driver
+    System.out.println("Setting up JDBC Driver");
```
rajdangwal, line 47: nit: Remove these.

rajdangwal (commented):
Thanks @KGaneshDatta
Left some comments.

src/main/java/io/confluent/connect/jdbc/dialect/SapHanaDatabaseDialect.java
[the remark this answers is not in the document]
rajdangwal: I would suggest to test this if possible.

Earlier the query was:

> "SELECT SCHEMA_NAME, TARGET_SCHEMA_NAME, TARGET_OBJECT_NAME "
        + "FROM SYS.SYNONYMS WHERE SCHEMA_NAME = ? AND SYNONYM_NAME = ?"))

The columns `TARGET_SCHEMA_NAME` and `TARGET_OBJECT_NAME` are not there in the [SYS.SYNONYMS](https://help.sap.com/docs/HANA_SERVICE_CF/7c78579ce9b14a669c1f3295b0d8ca16/210009ba75191014b061af7f6276da77.html) view.

And now it is:

> "SELECT OBJECT_NAME AS TARGET_SCHEMA_NAME "
                + "FROM SYS.SYNONYMS WHERE SYNONYM_NAME = ?"

Why do we need the alias for the OBJECT_NAME to be TARGET_SCHEMA_NAME?

src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java
rajdangwal: Now the query seems more complex. This should be verified IMO.

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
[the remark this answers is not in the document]
rajdangwal: Sorry, I am still not able to convince myself on why this fallback is required.
What are we achieving by this fall-back method?
Can you please describe a scenario where this fallback actually helps?

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
KGaneshDatta: As instead of referencing the config.getList(...) multiple times which could be the duplication of same code, I have updated the methods validateColumnsExist() and validateNullable(), both to have the argument tableType which is being passed.

Instead of calling the tableTypes() to get values in Uppercase, I am ignoring the case to check if the values contains synonym word in the list. I think here it is simply to check synonyms case insensitively and for that it this condition should be fine.

src/main/java/io/confluent/connect/jdbc/dialect/SqlServerDatabaseDialect.java
KGaneshDatta: Yes, agree. I have made the changes using the existing tabletypes variable and check the value in the String array.

src/main/java/io/confluent/connect/jdbc/dialect/SybaseDatabaseDialect.java
KGaneshDatta: Currently, I have removed the SAP HANA and Sybase from the support for SYNONYMS in the JDBC Connectors. Will update it once thoroughly tested with the instances.

src/main/java/io/confluent/connect/jdbc/dialect/OracleDatabaseDialect.java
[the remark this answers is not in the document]
KGaneshDatta: The fallback is added only when in the worst case to check with the regular flow. But that is not possible with the dialect based implementation. Removed the fallback for the custom implementation of resolveSynonyms method.

PR #1507, opened 2025-05-07
Add connectorname and Task ID for the threads

KGaneshDatta:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
@@ -63,7 +63,9 @@ public class TableMonitorThread extends Thread {
       long pollMs,
       Set<String> whitelist,
       Set<String> blacklist,
-      Time time
+      Time time,
+      String connectorName,
+      String taskId
   ) {
     this.dialect = dialect;
     this.connectionProvider = connectionProvider;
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java
```diff
@@ -75,6 +77,8 @@ public TableMonitorThread(DatabaseDialect dialect,
     this.blacklist = blacklist;
     this.tables = new AtomicReference<>();
     this.time = time;
+
+    this.setName(connectorName + "-" + taskId + "-TableMonitorThread")
```
skaushikcflt, line 81: who is the caller of this? I don't see where exactly are we creating a thread, if we are using executorService, it would be better to set the thread name over there.

KGaneshDatta: We are calling for creating new thread [here](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java#L126). We are not using any executorService in the JDBC Connector.

PR #1543, opened 2025-09-22
CC-36404: Add table include / exclude list regex and timestamp and incrementing column mapping

sangeet259:
- Introduces two new configs -- timestamp.columns.mapping and incrementing.column.mapping to specify the table regex to set of timestamp or incrementing columns to use in appropriate modes.
- Changed the include and exclude list behavior to first find tables which are included according to inclusion regexes and then filter out those excluded by exclusion regexes.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
@@ -46,6 +46,7 @@ import io.confluent.connect.jdbc.util.CachedConnectionProvider;
 import io.confluent.connect.jdbc.util.ColumnDefinition;
 import io.confluent.connect.jdbc.util.ColumnId;
 import io.confluent.connect.jdbc.util.RecordQueue;
+import io.confluent.connect.jdbc.util.TableCollectionUtils;
 import io.confluent.connect.jdbc.util.TableId;
 import io.confluent.connect.jdbc.util.Version;
 import io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig.TransactionIsolationMode;
@@ -66,6 +67,8 @@ public class JdbcSourceTask extends SourceTask {
   PriorityQueue<TableQuerier> tableQueue = new PriorityQueue<>();
   protected RecordQueue<SourceRecord> engine;
   private TableQuerierProcessor tableQuerierProcessor;
+  private final Map<String, String> tableToIncrCol = new HashMap<>();
+  private final Map<String, List<String>> tableToTsCols = new HashMap<>();
 
   public JdbcSourceTask() {
     this.time = Time.SYSTEM;
@@ -145,7 +148,7 @@ public class JdbcSourceTask extends SourceTask {
     );
     TableQuerier.QueryMode queryMode = !query.isEmpty() ? TableQuerier.QueryMode.QUERY :
                                        TableQuerier.QueryMode.TABLE;
-    List<String> tablesOrQuery = queryMode == TableQuerier.QueryMode.QUERY
+    final List<String> tablesOrQuery = queryMode == TableQuerier.QueryMode.QUERY
                                  ? Collections.singletonList(query) : tables;
 
     String mode = config.getString(JdbcSourceTaskConfig.MODE_CONFIG);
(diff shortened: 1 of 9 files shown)


src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -0,0 +1,535 @@
...
+    List<ConfigValue> configValues = new ArrayList<>(configValuesMap.values());
+    this.validationResult = new Config(configValues);
+  }
+
+
+  /**
+   * Perform validation and return the Config object with any validation errors.
+   */
+  public Config validate() {
+    try {
```
akanimesh7, line 59: Proceeed ahead only if there are no errors from validation done on lin 49

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -279,36 +312,133 @@ private void validateModeColumnRequirements() {
...
         .findFirst()
         .ifPresent(cv -> cv.addErrorMessage(errorMessage));
   }
+
+  // Helper methods for configuration access
+  
+  /**
+   * Get table whitelist configuration as a set.
+   */
+  private Set<String> getTableWhitelistSet() {
```
akanimesh7, line 478: All these functions are not needed. Move all this to ConnectorConfig class and use that method.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -279,36 +312,133 @@ private void validateModeColumnRequirements() {
...
+    List<String> whitelist = config.getList(JdbcSourceConnectorConfig.TABLE_WHITELIST_CONFIG);
+    return whitelist.isEmpty() ? null : new HashSet<>(whitelist);
```
akanimesh7, line 480: Conversion from empty set to null is a code smell. Nulls should be avoided completely.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -112,77 +117,90 @@ private void validateMultiConfigs() {
...
       String msg = JdbcSourceConnectorConfig.TABLE_EXCLUDE_LIST_CONFIG 
           + " cannot be used without " + JdbcSourceConnectorConfig.TABLE_INCLUDE_LIST_CONFIG
           + ". Exclude list only applies to tables that match the include list.";
       addConfigError(JdbcSourceConnectorConfig.TABLE_EXCLUDE_LIST_CONFIG, msg);
+      return false;
     }
 
     // Legacy and new configs are mutually exclusive
     boolean hasLegacyConfig = whitelistSet != null || blacklistSet != null;
     boolean hasNewConfig = includeListSet != null || excludeListSet != null;
```
akanimesh7, line 198: Only -- `includeListSet != null` ? Since you've already validated that excludelist not null and include null should not be ther

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -0,0 +1,314 @@
...
+        configsUsed.add(JdbcSourceConnectorConfig.TABLE_BLACKLIST_CONFIG);
+      }
+      if (includeListSet != null) {
+        configsUsed.add(JdbcSourceConnectorConfig.TABLE_INCLUDE_LIST_CONFIG);
+      }
+      if (excludeListSet != null) {
+        configsUsed.add(JdbcSourceConnectorConfig.TABLE_EXCLUDE_LIST_CONFIG);
+      }
+      
+      String msg = "Cannot use legacy whitelist/blacklist with new include/exclude lists. "
```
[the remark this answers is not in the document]
akanimesh7: @sangeet259 Please address this

akanimesh7:
The validations have got very complex.
Why can't we do it like this ? 

```
bool usingLegacyConfigs -- if config keys have any one of `legacyKeys`
bool usingNewConfigs -- if config keys have any one of `newKeys`

if using both 
   fail validation

if using none 
   fail validation

if using older config
   validation is complete

if using newer config
   validate whatever is done on snowflake source
```

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
sangeet259: I think for completeness and independence, the redundancy should be there. Otherwise changes in future to one becomes dependent on the other.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -0,0 +1,440 @@
...
+    if (usingLegacyConfigs && usingNewConfigs) {
+      return addConfigErrorsForLegacyAndNewConfigConflict();
+    }
+
+    if (!usingLegacyConfigs && !usingNewConfigs) {
+      return addConfigErrorsForNoConfigProvided();
+    }
+
+    if (usingLegacyConfigs) {
+      return true;
```
akanimesh7, line 143: nit -- a comment stating that this is done for backward compatibility

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    if (usingNewConfigs) {
```
akanimesh7, line 146: nit - this will always be true

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -0,0 +1,440 @@
...
+  protected boolean validatePluginSpecificNeeds() {
+    // Default implementation - no plugin-specific validation
+    // This can be overridden by specific connector implementations
+    return true;
+  }
+
+  /**
+   * Validate that mode-dependent column configurations are properly provided.
+   */
+  private boolean validateTsAndIncModeColumnRequirements() {
```
akanimesh7, line 301: Currently the functions in this also check for legacy configs. Let's remove that. We are only validating new configs. And if validation reaches here, it's definitely only new configs present.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -179,27 +182,69 @@ public void start(Map<String, String> properties) {
...
+
+    // Validate mapping configurations
+    if (config.modeUsesTimestampColumn() && !timestampColumnMappings.isEmpty()) {
+      // Convert string table names to TableId objects for validation
+      List<TableId> tableIds = tables.stream()
+          .map(dialect::parseTableIdentifier)
+          .collect(java.util.stream.Collectors.toList());
+      
+      TableCollectionUtils.validateEachTableMatchesExactlyOneRegex(
+          config.timestampColMappingRegexes(), tableIds, TableId::toUnquotedString,
```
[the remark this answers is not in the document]
sangeet259: This is required because every dialect has its own leading and trailing identifier character.

For example, for MySql, If you specify this - "table.include.list":".*new_users.*,.*new_orders.*" in connector configs, 

\`mydb\`.\`new_orders\`
\`mydb\`.\`new_users\`
\`mydb\`.\`new_users_timestamp\`

Similarly, there are different rules for different databases. Some have `[` while others have `\`


So we need to go via TableId unquotedString way to avoid any issues.

PR #1564, opened 2025-09-29
Make `addConfigError` protected

akanimesh7:
## Problem
The `addConfigError` was private. Hence subclasses didn't have ability to use it.

## Solution
Making it protected.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


Kalla-Shubham (approved):
lgtm

PR #1565, opened 2025-10-02
CC-36952 -- Add all modes IT using testcontainers

akanimesh7:
## Problem
Plugins using kafka-connect-jdbc as external dependency should be able to run a sanity test using real container.

## Solution
- Implemented an abstract IT which checks the behavior for all 4 modes. Individual plugins can just plug-in the testcontainers dependency and get a fully running IT.
- Exemplifies how to use this with implementing for postgres

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


sangeet259 (approved):
Great work 👏

PR #1572, opened 2025-10-07
Fix flakiness due to testTableLocksWithStreamingReads

akanimesh7:
## Problem
testTableLocksWithStreamingReads was flaky and lead to productivity dip

## Solution
Before the recent record queue enhancements, `poll` function was responsible to make the DB calls. This test relied on the old behaviour, hence before polling it expected no locks to be taken. 

Howeverm after change in implementation now querying DB part is done by some other thread, hence it's not tied to poll.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


Kalla-Shubham (approved):
lgtm

PR #1574, opened 2025-10-07
 Update AbstractJdbcSourceConnectorIT to use new configs

akanimesh7:
## Problem
`table.whitelist`, `table.blacklist`, `incrementing.column.name` and `timestamp.column.name` are deprecated
 and `table.include.list`, `table.exclude.list`, `incrementing.column.mapping` or `timestamp.column.mappings` should be the ones used.

## Solution
Updates AbstractJdbcSourceConnectorIT to use the newer configs.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


Kalla-Shubham (approved):
lgtm

Issue #1591, opened 2025-11-07
10.8.5 requires configuration that breaks workers that use `query`


whit0694:
When deploying a JDBC Source worker after upgrading to 10.8.5 I got this error:

```
At least one table filtering configuration is required. Provide one of: table.whitelist, table.blacklist, table.include.list, or table.exclude.list.
```

I believe this was added in https://github.com/confluentinc/kafka-connect-jdbc/pull/1543

But if an option like `table.include.list` is provided, then source workers that use the `query` param throw this error

```
org.apache.kafka.connect.errors.ConnectException: query may not be combined with whole-table copying settings
```

That error is coming from https://github.com/confluentinc/kafka-connect-jdbc/blob/5c46c3a81cec9ef1d73cc8d23d8de0638ada1516/src/main/java/io/confluent/connect/jdbc/JdbcSourceConnector.java#L128

So it seems that with 10.8.5 I can only deploy a `query` source worker with configuration that breaks my query source worker.

sangeet259:
Fixed with https://github.com/confluentinc/kafka-connect-jdbc/pull/1593

PR #1593, opened 2025-11-12
Extend query config with query.masked config with Type.PASSWORD for secure query masking

KGaneshDatta:
## Problem
- Introduced an internal query.masked configuration of `Type.Password` to ensure secure logging by masking sensitive values and preventing accidental exposure.
- Enhanced the existing validation to enforce that the connector operates exclusively in either `table.mode` or `query.mode`.
- Fixed the breaking connector validations in query mode with the new tableFilteringOptions introduced in 10.8.5. [Issue](https://github.com/confluentinc/kafka-connect-jdbc/issues/1591)

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/source/TableQuerierProcessor.java
+++ b/src/main/java/io/confluent/connect/jdbc/source/TableQuerierProcessor.java
@@ -133,30 +133,30 @@ public class TableQuerierProcessor {
 
   private void handleNonTransientException(RecordDestination<SourceRecord> destination, 
                                            TableQuerier querier, SQLNonTransientException sqle) {
-    SQLException trimmedException = shouldRedactSensitiveLogs
+    SQLException redactedException = shouldRedactSensitiveLogs
               ? LogUtil.redactSensitiveData(sqle) : sqle;
     log.error("Non-transient SQL exception while running query for table: {}",
-        querier, trimmedException);
+        querier, redactedException);
     resetAndRequeueHead(querier, true);
     // This task has failed, report failure to destination
-    destination.failWith(new ConnectException(trimmedException));
+    destination.failWith(new ConnectException(redactedException));
   }
 
   private void handleSqlException(RecordDestination<SourceRecord> destination,
                                   TableQuerier querier, SQLException sqle) {
-    SQLException trimmedException = shouldRedactSensitiveLogs
+    SQLException redactedException = shouldRedactSensitiveLogs
               ? LogUtil.redactSensitiveData(sqle) : sqle;
     log.error(
         "SQL exception while running query for table: {}." + " Attempting retry {} of {} attempts.",
         querier,
         querier.getAttemptedRetryCount() + 1,
         maxRetriesPerQuerier,
-        trimmedException);
+        redactedException);
 
     resetAndRequeueHead(querier, false);
     if (maxRetriesPerQuerier > 0 && querier.getAttemptedRetryCount() >= maxRetriesPerQuerier) {
       destination.failWith(
-          new ConnectException("Failed to query table after retries", trimmedException));
+          new ConnectException("Failed to query table after retries", redactedException));
       return;
     }
     querier.incrementRetryCount();


src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -825,6 +846,13 @@ private static void addSchemaAndDialectOptions(ConfigDef config, int orderInGrou
 
   private static final void addModeOptions(ConfigDef config) {
     int orderInGroup = 0;
+    orderInGroup = defineModeConfig(config, orderInGroup);
+    orderInGroup = defineIncrementTimestampConfigs(config, orderInGroup);
+    orderInGroup = defineQueryAndQuoteConfigs(config, orderInGroup);
+    defineTransactionAndRetryConfigs(config, orderInGroup);
+  }
```
KGaneshDatta: Extend this Options method, because of falling to checkstyle error of exceeding 128 lines in a method.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -305,6 +325,48 @@ private boolean validateTableInclusionConfigs() {
...
+
+      log.error("Validation failed: Both query and query.masked configs are set");
+      return false;
+    }
+
+    if ((hasQuery || hasQueryMasked) && (isUsingLegacyConfigs() || isUsingNewConfigs())) {
+      String msg =
+          "Do not specify table filtering configs with 'query' or 'query.masked'. "
+              + "Remove table.whitelist / table.blacklist / table.include.list / "
+              + "table.exclude.list.";
```
KGaneshDatta: Either of table.filtering.options or query should be provided as the connector would work in `table.mode` or `query.mode` and not both.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -150,6 +152,16 @@ private boolean validateMultiConfigs() {
     return true;
   }
 
+  private boolean hasAnyQueryConfig() {
+    String query = config.getString(JdbcSourceConnectorConfig.QUERY_CONFIG);
+    Password queryMasked =
+        config.getPassword(JdbcSourceConnectorConfig.QUERY_MASKED_CONFIG);
```
akanimesh7: If you make `config.getQuery` return optional, then this `hasAnyQueryConfig` method will not be needed.

akanimesh7 (commented):
Few things @KGaneshDatta 
1. Since query was never exposed on cloud before this, we should perform some manual testing to understand/validate the behaviour. (How frequently will the query be executed, What if the query is taking too long to exceute ? do we have any observability into this ? Can multiple queries be provided (`select * from table1;select* from table2;`))
3. If user has specified both table include/exclude list and query, we don't know which way they wanna go. The validation exception message currently instructs them to remove table include/exclude configs. Which might be other way round also.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -427,6 +431,20 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
   private static final EnumRecommender QUOTE_METHOD_RECOMMENDER =
       EnumRecommender.in(QuoteMethod.values());
 
+  private static final ConfigDef.Recommender HIDDEN_RECOMMENDER =
```
akanimesh7: Add a comment explaining why it is hidden

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -1439,6 +1495,33 @@ public List<String> incrementingColMappingRegexes() {
         .collect(java.util.stream.Collectors.toList());
   }
 
+  /**
+ * Get the query string from either query or query.masked config.
+ * Prioritizes query.masked over query if both are set (though validation should prevent this).
```
akanimesh7: If there is a validation in place, there should never be a situation where both are set. Update the comment.

src/main/java/io/confluent/connect/jdbc/source/TableQuerierProcessor.java
```diff
@@ -198,5 +214,9 @@ private synchronized void resetAndRequeueHead(TableQuerier expectedHead, boolean
     }
   }
 
-
+  private String getQuerierLogString(TableQuerier querier) {
+    return shouldTrimSensitiveLogs
+        ? LogUtil.sensitiveLog(true, querier.toString())
```
akanimesh7: simply return `LogUtil.sensitiveLog(shouldTrimSensitiveLogs, querier.toString())`

pom.xml
```diff
@@ -110,6 +110,11 @@
             <groupId>com.google.re2j</groupId>
             <artifactId>re2j</artifactId>
             <version>1.7</version>
+        </dependency>
+        <dependency>
+          <groupId>org.apache.commons</groupId>
```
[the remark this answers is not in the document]
KGaneshDatta: This has been part of this PR - https://github.com/confluentinc/kafka-connect-jdbc/pull/1595. Moved it and removed it.

KGaneshDatta:
> Missing validation -- If mode is query and timestamp.initial is provided validation passes.

The query mode would be working in timestamp criteria also. More details here - https://docs.confluent.io/kafka-connectors/jdbc/current/source-connector/overview.html#specifying-a-where-clause-with-query-modes 
This also has been tested on CCloud with timestamp mode. Hence, when query was working in timestamp mode, I feel it would be valid to configure `timestamp.initial` when it is used.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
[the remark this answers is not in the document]
KGaneshDatta: While in query mode also, we are capturing the offsets which are stored as in this format. 

`{
  "partition": {
    "query": "query"
  },
  "offset": {
    "timestamp_nanos": 379680000,
    "timestamp": 1764041177379
  }
`

`{
  "partition": {
    "query": "query"
  },
  "offset": {
    "incrementing": 4
  }
`

So, I think this method would be executed in query mode also. WDYT ?

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -223,85 +231,151 @@ private boolean addConfigErrorsForLegacyAndNewConfigConflict() {
...
+
     return false;
   }
-  
+
   /**
    * Validate that at least one configuration is provided.
    */
   private boolean addConfigErrorsForNoConfigProvided() {
+    if (config.getQuery().isPresent()) {
```
akanimesh7: Due to check on line 166 this code is redundant

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -442,4 +516,30 @@ protected void addConfigError(String configName, String errorMessage) {
...
   }
 
-}
+  /** Validate that provided query strings start with a SELECT statement. */
+  private boolean validateSelectStatement(String statement, String configKey) {
+    String trimmedStatement = statement.trim();
+    if (!SELECT_STATEMENT_PATTERN.matcher(trimmedStatement).find()) {
+      String msg =
+          String.format(
+              "Only SELECT statements are supported for '%s'. Please provide "
```
akanimesh7: This will print the message on UI with `query.masked`. However, the exposed config is `query` both on CC and CP. Better to just use `query` everywhere being agnostic of the specific type of query config that is being used.

src/test/java/io/confluent/connect/jdbc/source/TableQuerierTest.java
```diff
@@ -82,7 +84,7 @@ public void testTimestampIncrementingTableQuerierInTableModeWithSuffix() throws
   @Test
   public void testTimestampIncrementingTableQuerierInQueryModeWithSuffix() throws SQLException {	    
     TimestampIncrementingTableQuerier querier = new TimestampIncrementingTableQuerier(
-                                                    databaseDialectMock,
```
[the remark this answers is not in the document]
KGaneshDatta: Somehow this extra whitespace is not getting added. Hence, left it as it is.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
akanimesh7: Yes, on cloud it's unlikely. But still let's make sure that all validation error messages use "query" and not "query.masked"

akanimesh7 (commented):
trim.sensitive.log is a config which was already existing in the sink connectors. 
Now we are using the same config in source connectors. Although we have defined it in the ConfigDef, but there is no way to set a value for this config. Please check this.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -442,4 +516,30 @@ protected void addConfigError(String configName, String errorMessage) {
         .ifPresent(cv -> cv.addErrorMessage(errorMessage));
   }
 
-}
+  /** Validate that provided query strings start with a SELECT statement. */
+  private boolean validateSelectStatement(String statement, String configKey) {
```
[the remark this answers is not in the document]
KGaneshDatta: Yes, I have removed `query.masked` in all the validation messages and made it clear to show only `query` config.

src/main/java/io/confluent/connect/jdbc/source/TableQuerierProcessor.java
```diff
@@ -131,28 +133,30 @@ private void processQuerier(RecordDestination<SourceRecord> destination, TableQu
 
   private void handleNonTransientException(RecordDestination<SourceRecord> destination, 
                                            TableQuerier querier, SQLNonTransientException sqle) {
+    SQLException trimmedException = shouldTrimSensitiveLogs
+              ? LogUtil.trimSensitiveData(sqle) : sqle;
```
akanimesh7: should this be redact function call ?

PR #1599, opened 2025-12-22
Add SqlParser Utility to parse and validate SQL query

KGaneshDatta:
## Problem
Currently, the connector lacks validation for configured SQL queries, allowing syntax errors to pass through the initial configuration phase. Additionally, when query.masked is enabled, the SQL queries are entirely redacted in the logs, which significantly hinders troubleshooting and limits visibility into the execution context.

## Solution
1. Implemented a SqlParser Utility  (integrating [JSQLParser](https://github.com/JSQLParser/JSqlParser)) to validate the syntax of user-defined SQL queries during the connector validation phase.
2. Enhanced the utility to perform structural parsing, allowing for the selective redaction of sensitive literals while preserving the SQL structure. This ensures PII (Personally Identifiable Information) protection without logging.
3. Introduced additional metadata and contextual parameters like `sqlState`, `errorCodes` to logs to enhance debuggability for SQLExceptions, even when `query.masked` is enabled.

- [ ] yes
- [ ] no

## Test Strategy
- Added relevant UTs
- End to End testing in local Connect cluster using Confluent Platform

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/test/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidationTest.java
+++ b/src/test/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidationTest.java
@@ -836,7 +836,7 @@ public class JdbcSourceConnectorValidationTest {
 
     assertErrors(1);
     assertErrors(QUERY_CONFIG, 1);
-    assertErrorMatches(QUERY_CONFIG, ".*Only SELECT statements are supported for 'query'.*");
+    assertErrorMatches(QUERY_CONFIG, ".*Only SELECT statements are supported for query.*");
   }
 
   @Test
@@ -850,7 +850,7 @@ public class JdbcSourceConnectorValidationTest {
     assertErrors(QUERY_CONFIG, 1);
     assertErrorMatches(
         QUERY_CONFIG,
-        "Only SELECT statements are supported for 'query'"
+        "Only SELECT statements are supported for query"
     );
   }
 
@@ -865,7 +865,7 @@ public class JdbcSourceConnectorValidationTest {
     assertErrors(QUERY_CONFIG, 1);
     assertErrorMatches(
         QUERY_CONFIG,
-        ".*Invalid SQL syntax for 'query'.*"
+        ".*Invalid SQL syntax for query.*"
     );
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/jdbc/source/TableQuerier.java
```diff
@@ -188,6 +189,23 @@ protected void recordQuery(String query) {
...
+   * which is useful for troubleshooting during exceptions.
+   *
+   * @return the query with sensitive data redacted, or null if no query is available
+   */
+  public String getParsedQueryString() {
+    if (loggedQueryString != null) {
+      return SqlParser.redactSensitiveData(loggedQueryString);
+    }
+    if (query != null) {
+      return SqlParser.redactSensitiveData(query);
```
akanimesh7: Both the inputs to `redactSensitiveData` here should not have the original data. They should have only `?` in them. Can you confirm this ? 

If this is the case why do we even need to redact them ?

akanimesh7 (commented):
@KGaneshDatta One comment

src/main/java/io/confluent/connect/jdbc/source/TableQuerier.java
KGaneshDatta: No, when the query config is provided, the variable query & loggedQueryString would be having the configured query value only. Please find the below screenshot where I have added the logs in the immediately in the corresponding line numbers and also have confirmed from the exception being thrown whether the value is redacted or not.

String Redaction :- 
Screenshot 2026-01-06 at 3 26 35 PM

Number Redaction :-
image

akanimesh7: Got it. The intention here is to mask the values in where clause that the customer provided query might have. The `?` thing is true only for the where clause we add ourself.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -500,6 +502,18 @@ private boolean validateSelectStatement(String statement, String configKey) {
       log.error(msg);
       return false;
     }
+    try {
+      SqlParser.validateSqlSyntax(trimmedStatement);
+    } catch (ConfigException e) {
+      String msg =
+          String.format(
+              "Invalid SQL syntax for '%s'. Please provide a syntactically correct "
```
akanimesh7: This will print `query.masked`. Which we don't want.
Please ensure only `query` is what is present on all user facing things.

akanimesh7 (commented):
Currently there are multiple versions of redaction -- ****** or 0 or <redacted>. How about unifying this ? Why can't we use <redacted> or similar everywhere ?

pom.xml
```diff
@@ -63,6 +63,7 @@
         <jtds.driver.version>1.3.1</jtds.driver.version>
         <slf4j.version>1.7.36</slf4j.version>
         <reload4j.version>1.2.19</reload4j.version>
+        <jsqlparser.version>4.9</jsqlparser.version>
```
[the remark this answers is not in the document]
KGaneshDatta: JSQLParser 5.0+ requires at least Java 11, and 5.1+ requires a Java 17 toolchain - [REF](https://github.com/JSQLParser/JSqlParser?tab=readme-ov-file#java-version). The kafka-connect-jdbc project is currently configured to use `Java 8` hence was using the last supported version with `java8`

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -500,6 +502,18 @@ private boolean validateSelectStatement(String statement, String configKey) {
       log.error(msg);
       return false;
     }
+    try {
+      SqlParser.validateSqlSyntax(trimmedStatement);
+    } catch (ConfigException e) {
```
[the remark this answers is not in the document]
akanimesh7: Ok. This is confusing because of ConfigException. 

You should catch the `JSQLParserException` here and add `throws JSQLParserException` to the definition of `validateSqlSyntax`

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
akanimesh7: I still see configKey. Is it done ?

PR #1626, opened 2026-04-20
Attach connection error to CONNECTION_URL

pgajendran-confluent:
## Problem
JdbcSinkConnectorValidation.validateConnection() catches database connection errors (e.g., SSL failures) but silently drops them because it tries to attach the error to connection.host, which is not defined in JdbcSinkConfig.CONFIG_DEF.

connection.host is a connect-templates variable, not a native kafka-connect-jdbc config. The configValue() helper returns Optional.empty() and the ifPresent callback never fires — the error is logged but never added to the validation result.

## Solution
Attach the connection error to CONNECTION_URL (connection.url) instead of CONNECTION_HOST (connection.host), since connection.url IS defined in CONFIG_DEF.

- [ ] yes
- [X] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/validation/JdbcSinkConnectorValidation.java
+++ b/src/main/java/io/confluent/connect/jdbc/validation/JdbcSinkConnectorValidation.java
@@ -32,7 +32,7 @@ import java.util.Optional;
 
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.DELETE_ENABLED;
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.PK_MODE;
-import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.CONNECTION_HOST;
+import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.CONNECTION_URL;
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.PrimaryKeyMode.RECORD_KEY;
 
 public class JdbcSinkConnectorValidation {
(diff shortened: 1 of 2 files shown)


arihant-confluent (commented):
LGTM. 
Are there any existing UTs to test it?

pgajendran-confluent:
> LGTM. Are there any existing UTs to test it?

Thanks. No UTs currently to test it though.

sangeet259:
Can you raise the PR against the earliest feature branch in which the issue is present.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSinkConnectorValidation.java
```diff
@@ -32,14 +32,14 @@
 
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.DELETE_ENABLED;
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.PK_MODE;
-import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.CONNECTION_HOST;
+import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.CONNECTION_URL;
 import static io.confluent.connect.jdbc.sink.JdbcSinkConfig.PrimaryKeyMode.RECORD_KEY;
```
sangeet259, line 36: After we removed the usage of `CONNECTION_HOST`, is it a dead code or is it used anywhere else? If it is not used anywhere else, we can remove it as well.

pgajendran-confluent: It is removed, the diff here shows that `CONNECTION_HOST` is removed and `CONNECTION_URL` is added.

sangeet259: I am talking about the variable itself, `CONNECTION_HOST`. After you removed this from here, is there any other place where that variable is used ?

PR #1627, opened 2026-04-21
Attach connection error to CONNECTION_URL and add tests

pgajendran-confluent:
## Problem
JdbcSinkConnectorValidation.validateConnection() catches database connection errors (e.g., SSL failures) but silently drops them because it tries to attach the error to connection.host, which is not defined in JdbcSinkConfig.CONFIG_DEF.

connection.host is a connect-templates variable, not a native kafka-connect-jdbc config. The configValue() helper returns Optional.empty() and the ifPresent callback never fires — the error is logged but never added to the validation result.

## Solution
Attach the connection error to CONNECTION_URL (connection.url) instead of CONNECTION_HOST (connection.host), since connection.url IS defined in CONFIG_DEF.

- [ ] yes
- [X] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


arihant-confluent (dismissed):
LGTM, thanks

arihant-confluent (approved):

PR #1628, opened 2026-04-27
Update Query Validation to use LIMIT clause instead of EXPLAIN PLAN mechanism

KGaneshDatta:
## Problem
In this PR - https://github.com/confluentinc/kafka-connect-jdbc/pull/1616, introduced a query validation using `EXPLAIN PLAN <query>` mechanism. But this mechanism would require an extra privileges for the user on the EXPLAIN table which is not a better approach. 

## Solution
Replaces per-dialect `EXPLAIN / EXPLAIN PLAN` FOR query validation with a single, dialect-agnostic JDBC probe based on PreparedStatement.getMetaData(). The user's query is sent to the database unchanged; if it has a syntax error, references a missing table or column, or fails a permission check, the driver throws SQLException and the connector reports a validation error before the task starts.

`PreparedStatement.getMetaData()` asks the JDBC driver to return the column descriptors for the query. To do so, every supported driver has the database compile the query — parse, resolve objects, type-check, verify SELECT permission — and return the describe. Rows are not fetched.Any error like bad syntax, missing table, no permission) surfaces with validation error and corresponding SQL Codes.

Other alternative rejected approaches
- `EXPLAIN PLAN` — no special privileges, no PLAN_TABLE setup, no plan-table writes, and one implementation instead of four.
- wrapping the query as `SELECT * FROM (<query>) WHERE 1=0` — the wrap rewrites the user's SQL into a derived table, which is illegal for several legal top-level forms (ORDER BY on SQL Server, SELECT * over JOINs producing duplicate columns, FOR UPDATE, CTE-only queries). Passing the query through unmodified eliminates these false positives.
- `setMaxRows(1)` + `executeQuery()` — opens a result set and fetches a row; getMetaData() does not. Strictly cheaper or equal on every driver.
- prepareStatement() alone — several drivers prepare lazily on the client and never contact the server, so missing tables and columns slip through. `getMetaData()` forces the round trip.

- [ ] yes
- [ ] no

## Test Strategy
Tested it on local CP environment by deploying the changes for the JDBC Source connectors.

Verified end-to-end against PostgreSQL, MySQL, SQL Server, Oracle and IBM DB2 on the systest cluster. The matrix per dialect covers valid SELECT *, missing table, missing column, top-level ORDER BY, GROUP BY with unaliased COUNT(*), subquery in WHERE, self-JOIN with duplicate columns, incomplete predicates, and aliased projection with ORDER BY plus LIMIT/TOP/FETCH FIRST. All five dialects reject missing tables and columns; all five accept the legal forms that previously failed under the wrap-based probe.

<details>
<summary>Postgres JDBC Source</summary>

**Fail Fast Validation Error Message on Configuration Step**
Screenshot 2026-05-04 at 4 11 55 PM

**Even fails with semi-colon at the end of query**
Screenshot 2026-05-04 at 4 12 14 PM

**Successful Provisioning with correct query without any validation error**
Screenshot 2026-05-04 at 4 12 34 PM
</details>

<details>
<summary>MySQL JDBC Source</summary>

**Fail Fast Validation Error Message**
Screenshot 2026-05-04 at 4 17 14 PM

**Successful Provisioning with correct query without any validation error**
Pasted Graphic 2

</details>

<details>
<summary>Oracle JDBC Source</summary>

**Fail Fast Validation Error Message**
Screenshot 2026-05-04 at 4 22 35 PM

**Successful Provisioning with correct query without any validation error**
Screenshot 2026-05-04 at 4 32 39 PM

</details>

<details>
<summary>MSSQL Server JDBC Source</summary>

**Fail Fast Validation Error Message**
Screenshot 2026-05-04 at 4 30 21 PM

**Successful Provisioning with correct query without any validation error**
Screenshot 2026-05-04 at 4 58 08 PM

</details>

<details>
<summary>IBM DB2 JDBC Source</summary>

**Fail Fast Validation Error Message**
Screenshot 2026-05-05 at 5 39 16 PM

**Successful Provisioning with correct query without any validation error**
Screenshot 2026-05-05 at 5 39 47 PM

</details> 

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -707,9 +707,16 @@ public interface DatabaseDialect extends ConnectionProvider {
   String resolveSynonym(Connection connection, String synonymName) throws SQLException;
 
   /**
-   * Validate the query with a low-cost ANSI probe that wraps it as a derived table
-   * under a constant-false predicate. The default below is a {@code prepareStatement}
-   * fallback for custom dialects that do not extend {@code GenericDatabaseDialect}.
+   * Validate the user-supplied SQL by asking the driver to describe its result-set
+   * shape via {@link PreparedStatement#getMetaData()}. Every supported driver
+   * services this call by having the database compile the query (parse, resolve
+   * objects, check types, verify {@code SELECT} permission), so any problem
+   * surfaces as a {@link SQLException} without the rows being scanned.
+   *
+   * <p>The user query is passed through unmodified, which avoids the false
+   * positives that wrap-based probes suffer from (top-level {@code ORDER BY}
+   * on SQL Server, duplicate columns from {@code SELECT *} over JOINs,
+   * unaliased computed columns, {@code FOR UPDATE}, and so on).
    *
    * @param connection the database connection; may not be null
    * @param query      the SQL query to validate; may not be null
@@ -718,7 +725,7 @@ public interface DatabaseDialect extends ConnectionProvider {
    */
   default void validateQuery(Connection connection, String query) throws SQLException {
     try (PreparedStatement stmt = connection.prepareStatement(query)) {
-      // Preparing surfaces syntax errors as SQLException; no execution needed.
+      stmt.getMetaData();
     }
   }
 }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
@@ -707,10 +707,9 @@ public interface DatabaseDialect extends ConnectionProvider {
   String resolveSynonym(Connection connection, String synonymName) throws SQLException;
 
   /**
-   * Validate the query with a low-cost limit-1 probe. Each bundled dialect wraps the
-   * query and applies its native limit syntax ({@code LIMIT}, {@code TOP}, or
-   * {@code FETCH FIRST}); the default below is a {@code prepareStatement} fallback for
-   * custom dialects that do not extend {@code GenericDatabaseDialect}.
+   * Validate the query with a low-cost ANSI probe that wraps it as a derived table
+   * under a constant-false predicate. The default below is a {@code prepareStatement}
+   * fallback for custom dialects that do not extend {@code GenericDatabaseDialect}.
    *
    * @param connection the database connection; may not be null
    * @param query      the SQL query to validate; may not be null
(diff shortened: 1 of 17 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -707,20 +707,25 @@ interface ColumnConverter {
...
   /**
-   * Validate the given SQL query against the database without executing it.
-   * This method verifies that the query is valid by checking table and column existence,
-   * user permissions, and SQL correctness using database-specific mechanisms such as
-   * {@code EXPLAIN} or {@code prepareStatement}.
+   * Validate the user-supplied SQL by asking the driver to describe its result-set
+   * shape via {@link PreparedStatement#getMetaData()}. Every supported driver
+   * services this call by having the database compile the query (parse, resolve
+   * objects, check types, verify {@code SELECT} permission), so any problem
+   * surfaces as a {@link SQLException} without the rows being scanned.
```
rpindikura, line 714: > without the rows being scanned.

This is driver specific & so, we should remove this claim. eg. MySQL seems to internally do a "limit 1" [here](https://github.com/mysql/mysql-connector-j/blob/7d6b0800528b6b25c68b52dc10d6c1c8429c100c/src/main/user-impl/java/com/mysql/cj/jdbc/ClientPreparedStatement.java#L1151)

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
```diff
@@ -707,20 +707,25 @@ interface ColumnConverter {
...
-   * @throws SQLException if the query is invalid, references non-existent tables or columns,
-   *                      or the user lacks necessary permissions
+   * @throws SQLException if the query is invalid, references missing tables or columns,
+   *                      or the user lacks {@code SELECT} permission
    */
   default void validateQuery(Connection connection, String query) throws SQLException {
     try (PreparedStatement stmt = connection.prepareStatement(query)) {
-        // Validate the query by preparing it. If the query is invalid,
-        // an SQLException will be thrown which will indicate the validation failure.
+      stmt.getMetaData();
```
rpindikura, line 728: if the user gave a non-select statement (eg. SHOW TABLES), this will not throw an exception. Is that ok?

should we instead check for null/empty ?

```suggestion
      ResultSetMetaData md = stmt.getMetaData();
          if (md == null || md.getColumnCount() == 0) {
              throw new SQLException(
                  "Query did not return result-set metadata; expected a SELECT statement");
          }

```

rpindikura (commented):
can you pls test (manual is fine) the following integration cases

- Valid SELECT ✓
- SELECT with non-existent table → SQLException ✓
- SELECT against table with no permission → SQLException ✓
- DELETE FROM ... or SHOW TABLES → must fail validation
- SELECT with ORDER BY and SELECT * over JOIN (the wrap regression cases) → should pass ✓

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialect.java
rpindikura, line 728: should we add a timeout too?

`stmt.setQueryTimeout(30)`

whit0694:
This addresses #1633, right?

PR #1632, opened 2026-05-07
Add primitive value schema support in JDBC sink connector to allow StringConverter values to be written directly as VARCHAR

KGaneshDatta:
## Problem
The JDBC Sink Connector currently rejects every record whose value schema is not STRUCT. `RecordValidator.requiresValue()` enforces this with a hard schema-type check and raises a ConnectException for anything else — including the `Schema.STRING` records produced by StringConverter. As a result, records carrying raw String payloads (logs, raw JSON documents, plain text) cannot be written to a JDBC target unless they are first transformed upstream into a Struct.

## Solution
Extend the connector to accept records whose value schema is `Schema.STRING`. The `STRUCT` path is left entirely unchanged.

For a STRING record:
  - The value is written into a single column whose name is controlled by a new config, `string.output.value.column.name` (default `recordValue`).
  - When the destination table does not exist, auto.create provisions the table with this column as NULLABLE.
  - When the table exists but the column is missing, `auto.evolve=true` adds it via `ALTER TABLE … ADD COLUMN` (also NULLABLE, so the `ALTER` does not violate constraints on pre-existing rows). With `auto.evolve=false`, the connector raises the standard auto evolution disabled error — no new failure path.
  - The value is bound to the configured column via `PreparedStatement.setString(...)` regardless of database dialect.

Two configuration combinations are rejected up front with explicit, actionable error messages:
  - `pk.mode=record_value` with a STRING value — a String has no fields to designate as a primary key.
  - `string.output.value.column.name` colliding with a primary-key column name

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -267,6 +267,19 @@ public class JdbcSinkConfig extends AbstractConfig {
       + "specific dialect. All properly-packaged dialects in the JDBC connector plugin "
       + "can be used.";
 
+  public static final String STRING_OUTPUT_VALUE_COLUMN_NAME = "string.output.value.column.name";
+  public static final String STRING_OUTPUT_VALUE_COLUMN_NAME_DEFAULT = "recordValue";
+  private static final String STRING_OUTPUT_VALUE_COLUMN_NAME_DOC =
+      "When the Kafka record value uses the String schema (e.g. records produced with "
+      + "StringConverter), the connector writes the value into a single column with this name. "
+      + "If the destination table is being auto-created, the column is created with this name. "
+      + "If the table already exists and the column is missing, the column is added via ALTER "
+      + "when 'auto.evolve' is true; otherwise the connector fails with the standard "
+      + "auto-evolution-disabled error. Defaults to '"
+      + STRING_OUTPUT_VALUE_COLUMN_NAME_DEFAULT + "'.";
+  private static final String STRING_OUTPUT_VALUE_COLUMN_NAME_DISPLAY =
+      "String Value Column Name";
+
   public static final String DB_TIMEZONE_CONFIG = "db.timezone";
   public static final String DB_TIMEZONE_DEFAULT = "UTC";
   private static final String DB_TIMEZONE_CONFIG_DOC =
@@ -606,6 +619,18 @@ public class JdbcSinkConfig extends AbstractConfig {
             ConfigDef.Width.MEDIUM,
             DATE_CALENDAR_SYSTEM_DISPLAY
         )
+        .define(
+            STRING_OUTPUT_VALUE_COLUMN_NAME,
+            ConfigDef.Type.STRING,
+            STRING_OUTPUT_VALUE_COLUMN_NAME_DEFAULT,
+            new ConfigDef.NonEmptyString(),
+            ConfigDef.Importance.LOW,
+            STRING_OUTPUT_VALUE_COLUMN_NAME_DOC,
+            DATAMAPPING_GROUP,
+            10,
+            ConfigDef.Width.MEDIUM,
+            STRING_OUTPUT_VALUE_COLUMN_NAME_DISPLAY
+        )
         // DDL
         .define(
             AUTO_CREATE,
@@ -700,6 +725,7 @@ public class JdbcSinkConfig extends AbstractConfig {
   public final PrimaryKeyMode pkMode;
   public final List<String> pkFields;
   public final Set<String> fieldsWhitelist;
+  public final String stringOutputValueColumnName;
   public final Set<String> timestampFieldsList;
   public final String dialectName;
   public final ZoneId zoneId;
@@ -733,6 +759,7 @@ public class JdbcSinkConfig extends AbstractConfig {
     pkFields = getList(PK_FIELDS);
     dialectName = getString(DIALECT_NAME_CONFIG);
     fieldsWhitelist = new HashSet<>(getList(FIELDS_WHITELIST));
+    stringOutputValueColumnName = getString(STRING_OUTPUT_VALUE_COLUMN_NAME);
     String dbTimeZone = getString(DB_TIMEZONE_CONFIG);
     zoneId = ZoneId.of(dbTimeZone);
     DateTimezone dateTimezoneConfig =
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
@@ -268,7 +268,7 @@ public class JdbcSinkConfig extends AbstractConfig {
       + "can be used.";
 
   public static final String STRING_OUTPUT_VALUE_COLUMN_NAME = "string.output.value.column.name";
-  public static final String STRING_OUTPUT_VALUE_COLUMN_NAME_DEFAULT = "recordValue";
+  public static final String STRING_OUTPUT_VALUE_COLUMN_NAME_DEFAULT = "record_value";
   private static final String STRING_OUTPUT_VALUE_COLUMN_NAME_DOC =
       "When the Kafka record value uses the String schema (e.g. records produced with "
       + "StringConverter), the connector writes the value into a single column with this name. "
--- a/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
@@ -108,11 +108,11 @@ public class PreparedStatementBinder implements StatementBinder {
         case INSERT:
         case UPSERT:
           index = bindKeyFields(record, index);
-          bindNonKeyFields(record, valueStruct, isStringValue, index);
+          bindNonKeyFields(record, valueStruct, index);
           break;
 
         case UPDATE:
-          index = bindNonKeyFields(record, valueStruct, isStringValue, index);
+          index = bindNonKeyFields(record, valueStruct, index);
           bindKeyFields(record, index);
           break;
         default:
@@ -173,11 +173,11 @@ public class PreparedStatementBinder implements StatementBinder {
   protected int bindNonKeyFields(
       SinkRecord record,
       Struct valueStruct,
-      boolean isStringValue,
       int index
   ) throws SQLException {
     for (final String fieldName : fieldsMetadata.nonKeyFieldNames) {
-      if (isStringValue) {
+      if (valueStruct == null) {
+        // String-value record: bind the raw value into the synthetic column
         bindField(index++, record.valueSchema(), record.value(), fieldName);
       } else {
         final Field field = record.valueSchema().field(fieldName);
--- a/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
@@ -89,6 +89,9 @@ public class PreparedStatementBinder implements StatementBinder {
 
   @Override
   public void bindRecord(SinkRecord record) throws SQLException {
+    final Struct valueStruct = record.valueSchema() != null
+        && record.valueSchema().type() == Schema.Type.STRING
+        ? null : (Struct) record.value();
     final boolean isDelete = isNull(record.value());
     // Assumption: the relevant SQL has placeholders for keyFieldNames first followed by
     //             nonKeyFieldNames, in iteration order for all INSERT/ UPSERT queries
@@ -105,11 +108,11 @@ public class PreparedStatementBinder implements StatementBinder {
         case INSERT:
         case UPSERT:
           index = bindKeyFields(record, index);
-          bindNonKeyFields(record, index);
+          bindNonKeyFields(record, valueStruct, index);
           break;
 
         case UPDATE:
-          index = bindNonKeyFields(record, index);
+          index = bindNonKeyFields(record, valueStruct, index);
           bindKeyFields(record, index);
           break;
         default:
@@ -169,6 +172,7 @@ public class PreparedStatementBinder implements StatementBinder {
 
   protected int bindNonKeyFields(
       SinkRecord record,
+      Struct valueStruct,
       int index
   ) throws SQLException {
     if (record.valueSchema().type() == Schema.Type.STRING) {
@@ -176,7 +180,6 @@ public class PreparedStatementBinder implements StatementBinder {
         bindField(index++, record.valueSchema(), record.value(), fieldName);
       }
     } else {
-      final Struct valueStruct = (Struct) record.value();
       for (final String fieldName : fieldsMetadata.nonKeyFieldNames) {
         final Field field = record.valueSchema().field(fieldName);
         Object objectValue = this.replaceNullWithDefault
--- a/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
@@ -89,8 +89,10 @@ public class PreparedStatementBinder implements StatementBinder {
 
   @Override
   public void bindRecord(SinkRecord record) throws SQLException {
-    final Struct valueStruct = (Struct) record.value();
-    final boolean isDelete = isNull(valueStruct);
+    final boolean isPrimitiveValue = record.valueSchema() != null
+        && record.valueSchema().type().isPrimitive();
+    final Struct valueStruct = isPrimitiveValue ? null : (Struct) record.value();
+    final boolean isDelete = isNull(record.value());
     // Assumption: the relevant SQL has placeholders for keyFieldNames first followed by
     //             nonKeyFieldNames, in iteration order for all INSERT/ UPSERT queries
     //             the relevant SQL has placeholders for keyFieldNames,
@@ -106,11 +108,11 @@ public class PreparedStatementBinder implements StatementBinder {
         case INSERT:
         case UPSERT:
           index = bindKeyFields(record, index);
-          bindNonKeyFields(record, valueStruct, index);
+          bindNonKeyFields(record, valueStruct, isPrimitiveValue, index);
           break;
 
         case UPDATE:
-          index = bindNonKeyFields(record, valueStruct, index);
+          index = bindNonKeyFields(record, valueStruct, isPrimitiveValue, index);
           bindKeyFields(record, index);
           break;
         default:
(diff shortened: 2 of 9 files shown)


akanimesh7:
- Can you check copilot comments ?
- Could you please resolve outdated/applied comments ?

On the implementation side -- 
- The current implementation is coupled with "string" converter. There are multiple other converters which are schema-less. In future if we need to onboard those, then it would be difficult with the current implementation. Can you check how this is done in other sink connectors (if done), if not, can you see if there is a way to do it in a more extensible way ?

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -111,18 +123,24 @@ public static FieldsMetadata extract(
...
+        nonKeyFieldNames.add(colName);
+        allFields.put(colName, new SinkRecordField(valueSchema, colName, false));
+      } else {
+        for (Field field : valueSchema.fields()) {
+          if (keyFieldNames.contains(field.name())) {
+            continue;
+          }
+          if (!fieldsWhitelist.isEmpty() && !fieldsWhitelist.contains(field.name())) {
+            continue;
+          }
```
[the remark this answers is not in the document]
KGaneshDatta: Addressed. FieldsMetadata.extract now throws a ConnectException when `fields.whitelist` is non-empty and does not contain the configured value column name, with a message naming the offending whitelist, the config key, and the column. Three new tests in FieldsMetadataTest cover the reject / accept / empty-whitelist paths.

src/test/java/io/confluent/connect/jdbc/sink/BufferedRecordsTest.java
```diff
@@ -574,9 +574,9 @@ public void testAddRecordDeleteNotEnabledAndRecordKeyPkMode() throws SQLExceptio
     // Delete is not enabled, so therefore require non-null key and values with schemas
     assertValidRecord(true, true, true, true);
     // Fail when ingesting tombstones
-    assertInvalidRecord(true, true, false, true, "with a non-null Struct value and non-null Struct schema");
-    assertInvalidRecord(true, true, true, false, "with a non-null Struct value and non-null Struct schema");
-    assertInvalidRecord(true, true, false, false, "with a non-null Struct value and non-null Struct schema");
+    assertInvalidRecord(true, true, false, true, "Struct or primitive");
+    assertInvalidRecord(true, true, true, false, "Struct or primitive");
+    assertInvalidRecord(true, true, false, false, "Struct or primitive");
```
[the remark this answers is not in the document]
KGaneshDatta: Tightened the substring match to "requires records with a non-null Struct or String value", which is unique to RecordValidator.requiresValue() and won't accidentally match other error paths.

src/test/java/io/confluent/connect/jdbc/sink/PreparedStatementBinderTest.java
```diff
@@ -209,6 +209,87 @@ public void bindRecordInsert() throws SQLException, ParseException {
...
+          "primitiveTable", pkMode, Collections.<String>emptyList(),
+          Collections.<String>emptySet(), schemaPair
+      );
+
+      PreparedStatement statement = mock(PreparedStatement.class);
+      TableId tabId = new TableId(null, null, "primitiveTable");
+      List<ColumnDefinition> colDefs = new ArrayList<>();
+      colDefs.add(mock(ColumnDefinition.class));
+      when(colDefs.get(0).type()).thenReturn(Types.VARCHAR);
+      when(colDefs.get(0).id()).thenReturn(new ColumnId(tabId, "recordValue"));
```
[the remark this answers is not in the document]
KGaneshDatta: Replaced both literal occurrences in PreparedStatementBinderTest with FieldsMetadata.DEFAULT_STRING_VALUE_COLUMN_NAME so the tests track any future rename of the constant.

src/test/java/io/confluent/connect/jdbc/sink/BufferedRecordsTest.java
```diff
@@ -636,20 +636,20 @@ public void testAddRecordDeleteNotEnabledAndKafkaPkMode() throws SQLException {
...
     assertValidRecord(false, false, true, true);
 
-    assertInvalidRecord(true, true, true, false, "with a non-null Struct value and non-null Struct schema");
-    assertInvalidRecord(false, true, true, false, "with a non-null Struct value and non-null Struct schema");
-    assertInvalidRecord(true, false, true, false, "with a non-null Struct value and non-null Struct schema");
-    assertInvalidRecord(false, false, true, false, "with a non-null Struct value and non-null Struct schema");
+    assertInvalidRecord(true, true, true, false, "Struct or primitive");
+    assertInvalidRecord(false, true, true, false, "Struct or primitive");
+    assertInvalidRecord(true, false, true, false, "Struct or primitive");
+    assertInvalidRecord(false, false, true, false, "Struct or primitive");
```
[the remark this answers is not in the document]
KGaneshDatta: Same as above — replaced with the longer, more discriminating substring across all 15 occurrences in this test class.

src/test/java/io/confluent/connect/jdbc/sink/PreparedStatementBinderTest.java
```diff
@@ -209,6 +209,87 @@ public void bindRecordInsert() throws SQLException, ParseException {
...
+          JdbcSinkConfig.InsertMode.INSERT,
+          true
+      );
+
+      binder.bindRecord(new SinkRecord(
+          "topic", 0, null, null, Schema.STRING_SCHEMA, "hello world", 0
+      ));
+
+      verify(statement, times(1)).setString(1, "hello world");
+    }
```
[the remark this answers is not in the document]
KGaneshDatta: Done.
Re-indented bindRecordStringValueInsert and bindRecordStringValueWithPrimitiveKey from 4-space to 2-space to match the surrounding test class style.

KGaneshDatta:
Thanks for the review @akanimesh7 

I have addressed all the copilot and resolved all of them.

> The current implementation is coupled with "string" converter. There are multiple other converters which are schema-less. In future if we need to onboard those, then it would be difficult with the current implementation.
The current implementation is intentionally scoped to StringConverter, which is the most requested converter from customers. Support is added through minimal, localized String-specific predicates rather than a broader abstraction.

Looking at other Confluent sink connectors — Snowflake, HTTP v2, and GCS — they emit opaque payloads (file blobs, HTTP bodies) rather than typed columns, so they don't face the problem with non-schema formats. Snowflake additionally requires a vendor-specific Converter and does not support raw StringConverter at all.

BigQuery has similar implementation to JDBC, since both write into typed columns. Currently, it simply rejects any non-Struct value with a connector failure.

In order to provide support for StringConverters we are checking the schema format as done in this PR. Other primitive types (e.g. BYTES) are not supported in most of other connectors either. If broader support is needed later, we can introduce an enum-driven dispatch with a dedicated value-handler class to cleanly separate per-format logic.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -110,7 +162,21 @@ public static FieldsMetadata extract(
     }
 
     final Set<String> nonKeyFieldNames = new LinkedHashSet<>();
-    if (valueSchema != null) {
+    if (isStringValue) {
+      if (keyFieldNames.contains(stringValueColumnName)) {
```
akanimesh7, line 127: This clash is an issue even with non-pk columns right ?

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -69,6 +71,26 @@ public static FieldsMetadata extract(
         pkMode,
         configuredPkFields,
         fieldsWhitelist,
+        DEFAULT_STRING_VALUE_COLUMN_NAME,
+        schemaPair.keySchema,
+        schemaPair.valueSchema
+    );
+  }
+
+  public static FieldsMetadata extract(
```
akanimesh7, line 60: All this branching based on `final String stringValueColumnName,` is not needed. You can simply rely on `config.getStringOutputValueColumnName` to give you the correct value. i.e, default if not configured and the configured one otherwise.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
```diff
@@ -57,6 +57,8 @@ public FieldsMetadata(
     this.allFields = allFields;
   }
 
+  public static final String DEFAULT_STRING_VALUE_COLUMN_NAME = "recordValue";
```
akanimesh7: redundant, since defined in the config definition

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -267,6 +267,19 @@ public enum TimestampPrecisionMode {
       + "specific dialect. All properly-packaged dialects in the JDBC connector plugin "
       + "can be used.";
 
+  public static final String STRING_OUTPUT_VALUE_COLUMN_NAME = "string.output.value.column.name";
+  public static final String STRING_OUTPUT_VALUE_COLUMN_NAME_DEFAULT = "recordValue";
```
akanimesh7: Is this name in accordance with database column naming conventions ?

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -171,14 +173,19 @@ protected int bindKeyFields(SinkRecord record, int index) throws SQLException {
   protected int bindNonKeyFields(
       SinkRecord record,
       Struct valueStruct,
+      boolean isStringValue,
```
akanimesh7: Rather than introducing another boolean here, isn't there an alternative way ? Can't this be made out by the information already existing in the function ?

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
KGaneshDatta: In the String-value path, `nonKeyFieldNames` only ever contains the single column value itself — there are no other non-key columns being added. So the clash with a non-pk column isn't possible in this flow.

The PK collision check exists because `keyFieldNames` is independently populated from PK extraction (e.g.,a Struct key with a field named the same as the configured value column), where overlap is reachable. And this is not the case with String schema as the record value extraction is not possible.

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
KGaneshDatta: Done. 
Dropped the no-column-name `extract()` overloads. BufferedRecords now always passes `config.stringOutputValueColumnName` to the canonical `extract()`. Kept it as a parameter to match the existing pattern — `pkMode`, `pkFields`, and `fieldsWhitelist` are passed the same way (values from config, not a config reference)

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
KGaneshDatta: Done. 
Removed the duplicated constant from FieldsMetadata where the canonical value is added in `JdbcSinkConfig.STRING_OUTPUT_VALUE_COLUMN_NAME_DEFAULT` and all test references now point there.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
KGaneshDatta: Thanks for pointing it. 
Changed the default from `recordValue` to `record_value` to match standard SQL column naming conventions.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
KGaneshDatta: Ack. Dropped the isStringValue parameter. The function now branches on `valueStruct == null`.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -174,11 +176,16 @@ protected int bindNonKeyFields(
       int index
   ) throws SQLException {
     for (final String fieldName : fieldsMetadata.nonKeyFieldNames) {
-      final Field field = record.valueSchema().field(fieldName);
-      Object objectValue = this.replaceNullWithDefault
-              ? valueStruct.get(field)
-              : valueStruct.getWithoutDefault(field.name());
-      bindField(index++, field.schema(), objectValue, fieldName);
+      if (valueStruct == null) {
```
akanimesh7: Rather than this, use the `SinkRecord` as you've done in FieldsMetadata.java

src/main/java/io/confluent/connect/jdbc/sink/metadata/FieldsMetadata.java
akanimesh7: I was thinking from the perspective of when the table is not present and it will be created, but realized that it would use the fieldsMetadata created here in this method. So it's fine.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
KGaneshDatta: Done. 
`bindNonKeyFields` now checks on `record.valueSchema().type() == Schema.Type.STRING` — same as FieldsMetadata.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
```diff
@@ -89,8 +89,10 @@ public PreparedStatementBinder(
 
   @Override
   public void bindRecord(SinkRecord record) throws SQLException {
-    final Struct valueStruct = (Struct) record.value();
-    final boolean isDelete = isNull(valueStruct);
+    final Struct valueStruct = record.valueSchema() != null
```
akanimesh7: Why is it important to make it null for STRING case ?

akanimesh7:
Thanks for the thorough test coverage! Some thoughts on tightening it up. The general principle: **test new behavior at the layer where it lives**. If `pk.mode` branching lives in `FieldsMetadata`, then a `pk.mode` test in `BufferedRecordsTest` doesn't catch anything `FieldsMetadataTest` doesn't already catch — it just adds maintenance weight.

### `FieldsMetadataTest`

| Test | Verdict | Reason |
|---|---|---|
| `nonStringPrimitiveValueSchemaIsRejected` | **remove** | Exact duplicate of `valueSchemaMustBeStructOrStringIfPresent`. Both hit the same `valueSchema.type() != STRUCT && != STRING` check; INT64 vs MAP is the same code path. Keep one (the MAP one is more meaningfully "complex non-Struct"). |

### `RecordValidatorTest`

| Test | Verdict | Reason |
|---|---|---|
| `requiresValueRejectsMapSchema` | **remove** | Duplicate of `requiresValueRejectsNonStringPrimitiveSchema` — both verify "anything not STRUCT/STRING is rejected" through the same branch. |
| `requiresValueAcceptsStructSchema` | could be removed | Tests pre-existing behavior that the PR didn't touch. If `assertValidRecord`-style coverage exists elsewhere, this is regression-net only. |
| `requiresValueRejectsNullValue` | could be removed | Same reason — null-value rejection isn't changed in this PR. |

### `BufferedRecordsTest`

| Test | Verdict | Reason |
|---|---|---|
| `testStringValueInsertPkModeNone`, `testStringValueInsertPkModeKafka`, `testStringValueWithRecordKeyPk` | could be removed (collapse to one) | `BufferedRecords` has **no** `pk.mode` branching — that logic lives in `FieldsMetadata`, already covered by `stringValueSchemaPkModeNone` / `stringValueSchemaWithRecordKeyPk`. These three tests only verify "the new schema doesn't crash the buffer" — one is enough. |
| `testStringValueBatchingWithSchemaChange` | could be removed | The existing `correctBatching` test already proves schema-change triggers a flush. The check is \`SchemaPair\` equality — type-agnostic. Adding STRING as a new schema type doesn't introduce a new branch here. |
| `testStringValueValidation` | could be removed | One-line wrapper around the validator; `RecordValidatorTest` covers this comprehensively at the right layer. |
| `testStringValueDeleteWithRecordKeyPk` | could be removed | `BufferedRecords`' insert-vs-delete dispatch is `record.value() == null`-based and schema-agnostic. Existing struct-based delete tests prove the dispatch; STRING doesn't add a new branch here. |

The **4 custom-column-name tests** (`…NewTable`, `…ExistingTableHasColumn`, `…AutoEvolveAddsColumn`, `…NoAutoEvolveFails`) are genuinely load-bearing — they're the only place that exercises the new config through `DbStructure`'s four create/amend branches. Keep all four.

### `PreparedStatementBinderTest`

Nothing to flag — both new tests verify real bind semantics (`setString` / `setLong` ordering). The simpler test localizes failures faster if the more complex one regresses, so keeping both is fine.

### TL;DR

- **Definitely remove** (2 tests): `nonStringPrimitiveValueSchemaIsRejected`, `requiresValueRejectsMapSchema` — pure duplicates.
- **Consider removing** (~7 tests): mostly in `BufferedRecordsTest`, where polymorphic-by-construction logic is already covered by `correctBatching` or by lower-layer tests in `FieldsMetadataTest` / `PreparedStatementBinderTest` / `RecordValidatorTest`.

Rule of thumb: if you ask "what bug would this test catch that no other test catches?" — and the answer is "none," it's a candidate for removal.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
KGaneshDatta: Without this condition, `(Struct) record.value()` throws `ClassCastException` for every String-valued record i.e. `record.value()` is a `java.lang.String` when `value.converter=StringConverter` and this cast is evaluated at the assignment.

KGaneshDatta:
Thanks for the review @akanimesh7 
I have removed the few unnecessary tests from the test suite.

src/main/java/io/confluent/connect/jdbc/sink/PreparedStatementBinder.java
akanimesh7: Image

Not sure i understand. The `valueStruct` is not being used in the STRING flow right ?

KGaneshDatta: Understood.
Removed the valuestruct parameter and included only in the Struct Type flow.

akanimesh7 (approved):

PR #1637, opened 2026-05-18
Remove timestamp.precision.mode from String-value config gating

KGaneshDatta:
## Problem
Remove timestamp.precision.mode from String-value config gating

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


Kalla-Shubham (approved):
lgtm

PR #1638, opened 2026-05-19
Implement connection validation in JdbcSourceConnector

pgajendran-confluent:
## Problem
* JdbcSourceConnector is currently not validating connection

## Solution
* Implement connection validation in JdbcSourceConnectorValidator.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
+++ b/src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2018 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.connect.jdbc.validation;
+
+import io.confluent.connect.jdbc.dialect.DatabaseDialect;
+import org.apache.kafka.common.config.Config;
+import org.apache.kafka.common.config.ConfigValue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+/**
+ * Base class for JDBC connector validation with shared validation logic.
+ */
+public abstract class AbstractJdbcConnectorValidation {
+
+  private static final Logger log = LoggerFactory.getLogger(AbstractJdbcConnectorValidation.class);
+
+  protected Config validationResult;
+
+  /**
+   * Validate database connection by attempting to create a connection.
+   * This method is protected to allow override in tests.
+   *
+   * @param connectionUrlConfigKey the config key name for the connection URL
+   * @return true if connection succeeds, false otherwise
+   */
+  protected boolean validateConnection(String connectionUrlConfigKey) {
+    DatabaseDialect dialect = null;
+    try {
+      dialect = createDialect();
+      dialect.getConnection();
+      return true;
+    } catch (Exception e) {
+      // Connection error - attach to connection.url config
+      configValue(validationResult, connectionUrlConfigKey)
+          .ifPresent(connectionUrlValue ->
+              connectionUrlValue.addErrorMessage(
+                  String.format("Could not connect to database. %s", e.getMessage())));
+      log.error("SQLException during validation", e);
+      return false;
+    } finally {
+      if (dialect != null) {
+        dialect.close();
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.kafka.common.config.ConfigValue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+/**
+ * Base class for JDBC connector validation with shared validation logic.
+ */
+public abstract class AbstractJdbcConnectorValidation {
```
KGaneshDatta, line 29: Though having this common Abstract class would be helpful to reduce the common code, but in this use-case of connection validation, I see the repetitive code would be the validateConnection() method only. All the remaining helper methods are already present in the class. 
Hence, I feel this would be required as of now. We can extend it later when there are common validations across source and Sink

WDYT ?

src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Find a config value by name, only if it has no errors.
+   * This helper is used to attach errors only to configs that passed individual validation.
+   *
+   * @param config the Config object
+   * @param name the config key name
+   * @return Optional containing the ConfigValue if found and error-free, empty otherwise
+   */
+  protected Optional<ConfigValue> configValue(Config config, String name) {
```
KGaneshDatta, line 79: This configValue method is similar to addErrorMessage() method which adds the error to the config which needs to be shown. So, I see this method use-case is duplication of that method use-case.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -91,6 +92,7 @@ public Config validate() {
       boolean validationResult = validateMultiConfigs()
           && validateLegacyNewConfigCompatibility()
           && validateQueryConfigs()
+          && validateConnection(CONNECTION_URL_CONFIG)
```
KGaneshDatta, line 95: I see we had a guard that this connection validation should not happen when CREDENTIALS_PROVIDER class is set. Is that guard kicking in connector templates ?

src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
```diff
@@ -0,0 +1,86 @@
...
+      dialect = createDialect();
+      dialect.getConnection();
+      return true;
+    } catch (Exception e) {
+      // Connection error - attach to connection.url config
+      configValue(validationResult, connectionUrlConfigKey)
+          .ifPresent(connectionUrlValue ->
+              connectionUrlValue.addErrorMessage(
+                  String.format("Could not connect to database. %s", e.getMessage())));
+      log.error("SQLException during validation", e);
```
KGaneshDatta, line 54: I see we are capturing all the generic exception and making the log line as SQLException during validation. Can we make the log line more generic here by rephrasing it as Exception during connection validation.

KGaneshDatta (commented):
Thanks for the PR @pgajendran-confluent 
Added few comments. PTAL

src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
pgajendran-confluent: The `configValue` method retrieves the config value and does not add any error message; where as the `addErrorMessage` add an error to a specific config value. Don't think these two method are duplicates.

src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
pgajendran-confluent: `validateConnection` and `configValue` are duplicate methods. I feel we should avoid code duplication even for a single method to reduce deviations in the future. This will reduce the risk of sink validation deviating from source validation.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
pgajendran-confluent: Yes. But the connection validation is always skipped because the CREDENTIALS_PROVIDER has a default value. So even for jdbc connection, the connection validation is skipped.

src/main/java/io/confluent/connect/jdbc/validation/AbstractJdbcConnectorValidation.java
pgajendran-confluent: I wanted to retain the same message format that was present earlier, so that there is no deviation in the logs prior to and after the change is made.

PR #1640, opened 2026-05-27
CC-41005: honor trim.sensitive.log in JdbcDbWriter exception log sites

sangeet259:
## Overview
Even though we this feature gated behind a cofig, it previously gated only `JdbcSinkTask`'s log sites, `JdbcDbWriter` was still logging the exception untrimmed, leaving a gap. (confirmed this locally on CP)

Apply `LogUtil.trimSensitiveData` to the two `log.error` sites in `JdbcDbWriter`'s write/rollback catch block (via a `maybeTrim` helper). 

## Verified via kafka-docker-playground

To force the `BatchpdateException`: 
1. We pre-created the `orders` table with a primary key on `id` and seeded one row (`id=1`). 
2. Configured the connector with `insert.mode=insert`, `pk.mode=record_value`, `pk.fields=id`, `batch.size=100`, then produced a Kafka record whose `id=1` collides with the seeded row. 
3. The Postgres JDBC driver fails the batch with a `BatchUpdateException` whose message includes the full `INSERT INTO ... VALUES (...)`, the leak shape this flag was designed to redact.

**Before** (`trim.sensitive.log` absent => default false):
```
java.sql.BatchUpdateException: Batch entry 0 INSERT INTO "postgres"."public"."orders" ("id","product","quantity","price") VALUES (('1'::int4),('LEAK_CANARY_PIIPIIPII_42'),('1'::int4),('9.99'::real)) was aborted: ERROR: duplicate key value violates unique constraint "orders_pkey"
```

**After** (`trim.sensitive.log=true`):
```
java.sql.BatchUpdateException: Batch entry 0 INSERT INTO "postgres"."public"."orders" ("id","product","quantity","price"): ERROR: duplicate key value violates unique constraint "orders_pkey"
```

| trim.sensitive.log | value occurrences in log |
|---|---|
| absent/false | 7 |
| true   | 0 |

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
+++ b/src/main/java/io/confluent/connect/jdbc/sink/JdbcDbWriter.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 
 import io.confluent.connect.jdbc.dialect.DatabaseDialect;
 import io.confluent.connect.jdbc.util.CachedConnectionProvider;
+import io.confluent.connect.jdbc.util.LogUtil;
 import io.confluent.connect.jdbc.util.TableId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


sangeet259:
Closed in favor of https://github.com/confluentinc/kafka-connect-jdbc/pull/1642

sangeet259:
closed in favor of https://github.com/confluentinc/kafka-connect-jdbc/pull/1642 because we need the change in 10.8.x for connectors relying on the 10.8.x version of this connector

akanimesh7 (approved):

PR #1645, opened 2026-06-02
Add validation to prevent blank Table.Types property in JDBC connectors

KGaneshDatta:
## Problem
When a customer sets `table.types=""` (empty) on a JDBC sink connector, the task fails at startup with an exception that gives no indication of what's issue:

> java.lang.IllegalArgumentException: Collection is empty
>     at java.util.EnumSet.copyOf(EnumSet.java:177)
>     at io.confluent.connect.jdbc.util.TableType.parse(TableType.java:63)
>     at io.confluent.connect.jdbc.sink.JdbcSinkConfig.<init>(JdbcSinkConfig.java:778)

The message doesn't mention `table.types`, it is difficult to understand what to fix without reading the stack trace. 

## Solution
Catch the bad value in the field validator (`EnumRecommender.ensureValid`) so it is reported during `ConfigDef` validation instead of crashing later:

- Reject an **empty** list with a `ConfigException` that names the property and lists the valid values.
- Reject a **null** element in the list (previously a silent no-op).

Because the error is now raised inside `ConfigDef.validateAll()`, it is recorded as a **field-level config error**: the validate API returns **200** with the error attached to `table.types`, and connector creation returns **400** — instead of an uncaught 500 / stuck provisioning. Invalid/mixed values (e.g. `FOO`, `TABLE,FOO`) were already handled by the per-element check and are unchanged.

The fix is scoped to `EnumRecommender`'s **list branch**, which only `table.types` uses. Single-value enum configs that share `EnumRecommender` (e.g. `numeric.mapping`, default `null`) go through the untouched non-list path, so an unset/`null` value remains valid — no backward-compatibility impact. No change was needed in `TableType` (parsing can assume validated input).

- [ ] yes
- [ ] no

## Test Strategy

Local end-to-end on Confluent Platform and restarted Connect, exercised both REST paths (/connector-plugins/.../config/validate and POST /connectors) against a Postgres sink with table.types="":
Before fix:
- validate → 500 `{"message":"Collection is empty"}`
- POST /connectors → 500
- Connect log: IllegalArgumentException at EnumSet.copyOf:175 → TableType.parse:63 → JdbcSinkConfig.<init>:778
- The empty value slipped past field-level validation and failed later as an uncaught exception.

After fix:
- validate → 200 with error_count: 1; error attached to the table.types field:
`Invalid value [] for configuration table.types: At least one value must be specified. Valid values are: [table, partitioned table, view]`
- POST /connectors → 400 with the same validation message (proper config-error code, not a 500).
- The empty value is now rejected during `ConfigDef.validateAll()`, so it surfaces as a field-level config error — consistent with how an invalid type (`table.types=<value>`) was already handled.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
**Backward Compatibility**
This change is backward compatible for all functional configurations. The validation logic only adds rejection of two inputs that were never valid: an empty `table.types` and a null element.

- **Unset / valid / multiple-valid values** (`TABLE`, `VIEW`, `PARTITIONED TABLE`, or combinations): unchanged — accepted before and after.
- **Invalid values** (e.g. `FOO`, or `TABLE,FOO`): unchanged — rejected at config validation before and after (validate returns 200 with a field-level error, create returns 400).
- **Empty value (`table.types=""`):** this is the only changed path. Previously it did **not** pass validation — the per-field validator failed to flag the empty value, so `validate()` proceeded to construct the config, threw `IllegalArgumentException: Collection is empty` uncaught, and the **validation call itself failed (HTTP 500)**, leaving the connector stuck in provisioning and the task unable to start. Now it is rejected cleanly at validation time (validate returns 200 with a field-level error, create returns 400) with an actionable message. A connector with an empty value was never able to run, so no working connector is affected.
- **Null element:** not reachable through config-string parsing (a config string can only yield non-null values; an empty slot becomes `""`, not `null`), so no existing connector can be in this state. Handled defensively and covered by a unit test.

Only for one of the existing connectors, the probe failed and that connector is in FAILED/PROVISIONING state since long time due to table not found error. For remaining validations, all of them are the new connector validations and the validation has been failing.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/util/EnumRecommender.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/EnumRecommender.java
@@ -56,10 +56,10 @@ public class EnumRecommender implements ConfigDef.Validator, ConfigDef.Recommend
       }
       for (Object v : values) {
         if (v == null) {
-          validate(key, null);
-        } else {
-          validate(key, v);
+          throw new ConfigException(key, value,
+              "Null is not a valid value. Valid values are: " + canonicalValues);
         }
+        validate(key, v);
       }
     } else {
       validate(key, value);
--- a/src/main/java/io/confluent/connect/jdbc/util/EnumRecommender.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/EnumRecommender.java
@@ -50,6 +50,10 @@ public class EnumRecommender implements ConfigDef.Validator, ConfigDef.Recommend
   public void ensureValid(String key, Object value) {
     if (value instanceof List) {
       List<?> values = (List<?>) value;
+      if (values.isEmpty()) {
+        throw new ConfigException(key, value,
+            "At least one value must be specified. Valid values are: " + canonicalValues);
+      }
       for (Object v : values) {
         if (v == null) {
           validate(key, null);
--- a/src/main/java/io/confluent/connect/jdbc/util/TableType.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/TableType.java
@@ -46,19 +46,29 @@ public enum TableType {
     return value;
   }
 
+  private static final String VALID_TYPES =
+      EnumSet.allOf(TableType.class).stream()
+             .map(TableType::toString)
+             .collect(Collectors.joining(", "));
+
   public static TableType get(String name) {
     if (name != null) {
       name = name.trim();
     }
-    for (TableType method : values()) {
-      if (method.toString().equalsIgnoreCase(name)) {
-        return method;
+    for (TableType type : values()) {
+      if (type.toString().equalsIgnoreCase(name)) {
+        return type;
       }
     }
-    throw new IllegalArgumentException("No matching QuoteMethod found for '" + name + "'");
+    throw new IllegalArgumentException(
+        "No matching TableType found for '" + name + "'. Valid values are: " + VALID_TYPES);
   }
 
   public static EnumSet<TableType> parse(Collection<String> values) {
+    if (values == null || values.isEmpty()) {
+      throw new IllegalArgumentException(
+          "At least one table type must be specified. Valid values are: " + VALID_TYPES);
+    }
     Set<TableType> types = values.stream().map(TableType::get).collect(Collectors.toSet());
     return EnumSet.copyOf(types);
   }
--- a/src/main/java/io/confluent/connect/jdbc/util/TableType.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/TableType.java
@@ -46,29 +46,19 @@ public enum TableType {
     return value;
   }
 
-  private static final String VALID_TYPES =
-      EnumSet.allOf(TableType.class).stream()
-             .map(TableType::toString)
-             .collect(Collectors.joining(", "));
-
   public static TableType get(String name) {
     if (name != null) {
       name = name.trim();
     }
-    for (TableType type : values()) {
-      if (type.toString().equalsIgnoreCase(name)) {
-        return type;
+    for (TableType method : values()) {
+      if (method.toString().equalsIgnoreCase(name)) {
+        return method;
       }
     }
-    throw new IllegalArgumentException(
-        "No matching TableType found for '" + name + "'. Valid values are: " + VALID_TYPES);
+    throw new IllegalArgumentException("No matching QuoteMethod found for '" + name + "'");
   }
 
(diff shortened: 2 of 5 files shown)


src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkConfigTest.java
```diff
@@ -133,6 +135,22 @@ public void shouldCreateConfigWithTrailingWhitespaceInTableTypes() {
...
+  @Test
+  public void shouldFailToCreateConfigWithEmptyTableTypes() {
+    props.put("table.types", "");
+    ConfigException ex = assertThrows(ConfigException.class, this::createConfig);
+    assertTrue(ex.getMessage().contains("table.types"));
+    assertTrue(ex.getMessage().contains("At least one table type must be specified"));
+  }
+
+  @Test
+  public void shouldFailToCreateConfigWithInvalidTableType() {
```
akanimesh7, line 146: nit - In this file, only validating that the type of thrown exception is `ConfigException` and not `IllegalArgumentException` should be enough.

src/test/java/io/confluent/connect/jdbc/util/TableTypeTest.java
```diff
@@ -74,6 +77,37 @@ public void shouldHaveUpperCaseToString() {
...
+    assertTrue(ex.getMessage().contains("VIEW"));
+  }
+
+  @Test
+  public void parseShouldRejectNullCollection() {
+    assertThrows(IllegalArgumentException.class, () -> TableType.parse(null));
+  }
+
+  @Test
+  public void getShouldRejectUnknownTypeWithDescriptiveMessage() {
```
akanimesh7: nit add a similar test for parse too.

akanimesh7:
> After fix: response was 500 

Was it 500 from both create and validate APIs ? Ideally, validate API should return 200 with the actionable error message

akanimesh7 (dismissed):
Looks Good. 
Minor suggestions.

src/main/java/io/confluent/connect/jdbc/util/EnumRecommender.java
```diff
@@ -50,6 +50,10 @@ public static <E> EnumRecommender in(E... enumerators) {
   public void ensureValid(String key, Object value) {
     if (value instanceof List) {
       List<?> values = (List<?>) value;
+      if (values.isEmpty()) {
+        throw new ConfigException(key, value,
+            "At least one value must be specified. Valid values are: " + canonicalValues);
```
akanimesh7, line 55: We would like to catch all cases in this function itself. By all cases we mean
- Empty
- Wrong 
- List of (right, wrong)
- null

Let's write a UT for all the above cases.

src/main/java/io/confluent/connect/jdbc/util/TableType.java
```diff
@@ -46,19 +46,29 @@ public String toString() {
...
+        return type;
       }
     }
-    throw new IllegalArgumentException("No matching QuoteMethod found for '" + name + "'");
+    throw new IllegalArgumentException(
+        "No matching TableType found for '" + name + "'. Valid values are: " + VALID_TYPES);
   }
 
   public static EnumSet<TableType> parse(Collection<String> values) {
+    if (values == null || values.isEmpty()) {
```
akanimesh7: If [this](https://github.com/confluentinc/kafka-connect-jdbc/pull/1645/changes#r3346359787) is done, there is no need to make changes in this file. Basically, since all problems are caught in validation, during parsing we can simply assume that input will be sane and not check for these cases.

akanimesh7:
> Empty value (table.types=""):- this is the only changed path. Previously it passed config validation but then failed during config construction with IllegalArgumentException:

"Previously it passed config validation" -- This is not true. I tried creating a mysql sink connector and it was stuck in provisioning forever, since the validation call itself failed. Can you re-check? 

Also please check if you roll this with the fix, does it get fixed ?

KGaneshDatta:
@akanimesh7 I have done an upgrade testing on CCloud, and validated that these connectors which are stuck in PROVISIONING after post-roll, both MySQL and Postgres sinks now fail fast with a clean table.types field error instead of hanging.

akanimesh7 (approved):

akanimesh7 (approved):

PR #1648, opened 2026-06-11
Strip catalog from Postgres-discovered TableIds to fix table matching and offsets on pgjdbc 42.7+ [INC-11312]

sangeet259:
## Problem

pgjdbc 42.7.5 ([pgjdbc/pgjdbc#3390](https://github.com/pgjdbc/pgjdbc/pull/3390) changed `DatabaseMetaData.getTables(...)` to return the database name in the `TABLE_CAT` column, where older drivers returned `NULL`. Since the source connector builds its table identifier directly from that metadata (`GenericDatabaseDialect.tableIds()` → `new TableId(TABLE_CAT, TABLE_SCHEM, TABLE_NAME)`), so on the new driver every discovered Postgres identifier silently changed from two-part `schema.table` to three-part `db.schema.table`.

That breaks two things:

1. **Table matching.** `table.include.list` / `table.whitelist` are matched as whole-string regexes against the discovered identifier. The documented two-part form (e.g. `public.customers`) and schema-anchored regexes (`app\.tbl_.*`) no longer match a three-part identifier, so zero tables are selected and the task dies with *"Task is being killed because it was not assigned a table nor a query to execute"*.
2. **Offset continuity.** The same identifier is the source-offset partition key (`OffsetProtocols.sourcePartitionForProtocolV1`). Connectors whose include list still matched (e.g. leading-`.*` regexes) kept running but looked up their committed offsets under the new three-part key, found nothing, and re-read entire tables from the initial offset.

## Fix

`PostgreSqlDatabaseDialect` now overrides `tableIds()` and drops the catalog from every discovered `TableId`, pinning the identifier to the two-part `schema.table` form on every driver version.

Since PostgreSQL connection is bound to exactly one database and cannot query across databases, so `getTables()` only ever returns tables of `current_database()`. `TABLE_CAT` is therefore the same constant on every row and adds zero disambiguation, the schema is the real uniqueness boundary in Postgres. Dropping it restores the exact identifier the connector has always used, which also means existing committed offsets keep resolving (no re-reads, no duplicates) and generated SQL is unchanged.

It is only Postgres scoped because for MySQL and SQL Server the catalog is a genuine discriminator (a single connection can see multiple databases; for MySQL the catalog *is* the qualifier and the schema is null). 

`GenericDatabaseDialect` and all other dialects are untouched.

**Debezium does the same thing:** its Postgres connector normalizes the catalog out of discovered table identifiers — [`PostgresConnection#createTableId`](https://github.com/debezium/debezium/blob/v3.1.0.Final/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java#L846-L848) returns `new TableId(null, schemaName, tableName)` for exactly this single-database reason.

## Sink Connectors Safety

Sink is not affected because the direction in which we construct table id is opposite of what we do in source. In Sink, we take the `table.name.format` then we try to create a `TableId` object out of it. And for the information that is not present in the `table.name.format`, we backfill it by inferring it from connection. i.e. query the connection and get the values from the connection itself.

Now depending on how the user has configured the `table.name.format`, (whether `a.b.c` or `b.c` or just `c`), we do one of the following:

**1-part name**
- (e.g. mytable, or ${topic} with a "dotless" topic): in this case both the parsed catalog and parsed schema are null ->  both get backfilled -> effective identifier is   current_db.current_schema.mytable. 
- No effect on 1-part names as driver will return the same when checked, as we got this info from driver itself in the first place

**2-part name**
- (e.g. myschema.mytable): parsed catalog is null -> backfilled with getCatalog(); the user's schema is kept.
- Note: getCatalog() on the old driver was not NULL, it returns queryExecutor.getDatabase(), the actual database name.
- The thing that was null was DatabaseMetaData.getTables() -> TABLE_CAT column, this is changed to return current_database()
- Hence No effect on 2-part names

**3-part name**
- (mydb.myschema.mytable): the config value from the customer already fills all three slots, so the fallback never fires,  the user's first segment is used verbatim.
- Since the user provided catalog name is used verbatim in queries, if it was correct before, it will be correct now.
- If it was incorrect before(i.e. a user provides a database / catalog name that is different from what he is connected to, the rendered DML is a cross-database reference, which Postgres rejects regardless of driver.
- So 3-part names would also have no change in behavior from driver upgrade.

## Testing

- **Unit:** `PostgreSqlDatabaseDialectTest#shouldStripCatalogFromDiscoveredTableIds` — mocked `getTables` metadata returning a populated `TABLE_CAT` (the 42.7.5+ behavior); asserts discovered `TableId`s have a null catalog.
- **Integration (regression gate):** new `PostgresLiteralIncludeListIT` against the real 42.7.11 driver:
  - table in a **non-public schema** with a **literal two-part** `table.include.list` (`app.customers`): fails without the dialect change, passes with it;
  - **offset resumption** across a connector re-create: exactly 3 + 2 records, no re-reads. This asserts the offset-key half of the regression, not just matching.
- Existing source ITs use a leading-`.*` include-list regex, which tolerates a catalog-prefixed identifier on both driver versions, that is the coverage gap that let this regression through, and the new IT closes it.
- Tested on CP / Kafka Docker Playground

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -439,16 +439,16 @@ public class GenericDatabaseDialect implements DatabaseDialect {
       throw new IllegalArgumentException("Invalid fully qualified name: '" + fqn + "'");
     }
     if (parts.size() == 1) {
-      return new TableId(null, null, parts.get(0));
+      return createTableId(null, null, parts.get(0));
     }
     if (parts.size() == 3) {
-      return new TableId(parts.get(0), parts.get(1), parts.get(2));
+      return createTableId(parts.get(0), parts.get(1), parts.get(2));
     }
     assert parts.size() >= 2;
     if (useCatalog()) {
-      return new TableId(parts.get(0), null, parts.get(1));
+      return createTableId(parts.get(0), null, parts.get(1));
     }
-    return new TableId(null, parts.get(0), parts.get(1));
+    return createTableId(null, parts.get(0), parts.get(1));
   }
 
   /**
@@ -461,15 +461,13 @@ public class GenericDatabaseDialect implements DatabaseDialect {
   }
 
   /**
-   * Create a {@link TableId} from table identity components read from driver metadata
-   * ({@code getTables}, {@code getColumns}, {@code getPrimaryKeys}, result set metadata).
-   * All metadata-derived identifiers are built through this single seam so that a dialect
-   * can normalize them consistently; identifiers parsed from configuration go through
-   * {@link #parseTableIdentifier(String)} instead and are not affected.
+   * Single construction point for every {@link TableId} the dialect creates, whether read
+   * from driver metadata or parsed from a configured name. A dialect can override this to
+   * normalize identifiers in one place.
    *
-   * @param catalogName the catalog name reported by the driver; may be null
-   * @param schemaName  the schema name reported by the driver; may be null
-   * @param tableName   the table name reported by the driver
+   * @param catalogName the catalog name; may be null
+   * @param schemaName  the schema name; may be null
+   * @param tableName   the table name; never null
    * @return the table identifier; never null
    */
   protected TableId createTableId(String catalogName, String schemaName, String tableName) {
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -460,6 +460,22 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     return false;
   }
 
+  /**
+   * Create a {@link TableId} from table identity components read from driver metadata
+   * ({@code getTables}, {@code getColumns}, {@code getPrimaryKeys}, result set metadata).
+   * All metadata-derived identifiers are built through this single seam so that a dialect
+   * can normalize them consistently; identifiers parsed from configuration go through
+   * {@link #parseTableIdentifier(String)} instead and are not affected.
+   *
+   * @param catalogName the catalog name reported by the driver; may be null
+   * @param schemaName  the schema name reported by the driver; may be null
+   * @param tableName   the table name reported by the driver
+   * @return the table identifier; never null
+   */
+  protected TableId createTableId(String catalogName, String schemaName, String tableName) {
+    return new TableId(catalogName, schemaName, tableName);
+  }
+
   @Override
   public List<TableId> tableIds(Connection conn) throws SQLException {
     DatabaseMetaData metadata = conn.getMetaData();
@@ -473,7 +489,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
         String catalogName = rs.getString(1);
         String schemaName = rs.getString(2);
         String tableName = rs.getString(3);
-        TableId tableId = new TableId(catalogName, schemaName, tableName);
+        TableId tableId = createTableId(catalogName, schemaName, tableName);
         if (includeTable(tableId)) {
           tableIds.add(tableId);
         }
@@ -724,7 +740,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
         final String catalogName = rs.getString(1);
         final String schemaName = rs.getString(2);
         final String tableName = rs.getString(3);
-        final TableId tableId = new TableId(catalogName, schemaName, tableName);
+        final TableId tableId = createTableId(catalogName, schemaName, tableName);
         final String columnName = rs.getString(4);
         final ColumnId columnId = new ColumnId(tableId, columnName, null);
         final int jdbcType = rs.getInt(5);
@@ -816,7 +832,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
     String catalog = rsMetadata.getCatalogName(column);
     String schema = rsMetadata.getSchemaName(column);
     String tableName = rsMetadata.getTableName(column);
-    TableId tableId = new TableId(catalog, schema, tableName);
+    TableId tableId = createTableId(catalog, schema, tableName);
     String name = rsMetadata.getColumnName(column);
     String alias = rsMetadata.getColumnLabel(column);
     ColumnId id = new ColumnId(tableId, name, alias);
@@ -875,7 +891,7 @@ public class GenericDatabaseDialect implements DatabaseDialect {
         String catalogName = rs.getString(1);
         String schemaName = rs.getString(2);
         String tableName = rs.getString(3);
-        TableId tableId = new TableId(catalogName, schemaName, tableName);
+        TableId tableId = createTableId(catalogName, schemaName, tableName);
         final String colName = rs.getString(4);
         ColumnId columnId = new ColumnId(tableId, colName);
         pkColumns.add(columnId);
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -439,16 +439,16 @@ public class GenericDatabaseDialect implements DatabaseDialect {
       throw new IllegalArgumentException("Invalid fully qualified name: '" + fqn + "'");
     }
     if (parts.size() == 1) {
-      return createTableId(null, null, parts.get(0));
+      return new TableId(null, null, parts.get(0));
     }
     if (parts.size() == 3) {
-      return createTableId(parts.get(0), parts.get(1), parts.get(2));
+      return new TableId(parts.get(0), parts.get(1), parts.get(2));
     }
     assert parts.size() >= 2;
     if (useCatalog()) {
-      return createTableId(parts.get(0), null, parts.get(1));
+      return new TableId(parts.get(0), null, parts.get(1));
     }
-    return createTableId(null, parts.get(0), parts.get(1));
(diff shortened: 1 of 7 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -126,6 +128,28 @@ protected Properties buildAuthenticationProperties(JdbcCredentials jdbcCredentia
...
+   * {@code DatabaseMetaData.getTables(...)} belongs to {@code current_database()} and the
+   * catalog adds no disambiguation. pgjdbc 42.7.5+ populates {@code TABLE_CAT} with the
+   * database name where older drivers returned {@code null}, which would turn the
+   * connector's table identifiers into three-part {@code db.schema.table} names — breaking
+   * {@code table.include.list}/{@code table.whitelist} matching against the documented
+   * two-part {@code schema.table} form and changing the source-offset partition keys.
+   * Dropping the catalog keeps the identifier two-part on every driver version.
+   */
+  @Override
+  public List<TableId> tableIds(Connection conn) throws SQLException {
```
akanimesh7: The driver call that returned `null` before now returns the catalog name, and that value is used in other paths too — `describeColumns`/`describeColumnsByQuerying` build `ColumnId`s from `getColumns`/`getPrimaryKeys`, and `ColumnId` equality includes it (e.g. `pkColumns.contains` in `describeColumns`). Are those paths safe on the new driver, or do they also need a change?

akanimesh7:
What's the impact of this driver upgrade on the Postgres JDBC sink connector? Should we add tests for the sink too?

akanimesh7:
**Alternative worth considering — a single catalog seam instead of a post-strip (source-scoped):**

*Why the current approach isn't ideal:* the override only post-processes `tableIds()`, so inside the Postgres dialect a discovered `TableId` is catalog-less while `describeColumns`/`primaryKeyColumns` still build catalog-**bearing** `ColumnId`s from metadata on 42.7.11. Same concept, two representations. It works today only because nothing compares across that boundary (e.g. `pkColumns.contains(columnId)` happens to have both sides catalog-bearing) — but that's an unenforced invariant. The first `Set<TableId>` / `Map<ColumnId,…>` that mixes a stripped id with a metadata-derived one breaks silently.

*Proposed:* one construction seam in `GenericDatabaseDialect`, with all source metadata sites routed through it —

```java
// GenericDatabaseDialect — default keeps the catalog
protected TableId createTableId(String catalog, String schema, String table) {
  return new TableId(catalog, schema, table);
}
```
Route `tableIds()` (L476), `describeColumns` (L727), and `primaryKeyColumns` (L878) through `createTableId(...)`, then override in Postgres and drop the `tableIds()` override:
```java
@Override
protected TableId createTableId(String catalog, String schema, String table) {
  return new TableId(null, schema, table);   // PG connection is single-DB; catalog is noise
}
```

*This is exactly what Debezium does* — `JdbcConnection.createTableId` returns `new TableId(databaseName, schemaName, tableName)` (default, keeps catalog), and `PostgresConnection` overrides it to `new TableId(null, schemaName, tableName)`. Same base-default / Postgres-override seam, not a discovery-time post-strip. (The description already links Debezium for the *why*; this is the *how*.)

*Why this is better:* Postgres never emits a catalog from **any** source path → one representation, no null/non-null split. Every id comparison (`pkColumns.contains(...)` etc.) is uniformly catalog-free instead of correct-by-coincidence; it's a smaller conceptual change than a post-pass (no re-iterating/re-allocating the discovered list); and it mirrors the proven Debezium structure. Worth a dialect test asserting no source path emits a catalog, so the invariant is enforced rather than assumed.

*(Scoping to source here — sink is the separate question above.)*

src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
```diff
@@ -460,6 +460,22 @@ protected boolean useCatalog() {
     return false;
   }
 
+  /**
+   * Create a {@link TableId} from table identity components read from driver metadata
+   * ({@code getTables}, {@code getColumns}, {@code getPrimaryKeys}, result set metadata).
+   * All metadata-derived identifiers are built through this single seam so that a dialect
```
akanimesh7: `All metadata-derived identifiers are built through this single seam so that a dialect can normalize them consistently; identifiers parsed from configuration go through parseTableIdentifier(String) instead and are not affected.`

Sounds confusing. Can you simplify this.

akanimesh7 (commented):
I see some other usages of `new TableId` still lying around.
It would be ideal to remove all of them.

Highest priority would be 
`PostgreSqlDatabaseDialect#parseTableIdentifier`

Please check for other db types too if not that difficult to do.

akanimesh7:
First off — this routing-everything-through-the-seam change was my suggestion in an earlier review, so apologies for sending you down this path. I've since convinced myself it was wrong, and want to walk it back.

The core fix is right: stripping the catalog in `tableIds()` (where we read tables back from `getTables()`) is exactly what INC-11312 needs. That's the path that feeds the include-list matching and the offset partition key, so pinning it to `schema.table` fixes both. No concern there.

My concern is the follow-up commit that also routes `parseTableIdentifier` through the same seam. That strips the catalog from config-supplied names too (`table.name.format`), and those aren't the same as the discovered ones:

- In `tableIds()` the catalog comes from the driver — it's just the connected DB echoed back, so dropping it loses nothing.
- In `table.name.format` the catalog is something the user typed. Dropping it isn't normalizing, it's throwing away what they asked for.

Concrete case — PG sink connected to `main_db`, `table.name.format = some_other_db.public.table`:

- **Before**: `parseTableIdentifier` keeps the catalog → `JdbcDbWriter` builds `TableId(some_other_db, public, table)` → Postgres rejects it loudly: *"cross-database references are not implemented"*.
- **After**: catalog stripped → `JdbcDbWriter` backfills `main_db` from the connection → writes silently to `main_db.public.table`.

So a wrong cross-DB prefix goes from a hard error to a silent write into the wrong database. The "it was already failing anyway" reasoning holds for working configs, but turning a clear failure into silent misrouting is worse, and it isn't needed for the fix. It also undercuts the sink-safety argument in the PR description — backfill is only safe as long as a user-supplied catalog survives parsing.

Ask: keep the catalog strip in `tableIds()` only (the SqlServer `tableIds()`/synonym change is fine — same discovery path), and revert the `parseTableIdentifier` routing so config-supplied catalogs survive.

src/test/java/io/confluent/connect/jdbc/integration/PostgresLiteralIncludeListIT.java
```diff
@@ -0,0 +1,201 @@
...
+    props.put(JdbcSourceConnectorConfig.MODE_CONFIG, JdbcSourceConnectorConfig.MODE_INCREMENTING);
+    props.put(JdbcSourceConnectorConfig.INCREMENTING_COLUMN_MAPPING_CONFIG,
+        SCHEMA_NAME + "." + TABLE_NAME + ":id");
+    props.put(JdbcSourceConnectorConfig.TOPIC_PREFIX_CONFIG, TOPIC_PREFIX);
+    props.put(JdbcSourceConnectorConfig.POLL_INTERVAL_MS_CONFIG, String.valueOf(POLLING_INTERVAL_MS));
+    props.put(JdbcSourceConnectorConfig.POLL_LINGER_MS_CONFIG, "0");
+    props.put(JdbcSourceConnectorConfig.VALIDATE_NON_NULL_CONFIG, "false");
+    // The literal, documented two-part form. A leading-.* regex would mask the regression
+    // because it tolerates a catalog-prefixed identifier.
+    props.put(JdbcSourceConnectorConfig.TABLE_INCLUDE_LIST_CONFIG, SCHEMA_NAME + "." + TABLE_NAME);
```
akanimesh7, line 127: Worth adding a case where the three part name fails. In future if it suddenly starts working fine, it might  be an unverified behaviour

src/test/java/io/confluent/connect/jdbc/integration/PostgresLiteralIncludeListIT.java
```diff
@@ -0,0 +1,201 @@
...
+  @After
+  public void tearDown() throws SQLException {
+    stopConnect();
+    try (Statement stmt = connection.createStatement()) {
+      stmt.execute("DROP SCHEMA " + SCHEMA_NAME + " CASCADE");
+    }
+  }
+
+  @Test
+  public void shouldDiscoverTableWithLiteralIncludeListAndResumeOffsetsAcrossRestart()
```
akanimesh7: The `shouldDiscoverTableWithLiteral` was already tested in prev test.

The other bit -- `ResumeOffsetsAcrossRestart` is a vanilla check on if there is an already existing offset, it should be respected. Is IT a right place for this ? I remember there already are unit tests for this. Can you check them once ? Is there a case that those will not be able to catch which this test case would catch

src/test/java/io/confluent/connect/jdbc/integration/PostgresLiteralIncludeListIT.java
sangeet259: added `shouldFailWithThreePartIncludeList`

src/test/java/io/confluent/connect/jdbc/integration/PostgresLiteralIncludeListIT.java
sangeet259: right, checked `JdbcSourceTaskUpdateTest`, they already cover the mechanism. removed it from here.

akanimesh7 (approved):

PR #148, opened 2016-10-17
Update licenses, tweaking some dependency scopes, files to include in packaging, and ensuring the create-licenses scope can run cleanly without extra manual steps.

ewencp:

PR #1649, opened 2026-06-12
CC-41006: extend LogUtil right-edge markers to cover Redshift batched errors

jjain1259:
## Summary

`LogUtil.getNonSensitiveErrorMessage` was hardcoded to find `"\n  Detail: "` as the right-edge of the safe `": ERROR: <reason>"` segment that follows a stripped `VALUES (...)` body. That marker is only emitted when the server populates a `DETAIL` field, so for error classes where the server omits `DETAIL` (e.g., `22001` `string_data_right_truncation` raised from `varchar.c:484` on a multi-row `INSERT INTO ... VALUES (...), (...)` failure surfaced through Redshift's `redshift-jdbc42` driver), the function fell through to prefix-only output, dropping the human-readable error reason along with the row data.

Replace the single marker with a small ordered list:

| Marker | Source |
|---|---|
| `\n  Detail: ` | pgjdbc / redshift-jdbc when the server includes a DETAIL field |
| `\n  Hint: ` | pgjdbc / redshift-jdbc when only HINT is set (no DETAIL) |
| `  Call getNextException ` | pgjdbc `BatchResultHandler` suffix; redshift-jdbc reuses the same template verbatim, so this bounds Redshift's multi-row INSERT errors even when DETAIL/HINT are absent |

The earliest matching marker wins, so existing Postgres behavior is unchanged when `DETAIL` is present. When no marker matches, the function still falls back conservatively to prefix-only output — no row-data leak.

## Why

For sink connectors that batch records into a single multi-row `VALUES (...), (...)` INSERT for throughput (the JDBC sink's default `insert.mode=insert`), the BatchUpdateException message embeds the full row data plus the error reason. The `kafka-connect-aws-redshift` connector and the Postgres sink rely on logredactor rules in `logredactor-rules/public/prod-connect.rules` (e.g. `RedshiftSink BatchUpdateException`, `PostgresSink BatchUpdateException`) to mask this externally. CC-41006 aims to retire those rules in favor of `trim.sensitive.log=true` (which routes the exception through `LogUtil` at both `JdbcDbWriter.write` and `JdbcSinkTask.put` log sites).

End-to-end testing of just `trim.sensitive.log=true` against a dev Redshift cluster via kafka-docker-playground showed that the row data is masked correctly, but the `: ERROR: value too long for type character varying(5)` reason was also dropped — leaving operators with `Batch entry 0 ... INSERT INTO "..." ("id","name")` and no indication of *why* the batch failed. This patch restores the error reason for Redshift's batched INSERT shape (and, secondarily, for Postgres errors where only HINT is set), without weakening the fallback for unknown shapes.

### Before (`v10.8.6` against Redshift multi-row INSERT failure)
```
java.sql.BatchUpdateException: Batch entry 0 /* -partner Confluent Redshift Connector */ INSERT INTO "db"."public"."t" ("id","name")
```
(error reason silently dropped)

### After
```
java.sql.BatchUpdateException: Batch entry 0 /* -partner Confluent Redshift Connector */ INSERT INTO "db"."public"."t" ("id","name"): ERROR: value too long for type character varying(5)
```
(row data still masked; reason preserved)

## Tests

Added four `LogUtilTest` cases:

| Test | Covers |
|---|---|
| `testRedshiftMultiRowBatchUpdateSensitive` | Redshift multi-row INSERT failure — no DETAIL, ends with `  Call getNextException` |
| `testBatchUpdateSensitiveHintOnly` | Postgres failure with only HINT (e.g., undefined function) |
| `testBatchUpdateSensitiveDetailBeforeHint` | Both markers present — earliest wins (existing conservative behavior preserved) |
| `testBatchUpdateSensitiveNoKnownMarker` | Future error format with no known marker — falls back to prefix-only |

All 15 LogUtilTest cases pass locally (11 existing + 4 new):
\`\`\`
[INFO] Tests run: 15, Failures: 0, Errors: 0, Skipped: 0 - in io.confluent.connect.jdbc.util.LogUtilTest
[INFO] BUILD SUCCESS
\`\`\`

Behavior on existing inputs is unchanged — when DETAIL is present, the earliest-match selection still picks DETAIL first, identical to the prior single-marker logic.

## Test plan

- [x] \`mvn test -Dtest=LogUtilTest\` — 15/15 pass
- [ ] CI green
- [ ] End-to-end repro on Redshift dev cluster (after this lands in a release + bump in \`kafka-connect-aws-redshift\`)

## Related

- CC-41006 — Remove RedshiftSink logredactor rules
- \`kafka-connect-aws-redshift\` PR #148 — bumps \`kafka-connect-jdbc\` to \`10.8.6\` and flips \`trim.sensitive.log=true\` (will pick up this patch on next bump)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
@@ -48,14 +48,28 @@ public class LogUtil {
         e.getUpdateCounts());
   }
 
+  // Structured ServerErrorMessage labels — only ever appear at field boundaries, so safe to trust.
+  // Redshift's redshift-jdbc42 driver is a pgjdbc fork and emits the same shape.
+  private static final String[] STRUCTURED_END_MARKERS = {
+      "\n  Detail: ",
+      "\n  Hint: "
+  };
+
+  // pgjdbc BatchResultHandler suffix (reused verbatim by redshift-jdbc42). Free-form sentence text,
+  // so used only as a fallback when no structured label is present — a reason could plausibly
+  // contain this phrase (e.g., a trigger's RAISE EXCEPTION message), in which case earliest-wins
+  // across both tiers would truncate the reason mid-sentence.
+  private static final String BATCH_SUFFIX_FALLBACK = "  Call getNextException ";
+
   // This implementation assumes it to be Postgres, see toString() of ServerErrorMessage.java
   // as well as the constructor of PSQLException.java with "boolean detail" flag in
   // https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/util/
+  // Redshift's redshift-jdbc42 driver is a pgjdbc fork that emits the same message shape,
+  // including the BatchResultHandler "Call getNextException" suffix used as the Tier 2 fallback.
   // For other JDBC Databases it would not fail but might return the same input string back!
   private static String getNonSensitiveErrorMessage(String errMsg) {
     final String sensitiveStartSearchText = ") VALUES (";
     final String errorStartSearchText = ": ERROR: ";
-    final String errorEndSearchText = "\n  Detail: ";
 
     if (errMsg == null) {
       return null;
@@ -70,13 +84,26 @@ public class LogUtil {
     String msg1 = errMsg.substring(trimStartIdx, trimEndIdx + 1);
 
     int errorStartIdx = errMsg.indexOf(errorStartSearchText);
-    int errorEndIdx = errMsg.indexOf(errorEndSearchText);
-    if (errorStartIdx < trimEndIdx || errorEndIdx < trimEndIdx
-            || errorEndIdx < errorStartIdx) {
+    if (errorStartIdx < trimEndIdx) {
       return msg1;
     }
 
-    String msg2 = errMsg.substring(errorStartIdx, errorEndIdx);
-    return msg1 + msg2;
+    // Tier 1: structured server-side field labels. Earliest match wins between them.
+    int errorEndIdx = -1;
+    for (String marker : STRUCTURED_END_MARKERS) {
+      int idx = errMsg.indexOf(marker, errorStartIdx);
+      if (idx > 0 && (errorEndIdx < 0 || idx < errorEndIdx)) {
+        errorEndIdx = idx;
+      }
+    }
+    // Tier 2: fall back to the BatchResultHandler suffix only if no structured marker matched.
+    if (errorEndIdx < 0) {
+      errorEndIdx = errMsg.indexOf(BATCH_SUFFIX_FALLBACK, errorStartIdx);
+    }
+    if (errorEndIdx < 0) {
+      return msg1;
+    }
+
+    return msg1 + errMsg.substring(errorStartIdx, errorEndIdx);
   }
 }
--- a/src/test/java/io/confluent/connect/jdbc/util/LogUtilTest.java
+++ b/src/test/java/io/confluent/connect/jdbc/util/LogUtilTest.java
@@ -168,6 +168,121 @@ public class LogUtilTest {
     assertEqualsSQLException(expectedTrimmed, actualTrimmed);
   }
 
+  // Redshift's redshift-jdbc42 driver emits this message shape for multi-row
+  // `INSERT INTO ... VALUES (...), (...)` failures (the form JDBC sink connectors batch into for
+  // throughput). The server omits the DETAIL block for many error classes (e.g., string truncation),
+  // so the trailing "  Call getNextException ..." text is the only stable right-edge marker.
+  @Test
+  public void testRedshiftMultiRowBatchUpdateSensitive() {
+    BatchUpdateException e1 = new BatchUpdateException(
+        "Batch entry 0 /* -partner Confluent Redshift Connector */ INSERT INTO "
+            + "\"db\".\"public\".\"t\" (\"id\",\"name\") VALUES (('1'::int4),('ok')),"
+            + "(('2'::int4),('secret-payload-customer-PII')) was aborted: "
+            + "ERROR: value too long for type character varying(5)"
+            + "  Call getNextException to see other errors in the batch.",
+        new int[0]);
+
+    BatchUpdateException expectedTrimmed = new BatchUpdateException(
+        "Batch entry 0 /* -partner Confluent Redshift Connector */ INSERT INTO "
+            + "\"db\".\"public\".\"t\" (\"id\",\"name\"): "
+            + "ERROR: value too long for type character varying(5)",
+        new int[0]);
+
+    SQLException actualTrimmed = LogUtil.trimSensitiveData(e1);
+    assertNoSensitiveLeak(actualTrimmed, "secret-payload-customer-PII");
+    assertEqualsSQLException(expectedTrimmed, actualTrimmed);
+  }
+
+  // Postgres can return HINT without DETAIL for some error classes (e.g., undefined function with
+  // a suggested replacement). The HINT marker should bound the safe error segment.
+  @Test
+  public void testBatchUpdateSensitiveHintOnly() {
+    BatchUpdateException e1 = new BatchUpdateException(
+        "Batch entry 0 INSERT INTO \"t\" (\"c\") VALUES ('secret') was aborted: "
+            + "ERROR: function lower(integer) does not exist\n"
+            + "  Hint: No function matches the given name and argument types.",
+        new int[0]);
+
+    BatchUpdateException expectedTrimmed = new BatchUpdateException(
+        "Batch entry 0 INSERT INTO \"t\" (\"c\"): "
+            + "ERROR: function lower(integer) does not exist",
+        new int[0]);
+
+    SQLException actualTrimmed = LogUtil.trimSensitiveData(e1);
+    assertNoSensitiveLeak(actualTrimmed, "secret");
+    assertEqualsSQLException(expectedTrimmed, actualTrimmed);
+  }
+
+  // When more than one structured marker is present, the earliest one (closest to ": ERROR: ")
+  // wins. Here DETAIL precedes HINT so the segment stops at DETAIL; the HINT block is dropped
+  // along with anything that might follow (preserves the existing conservative behavior).
+  @Test
+  public void testBatchUpdateSensitiveDetailBeforeHint() {
+    BatchUpdateException e1 = new BatchUpdateException(
+        "Batch entry 0 INSERT INTO \"t\" (\"c\") VALUES ('secret') was aborted: "
+            + "ERROR: null value in column \"c\" violates not-null constraint\n"
+            + "  Detail: Failing row contains ('secret').\n"
+            + "  Hint: ignore",
+        new int[0]);
+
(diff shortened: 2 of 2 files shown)


src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
```diff
@@ -70,13 +73,21 @@ private static String getNonSensitiveErrorMessage(String errMsg) {
...
     int errorStartIdx = errMsg.indexOf(errorStartSearchText);
-    int errorEndIdx = errMsg.indexOf(errorEndSearchText);
-    if (errorStartIdx < trimEndIdx || errorEndIdx < trimEndIdx
-            || errorEndIdx < errorStartIdx) {
+    if (errorStartIdx < trimEndIdx) {
+      return msg1;
+    }
+
+    int errorEndIdx = -1;
+    for (String marker : ERROR_END_MARKERS) {
```
sangeet259: tl;dr: 

treating all three markers equally risks accidentally truncating valid Postgres error messages. Let's use the structured markers (`Detail:`, `Hint:`) as Tier 1, and only use `Call getNextException` as a Tier 2 fallback.

---

Here we are treating the three markers as equals, and selecting the earliest one. But they are not equals. `\n  Detail: ` and `\n  Hint: ` are structured fields the Postgres server emits via `ServerErrorMessage`, pgjdbc just renders those labels. Those markers are stable and can only appear at a structured field boundary.

But `  Call getNextException ` is different. It's more like a free-form text meant for a human to read, rather than a strict, system-generated marker. The pgjdbc's `BatchResultHandler` (and `redshift-jdbc42`'s equivalent) adds it to the message to nudge the caller toward `getNextException()`. 

Consider a message where the reason text itself contains the substring, most plausibly a trigger function whose `RAISE EXCEPTION` uses the phrase, a constraint or schema name that includes it, (very unlikely, but a valid extreme edge case) or a nested driver exception echoed back into the outer error (more likely):

```
Batch entry 0 INSERT INTO "t" ("c") VALUES ('x') was aborted: ERROR: trigger raised exception: please Call getNextException helper to retry
    Detail: Failing row contains (x).
```

Under the currently proposed earliest-wins, the sentence-fragment marker `Call getNextException` at the earlier index beats the `\n  Detail: `, and the trimmed output gets cut mid-reason:

Result:
```
  Batch entry 0 INSERT INTO "t" ("c"): ERROR: trigger raised exception: please
```

In this ^ case, we failed to handle the failure mode (/ edge case) that this PR is fixing. I understand that collision probability is low, but the reason this codepath anchors on literal substrings instead of running a regex is that we don't fully trust the wire format.


I am thinking something on these lines, check if you agree with this, (just for representation of the idea):

  ```java
  int errorEndIdx = -1;
  // Tier 1: structured server-side fields (ServerErrorMessage labels).
  for (String marker : new String[]{"\n  Detail: ", "\n  Hint: "}) {
    int idx = errMsg.indexOf(marker, errorStartIdx);
    if (idx >= 0 && (errorEndIdx < 0 || idx < errorEndIdx)) {
      errorEndIdx = idx;
    }
  }
  // Tier 2: only if no structured marker matched, fall back to the
  // pgjdbc / redshift-jdbc BatchResultHandler suffix.
  if (errorEndIdx < 0) {
    errorEndIdx = errMsg.indexOf("  Call getNextException ", errorStartIdx);
  }
```

Only matters when the collision actually happens, and preserves the reason.

File: src/main/java/io/confluent/connect/jdbc/util/LogUtil.java [hunk 2]
```diff
@@ -48,14 +48,17 @@ public static Throwable trimSensitiveData(Throwable t) {
         e.getUpdateCounts());
   }
 
-  // This implementation assumes it to be Postgres, see toString() of ServerErrorMessage.java
```
sangeet259, line 51: This previous comment block pointed at `ServerErrorMessage.toString()` and `PSQLException(boolean detail)` as the authority for the message shape. Worth keeping, with the redshift-jdbc note added.

src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
```diff
@@ -48,14 +48,17 @@ public static Throwable trimSensitiveData(Throwable t) {
...
-  // as well as the constructor of PSQLException.java with "boolean detail" flag in
-  // https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/util/
-  // For other JDBC Databases it would not fail but might return the same input string back!
```
sangeet259, line 54: This comment is worth retaining as well, as this would still apply for other dialects and might be useful for a future reader to be aware of upfront.

src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
```diff
@@ -48,14 +48,17 @@ public static Throwable trimSensitiveData(Throwable t) {
...
+  // Earliest match wins. Falls back to prefix-only if none are present.
```
sangeet259: nit: this comment explains the selection logic, something that happens around 25 lines down. Minor but a reader looking at that loop has to scroll back up.

src/test/java/io/confluent/connect/jdbc/util/LogUtilTest.java
```diff
@@ -168,6 +168,86 @@ public void testBatchExceptionWithChild() {
...
+        new int[0]);
+
+    BatchUpdateException expectedTrimmed = new BatchUpdateException(
+        "Batch entry 0 INSERT INTO \"t\" (\"c\")",
+        new int[0]);
+
+    SQLException actualTrimmed = LogUtil.trimSensitiveData(e1);
+    assertEqualsSQLException(expectedTrimmed, actualTrimmed);
+  }
+
```
sangeet259, line 276: Can we add a test like `testBatchUpdateSensitiveDetailBeatsGetNextException`, a `pgjdbc` shaped message with both `Detail:` and `Call getNextException` present.

Assertion: the trimmed output ends at Detail:, not at Call getNextException. 

This would explicitly lock in the rule that structured markers should take precedence, protecting the logic against any future code changes. (following up on the previous comment about two tiers of markers)

sangeet259: something like

```java
  @Test
  public void testBatchUpdateSensitiveDetailBeatsGetNextException() {
    // pgjdbc shape: DETAIL present *and* the "Call getNextException" suffix present.
    // Locks that DETAIL takes precedence on the right-edge, independent of marker-array order.
    String errMsg =
        "Batch entry 0 INSERT INTO \"t\" (\"c\") VALUES ('x'),('toolong') was aborted: "
            + "ERROR: value too long for type character varying(5)\n"
            + "  Detail: Failing row contains (toolong).  Call getNextException to see "
            + "other errors in the batch.";
    String expectedTrimmed =
        "Batch entry 0 INSERT INTO \"t\" (\"c\"): "
            + "ERROR: value too long for type character varying(5)";
    // ... standard assertEquals against getNonSensitiveErrorMessage(errMsg)
  }
```

src/test/java/io/confluent/connect/jdbc/util/LogUtilTest.java
```diff
@@ -168,6 +168,86 @@ public void testBatchExceptionWithChild() {
...
+        new int[0]);
+
+    BatchUpdateException expectedTrimmed = new BatchUpdateException(
+        "Batch entry 0 /* -partner Confluent Redshift Connector */ INSERT INTO "
+            + "\"db\".\"public\".\"t\" (\"id\",\"name\"): "
+            + "ERROR: value too long for type character varying(5)",
+        new int[0]);
+
+    SQLException actualTrimmed = LogUtil.trimSensitiveData(e1);
+    assertEqualsSQLException(expectedTrimmed, actualTrimmed);
```
sangeet259, line 193: nit: lets add an explicit no-leak asserts first which fails with a clear readable message instead of an opaque string diff incase of `assertEqualsSQLException(expectedTrimmed, actualTrimmed);`


something like:

```java
    String actualMsg = actualTrimmed.getMessage();

    // 1. explicit no-leak asserts first. 
    // fail with a clear, readable message
    assertFalse("Row data leaked: " + actualMsg, actualMsg.contains("secret-payload-customer-PII"));
    assertFalse("VALUES clause leaked: " + actualMsg, actualMsg.contains("VALUES"));

    // 2. strict equality check last.
    assertEqualsSQLException(expectedTrimmed, actualTrimmed);
```

src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
jjain1259: Good catch — agreed on the tier split. The `Call getNextException` fragment is human-readable sentence text and can plausibly appear inside a reason (your trigger/RAISE EXCEPTION example), where earliest-wins across both tiers would truncate the reason mid-sentence. Updated to Tier 1 = structured `\n  Detail: ` / `\n  Hint: ` (earliest wins between those two, since both are stable field boundaries), Tier 2 = `  Call getNextException ` only if no Tier 1 marker matched. Added `testBatchUpdateSensitiveDetailBeatsGetNextException` to lock the precedence.

src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
jjain1259: Restored the original comment block (links to `ServerErrorMessage.toString()` and `PSQLException(boolean detail)`) and appended the redshift-jdbc42 note + the new Tier 2 fallback reference.

src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
jjain1259: Kept the "For other JDBC Databases ... might return the same input string back!" line — still applies to dialects we have not encountered yet and is useful upfront context.

src/main/java/io/confluent/connect/jdbc/util/LogUtil.java
jjain1259: Moved the selection-logic comments down to the loop site itself (each tier annotated where the code lives). The constants only carry a one-liner about what they are.

src/test/java/io/confluent/connect/jdbc/util/LogUtilTest.java
jjain1259: Added — `testBatchUpdateSensitiveDetailBeatsGetNextException` covers the pgjdbc-shaped message with both markers, asserts the trimmed output ends at DETAIL.

src/test/java/io/confluent/connect/jdbc/util/LogUtilTest.java
jjain1259: Done — added `assertNoSensitiveLeak(SQLException, String sensitiveSubstring)` helper that asserts the trimmed message does not contain the row data nor the literal `VALUES` clause. Applied to all sensitive tests; runs before `assertEqualsSQLException` so a regression surfaces as `"Row data leaked: <msg>"` instead of a string diff.

sangeet259 (approved):
Thanks for addressing the comments @jjain1259 . Looks good to me.

PR #1651, opened 2026-06-14
Support Additional Data Types for Postgres JDBC Connectors

KGaneshDatta:
## Problem

The PostgreSQL JDBC connectors had no first-class support for complex column types — `json`/`jsonb`, `hstore`, and native arrays were either silently dropped (arrays, hstore) or surfaced only as opaque strings, so these columns could not round-trip between source and sink.

## Solution

This is the **foundation** PR for opt-in complex-type support, gated behind a new `sql.complex.types.enable` flag (default `false`, so existing pipelines are unchanged). It carries no per-type source mapping of its own; it adds the shared building blocks the per-type PRs sit on.

- **Configs:** `sql.complex.types.enable` on both the source and the sink (default `false`), and `hstore.handling.mode` on the source (`map` the default, or `json`), which has no effect unless the flag is on.
- **Logical types:** `Json` — a lossless JSON STRING aligned with Debezium's `io.debezium.data.Json`; and `VariableScaleDecimal` — a `{scale INT32, value BYTES}` struct for arbitrary-precision decimals whose scale is not fixed at the column level, aligned with Debezium's `io.debezium.data.VariableScaleDecimal`.
- **Sink DDL and bind:** a `Json`-tagged STRING and a `MAP<STRING,STRING>` field are provisioned as a native `jsonb` column and written through PostgreSQL's `::jsonb` cast. `MAP<STRING,STRING>` is the only Connect container mapped, because it is the shape an `hstore` column takes on the topic; generic STRUCT values and every other map shape fall through to the generic dialect and fail rather than being silently coerced into `jsonb`. Both mappings are gated, so a sink that never opted in keeps its previous column types.
- **`JsonConverter` utility:** serializes a Connect `MAP<STRING,STRING>` to JSON object text for those `jsonb` columns.
- **Shared hstore schema:** `hstoreSchema(boolean optional)` resolves the on-topic representation for an `hstore` value under `hstore.handling.mode`, so the scalar column path (#1661) and the array element path (#1662) cannot drift apart.
- **Dialect accessors:** `dateTimeZoneId()`, `timestampGranularity()` and `dateCalendarSystem()` are exposed to subclasses on `GenericDatabaseDialect` so the PostgreSQL dialect can honour the existing time configs on complex values.
- **Dependency:** `jackson-databind` at `provided` scope, matching the Jackson the Connect runtime already ships.

One caveat worth knowing when the follow-ups land: an `hstore` column is recognised by its unqualified type name, so the schema owning the extension must be on the connection's `search_path`; otherwise the driver reports a qualified name and the column is skipped like any other unsupported type. This is documented on `hstore.handling.mode`.

The per-type source mappings build on this foundation in follow-up PRs: **#1660 (JSON/JSONB)**, **#1661 (HSTORE)** and **#1662 (native arrays)**. Scope is confined to the PostgreSQL dialect; no other dialects are affected.

- [ ] yes
- [x] no

## Test Strategy

Unit tests cover the two logical types and their round-trip through `toLogical`/`fromLogical` (`LogicalTypesTest`), the new config accessors and the `hstore.handling.mode` validation (`JdbcSourceConnectorConfigTest`), the `MAP<STRING,STRING>` serializer including empty maps, null values, key and value escaping, a null key and non-map input (`JsonConverterTest`), and on the dialect: that a `Json`-tagged STRING and a `MAP<STRING,STRING>` become `jsonb` only when the flag is on and otherwise fall back, that the bind path writes the serialized text, and that other map shapes are rejected on both the DDL and the bind path (`PostgreSqlDatabaseDialectTest`).

On the integration side this PR contributes the shared PostgreSQL harness — source-to-Kafka-to-sink round-trips, destination row assertions and schema introspection — rather than new test cases; the round-trips that exercise it arrive with the per-type PRs. The pre-existing `PostgresDatatypeIT` suite continues to run unchanged.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Merges into the `10.9.x` release branch via the standard weekly cc-connect release cycle. The feature is default-off (`sql.complex.types.enable=false`), so the change is backward compatible and safe to roll back.


akanimesh7:
> ⚙️ _This review was generated with Claude Code. Treat every point as a hypothesis to verify — the onus is on the author to confirm validity before acting. Test-coverage feedback is tracked separately._

### Solution verdict — approve the approach

Structure is sound and matches the design doc: single `sql.complex.types.enable` gate, off by default, Postgres-dialect–scoped, symmetric source/sink, unsupported element types falling back to skip-with-WARN. `ZonedTimestamp` is a genuine improvement over the doc (which conceded `timestamptz[] → timestamp[]` zone loss). No redesign needed. One solution-level inconsistency below, then code-level points.

### Code comments

**`PostgreSqlDatabaseDialect.java`**

**(a) Scalar `timestamptz` vs `timestamptz[]` — zone fidelity is asymmetric.** We now preserve the zone for `timestamptz[]` (→ `ZonedTimestamp` → native `TIMESTAMPTZ[]`), but a scalar `timestamptz` column still falls through to the generic `Types.TIMESTAMP` path → built-in `Timestamp` (zone dropped) → sink `TIMESTAMP`. So a `timestamptz` and a `timestamptz[]` column round-trip with different fidelity. Why the asymmetry? If `ZonedTimestamp` is the right representation for the array element, shouldn't the scalar use it too? Either both preserve the zone or we document why arrays are special. (This also makes the design-doc claim stale — worth updating the wiki.)

**(b) `getSqlType` advertises `JSONB[]` for `ARRAY<STRUCT>`/`ARRAY<MAP>` but `bindArray` can't bind it.** `getSqlType` recurses `ARRAY<STRUCT>`/`ARRAY<MAP>` → element `JSONB` → `JSONB[]`, so auto-create makes a `jsonb[]` column. But `bindArray` has no STRUCT/MAP element branch — it returns `false`, falls through to `super.maybeBindPrimitive`, and the write fails with an unsupported-array error. So we create a column we then can't write to. Reachable on the sink from any array-of-struct topic. Either bind it (serialize each element via `connectValueToJson` + `createArrayOf("jsonb", …)`) or don't advertise the type. (Copilot flagged this on an earlier rev — still valid.)

**(c) `isJsonBindCandidate` treats every STRUCT/MAP as JSON — collides with `VariableScaleDecimal`.** `maybeBindJson` runs first and fires for any STRUCT. `VariableScaleDecimal` is a named STRUCT, and `getSqlType` maps it to `NUMERIC` — so a *scalar* VSD field would get serialized to JSON and `setString` into a numeric column. Not reachable from the PG source today (VSD only appears as an array element), but the STRUCT→JSONB rule and `getSqlType` disagree — an unenforced invariant. Can we guard it (e.g. skip the known logical-type names in `isJsonBindCandidate`)?

**(d) Multi-dimensional arrays: skipped at read time but not at schema time.** `isMultiDimensional` is only checked inside the value converters, which return `null`. But `arrayElementSchemaFor` keys off the type name (`_int4`), identical for `int[]` and `int[][]` — so the schema still exposes it as a 1-D `ARRAY<elem>` and every value comes back `null`. If the column is `NOT NULL`, the field is built required and the converter feeds it `null` → task fails on a required field, contradicting the "never fail the task" guarantee. Can we skip the column entirely (or force the field optional) for the multi-dim case?

**(e) hstore both modes assume `rs.getObject` returns a `Map`.** Map mode returns `rs.getObject(col)` straight into a `MAP<STRING,STRING>` field; json mode passes it to `connectValueToJson(null, value)`. Both assume the driver hands back a `java.util.Map`. If pgjdbc returns the hstore *text* form (a `String`), map mode fails the `Struct.put` and json mode double-encodes it as a quoted JSON string. Have we confirmed pgjdbc returns a `Map` here (hstore OID registered on the connection)?

**(f) nit: import `JdbcSourceConnectorConfig` instead of the inline FQN.** `complexTypesEnabled()`, `jsonAsString()`, `hstoreAsJson()` all reference `io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig` by fully-qualified name — just import it, matches the rest of the file.

**`JsonConverter.java`**

**(g) nit (naming): collides with Connect's own `JsonConverter`.** `org.apache.kafka.connect.json.JsonConverter` is a well-known Connect `Converter` SPI. A `JsonConverter` in our `util` package reads like that and isn't. The earlier name `ConnectJsonConverterUtil` was clearer — rename to something unambiguous (`ConnectJsonCodec` / `JsonSerde`)?

**`pom.xml`**

**(h) `jackson-databind` added as `provided`, no version.** This connector had no jackson dependency before — now it hard-depends on it at runtime on both paths (source map parsing + sink serialization). `provided` means we trust the runtime to supply it and it won't be packaged in the plugin. Have we confirmed jackson-databind is on the plugin classloader in the fully-managed Connect runtime? If not it's a `NoClassDefFoundError` at first use. Why `provided` rather than the default, and can we pin the version?

### Scope / follow-ups

- Skipped element types (`uuid[]`, `inet[]`, …) are handled correctly — no action.
- If (b) and (c) are judged out of scope because the PG source can't produce those shapes, don't leave `getSqlType` silently advertising an unbindable type — file a ticket under CC-40460 and add a `TODO(CC-xxxxx)`.
- Please resolve the stale Copilot comments (the MySQL ones are against a dropped revision) and confirm the rebase intentionally narrowed scope to Postgres.

KGaneshDatta:
Thanks for the quick review @akanimesh7 

As discussed, I have moved the complete code changes individually to separate PR's containing each new data type.
Please find the new PR's for review
- https://github.com/confluentinc/kafka-connect-jdbc/pull/1661
- https://github.com/confluentinc/kafka-connect-jdbc/pull/1662
- https://github.com/confluentinc/kafka-connect-jdbc/pull/1660

On the above review questions :-
(a) - Currently, the timestampz and timestampz[] are handled separately. In this PR, the timestampz is not being changed. In order to support the timezones in new timestampz[] which is newly being added, I have added a new logical type i.e. ZonedTimestamp in order to keep the timezones along with timestamp. This was added after we discussed in the design discussion

(b) - Currently, ARRAY<MAP> and ARRAY<STRUCT> are not being produced by the JDBC Source connector. Hence, I haven't added the support for these types while being processed by the sink connector. But, it could be possible the sink connector could process these types as a consumer and these records can be from any source. Hence, in order to support these types also, I have updated the code to sink these record types as JSONB[] on the Postgres Database.

(c) - Every VariableScaleDecimal reference in the code is array-scoped i.e. for numeric[]. A VariableScaleDecimal value is therefore always inside an ARRAY field, and an ARRAY field bypasses `isJsonBindCandidate`/`maybeBindJson` completely. The mismatched scalar path is not possible with the connector reaching it. This VariableScaleDecimal logical type is added in order to maintain the scale of the values in the numeric[] so that they would be sync between the source and sink data sources.

(d) - When the source connector reads and builds schema from the source table, it cannot differentiate between the single-dimension and multi-dimension array types. It just takes the type of the array and builds the corresponding schema. But while processing the records, if we identify the multi-dimension array value which is currently not supported, we log the NULL and then fail the connector. Currently, if any of the unsupported type has been processed then the connector adds the NULL value in such field and moves-on. In order to resemble the same behavior, I have updated the code to do the same behavior i.e. to log the NULL value and move to next record by accepting the NULL values.

(e) - There is some contradiction with this comment. Currently, the driver returns the hstore value as key-value map and do not return the string values anytime. Hence, it is highly impossible with the driver expecting the String values. 

(f) - Provided the fix for this comment. Added the import and used the short name, matching the rest of the file.

(g) - As, this is part of the Utils package in the JDBC Connectors, I feel the FQN of the package would be a differentiating in this case and should not be a problem. This name would be helpful in order to identify the functionality of the package class.

(h) - I have added the Jackson JSON library marked "provided" (supplied by the runtime) without pinning a version. Kafka Connect always ships Jackson and tests confirm it loads, so it's safe and not an issue.

src/main/java/io/confluent/connect/jdbc/data/VariableScaleDecimal.java
```diff
@@ -0,0 +1,107 @@
...
+ * A Kafka Connect logical type for an arbitrary-precision decimal whose scale is not fixed at the
+ * column level — for example a PostgreSQL unconstrained {@code numeric}/{@code numeric[]} where
+ * each value may carry a different scale.
+ *
+ * <p>Connect's built-in {@link org.apache.kafka.connect.data.Decimal} requires a single fixed
+ * scale shared by every value, which cannot represent such columns. This type instead encodes the
+ * scale alongside the value as a STRUCT of {@code {scale INT32, value BYTES}}, where {@code value}
+ * is the two's-complement big-endian unscaled magnitude. Together they reconstruct the original
+ * {@link BigDecimal} losslessly with its own scale.
+ */
```
KGaneshDatta, line 35: VariableScaleDecimal is modeled on `io.debezium.data.VariableScaleDecimal` (used by its Postgres connector for unbounded numerics) — same STRUCT {scale: INT32, value: BYTES} encoding that reconstructs the exact BigDecimal losslessly. We reimplemented it under the Confluent namespace as a trimmed variant (plain BigDecimal and no SpecialValueDecimal), and apply it to numeric[] array elements

This stays aligned with the scalar path: scalar numeric(p,s) still uses Connect's Decimal.builder(scale) with the fixed scale from the column typmod. That only works because the column declares one scale for all rows; an unbounded numeric[] doesn't, and elements can differ (e.g. 1.50, 3.14159). 

VariableScaleDecimal is the array-side counterpart — same lossless exact-decimal guarantee, carrying the scale per value instead of per schema — the same tradeoff Debezium makes.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -345,6 +348,44 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+        && ((JdbcSinkConfig) config).sqlComplexTypesEnable;
+  }
+
+  private boolean maybeBindJson(
+      PreparedStatement statement,
+      int index,
+      Schema schema,
+      Object value
+  ) throws SQLException {
+    if (!isJsonBindCandidate(schema)) {
```
akanimesh7, line 743: This should have been in https://github.com/confluentinc/kafka-connect-jdbc/pull/1660/changes

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -345,6 +348,44 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
     return JSON_TYPE_NAME.equalsIgnoreCase(typeName) || JSONB_TYPE_NAME.equalsIgnoreCase(typeName);
   }
 
+  private boolean isJsonBindCandidate(Schema schema) {
+    Schema.Type type = schema.type();
+    if (type != Schema.Type.STRUCT && type != Schema.Type.MAP) {
```
akanimesh7: I'm confused - from what I remember the source side of it maps JSON columns to Connect type STRING, which is not listed below.

src/main/java/io/confluent/connect/jdbc/util/JsonConverter.java
```diff
@@ -0,0 +1,153 @@
...
+      return null;
+    }
+    try {
+      return MAPPER.writeValueAsString(toJsonNode(schema, value));
+    } catch (JsonProcessingException e) {
+      throw new DataException("Failed to serialize Connect value to JSON", e);
+    }
+  }
+
+  private static JsonNode toJsonNode(Schema schema, Object value) {
```
akanimesh7: Now that MAP mode is gone from the source side, is this STRUCT/MAP -> jsonb serialization path (toJsonNode and everything it feeds) still reachable from anything in these PRs, or is it orphaned? JSON/JSONB columns are always Connect STRING on the source, so this never fires for the JSON feature itself.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: `maybeBindJson` only fires for Connect STRUCT and MAP values — `isJsonBindCandidate` checks exactly
those two types. As this same is being used in HSTORE PR - https://github.com/confluentinc/kafka-connect-jdbc/pull/1661/changes - hence I have included in the base PR.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: > I remember the source side of it maps JSON columns to Connect type STRING, which is not listed below.

Yes, that is TRUE. But here the check is 

> if (type != Schema.Type.STRUCT && type != Schema.Type.MAP) {
>       return false;
>     }

which means the logical type JSON string which we are storing goes into a different path, but the HSTORE goes into this path finally ending up in the JSONB column on the sink. 

**UPDATE :-**
Removed the STRUCT support, but kept the MAP support to support HSTORE in https://github.com/confluentinc/kafka-connect-jdbc/pull/1661

src/main/java/io/confluent/connect/jdbc/util/JsonConverter.java
KGaneshDatta: No, it is not orphaned. 
1. `maybeBindJson`, for STRUCT/MAP values written into a jsonb column. Reachable from any upstream topic carrying a nested record
2. hstore `json` mode on the **source** side: `connectValueToJson(null, value)` turns the driver's hstore Map into a JSON-object string.
3. `jsonb[]` elements: `jsonbArrayFor` serializes each STRUCT/MAP element for a native `jsonb[]` column.

As, these are common code hence, I included in this base PR.

src/main/java/io/confluent/connect/jdbc/util/JsonConverter.java
```diff
@@ -0,0 +1,80 @@
...
+      return MAPPER.writeValueAsString(mapToJsonNode((Map<?, ?>) value));
+    } catch (JsonProcessingException e) {
+      throw new DataException("Failed to serialize Connect value to JSON", e);
+    }
+  }
+
+  private static ObjectNode mapToJsonNode(Map<?, ?> map) {
+    ObjectNode object = JSON_NODE_FACTORY.objectNode();
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      if (entry.getKey() == null) {
```
akanimesh7: Can this happen, and is it unrecoverable ?

Two paths: **source** (hstore json mode) can't hit it — Postgres hstore keys are never NULL, so the driver's map has no null key. **Sink** (`maybeBindJson` with a `MAP<STRING,STRING>` off an arbitrary topic) is reachable in principle, though Connect's own `validateValue` normally rejects a null key against a non-optional STRING key schema before it gets here.

If it did fire: yes, with the default `errors.tolerance=none` the sink task fails, restarts, re-reads the same record and fails again — a poison-pill loop, exactly as you suspected. The escape hatch is the standard `errors.tolerance=all` + DLQ, which is Connect's answer for a bad record; not something this method should try to solve.

Which leads to the bigger point — this guard is redundant. `mapToJsonNode` hand-builds what Jackson's default Map serializer already does (escaping, `null` -> JSON `null`, keys via `toString`), and Jackson's own default null-key serializer is a `FailingSerializer` ("Null key for a Map not allowed in JSON") that throws `JsonMappingException` — a `JsonProcessingException`, so the existing catch already converts it to a `DataException`. Same failure, same recoverability, without the hand-rolled walk:

```java
public static String connectMapToJson(Object value) {
  if (value == null) {
    return null;
  }
  if (!(value instanceof Map)) {
    throw new DataException("Expected a Map to serialize to JSON but was " + value.getClass());
  }
  try {
    return MAPPER.writeValueAsString(value);
  } catch (JsonProcessingException e) {
    throw new DataException("Failed to serialize Connect value to JSON", e);
  }
}
```

That drops `mapToJsonNode`, the `JSON_NODE_FACTORY` field and the `ObjectNode`/`JsonNodeFactory` imports. The escaping test stays green — Jackson escapes keys and values natively.

One real delta to be aware of: the non-String-value guard is **not** covered by Jackson — it would write an Integer as a JSON number rather than failing. But pgjdbc's hstore map only ever holds String or null and that's the only caller, so it's guarding an input we can't produce; the `JsonConverterTest` case pinning it would go with it.

_🤖 refactor sketch flagged by Claude_

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceConnectorConfig.java
```diff
@@ -169,6 +169,26 @@ public class JdbcSourceConnectorConfig extends AbstractConfig {
...
+      + "(json/jsonb, hstore, and native arrays) to complex-aware Connect representations "
+      + "instead of plain STRING: json/jsonb become a logical JSON STRING, hstore becomes a "
+      + "logical JSON STRING or a Connect Map per hstore.handling.mode, and arrays become a "
+      + "Connect ARRAY of the mapped element type. When disabled (the default), json/jsonb "
+      + "and hstore are emitted as STRING and array columns are skipped, preserving "
+      + "backwards compatibility with existing pipelines.";
+  private static final String SQL_COMPLEX_TYPES_ENABLE_DISPLAY = "Enable SQL Complex Types";
+
+  public static final String HSTORE_HANDLING_MODE_CONFIG = "hstore.handling.mode";
+  private static final String HSTORE_HANDLING_MODE_DOC =
```
akanimesh7, line 185: Few comments on the doc 1) sql.complex.types.enable gating is mentioned twice. 2) Doesn't mention about the search_path requirement.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -560,6 +635,9 @@ protected boolean maybeBindPrimitive(
       Object value,
       String fieldName
   ) throws SQLException {
+    if (maybeBindJson(statement, index, schema, value)) {
```
akanimesh7: nit: `maybeBindJson` only ever fires for a MAP — `isJsonBindCandidate` starts with `isStringToStringMap(schema)`, so every non-MAP schema returns false here and falls through regardless. Better as a `case MAP:` in the switch below, so type dispatch happens in one place:

```java
case MAP:
  if (maybeBindJson(statement, index, schema, value)) {
    return true;
  }
  break;
```

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -394,6 +462,13 @@ protected String getSqlType(SinkRecordField field) {
             field.isPrimaryKey()
         );
         return getSqlType(childField) + "[]";
+      case MAP:
+        if (isStringToStringMap(field.schema())
+            && config instanceof JdbcSinkConfig
+            && ((JdbcSinkConfig) config).sqlComplexTypesEnable) {
```
akanimesh7: nit: `complexTypesEnabled()` at line 401 already wraps exactly this check, but it's hand-rolled here and again in `isJsonBindCandidate` — three sites, two idioms. Can both use the helper ?

Both are sink-only paths (`getSqlType` builds sink DDL, `maybeBindJson` is on the bind path), so `complexTypesEnabled()` also answering for a source config doesn't change behavior at either.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack. Updated as per suggestion.

akanimesh7 (dismissed):

akanimesh7 (approved):

PR #1653, opened 2026-06-17
Add fail-fast validation to incrementing.column.name in query mode for incrementing/timestamp+incrementing modes

KGaneshDatta:
## Problem
In query mode the table querier is constructed with a null `TableId`. For `mode=incrementing` or `timestamp+incrementing` with an empty `incrementing.column.name`, `TimestampIncrementingTableQuerier.findDefaultAutoIncrementingColumn` tries to auto-discover the column and dereferences the null `TableId`, throwing an unrecoverable `NullPointerException` on every poll (INC-10982). The task is killed and restarted indefinitely, yet the misconfiguration is silently accepted at create / `config/validate` time.

```
java.lang.NullPointerException: Cannot invoke
  "io.confluent.connect.jdbc.util.TableId.catalogName()" because "this.tableId" is null
    at TimestampIncrementingTableQuerier.findDefaultAutoIncrementingColumn(:203)
    at TimestampIncrementingTableQuerier.createPreparedStatement(:140)
```

## Solution
Add `validateQueryModeIncrementingColumnRequired()` to `JdbcSourceConnectorValidation`, in the validation chain after `validateQueryConfigs()` and before the DB-touching `validateQuerySemantics()` so it fails fast without needing a connection.

When a custom query (`query` or `query.masked`) is configured **and** the mode tracks an incrementing column (`incrementing` / `timestamp+incrementing`), require a non-empty `incrementing.column.name`. The error is attached to the `incrementing.column.name` config with an actionable message, so the misconfiguration is rejected at create / `config/validate`.

**Scope (deliberately narrow, per review feedback):**
- Limited to the **incrementing column only**, because that is the only shape that NPEs. A missing `timestamp.column.name` does **not** crash — `TimestampTableQuerier` passes `null` for the incrementing column, so the auto-discovery branch is skipped. Such connectors run today (pure `timestamp` re-reads without advancing offset; `timestamp+incrementing` with the incrementing column set tracks correctly), so they are intentionally **not** rejected — avoiding flipping currently-running connectors to a validation error on the next config roll.
- Column mapping configs (`incrementing.column.mapping` / `timestamp.columns.mapping`) are **not** rejected with a query either; they are silently ignored at runtime, and connectors run with that combination today.

This keeps the change strictly NPE-prevention and **not stricter-than-before** for any config that currently runs, so no mothership audit is required to ship.

- [ ] yes
- [x] no

## Test Strategy
Tests in `JdbcSourceConnectorValidationTest`, scoped to the new check:
- **Rejected:** `query` + `incrementing` / `timestamp+incrementing` + empty `incrementing.column.name` — covering unset, whitespace-only, `query.masked`, and the `timestamp+incrementing` both-columns-empty incident repro.
- **Allowed (no error):** `incrementing.column.name` provided; `timestamp+incrementing` with only the incrementing column set (timestamp missing); non-incrementing (`timestamp`) mode in query mode.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan
Merging to master. Forward-only validation hardening; safe to revert (validation-only, no data-path changes).

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
+++ b/src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
@@ -382,10 +382,14 @@ public class JdbcSourceConnectorValidation extends AbstractJdbcConnectorValidati
       String msg = String.format(
           "'%s' must be provided when using mode '%s' or '%s' with a custom query. "
           + "The incrementing column cannot be auto-discovered in query mode because there "
-          + "is no table to inspect.",
+          + "is no table to inspect. Note that '%s' is not applied in query mode (it maps "
+          + "columns by table-name regex, and a custom query has no underlying table to match), "
+          + "so '%s' must be set explicitly.",
           JdbcSourceConnectorConfig.INCREMENTING_COLUMN_NAME_CONFIG,
           JdbcSourceConnectorConfig.MODE_INCREMENTING,
-          JdbcSourceConnectorConfig.MODE_TIMESTAMP_INCREMENTING);
+          JdbcSourceConnectorConfig.MODE_TIMESTAMP_INCREMENTING,
+          JdbcSourceConnectorConfig.INCREMENTING_COLUMN_MAPPING_CONFIG,
+          JdbcSourceConnectorConfig.INCREMENTING_COLUMN_NAME_CONFIG);
       addConfigError(JdbcSourceConnectorConfig.INCREMENTING_COLUMN_NAME_CONFIG, msg);
       log.error(msg);
       return false;
--- a/src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
+++ b/src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
@@ -365,22 +365,12 @@ public class JdbcSourceConnectorValidation extends AbstractJdbcConnectorValidati
   }
 
   /**
-   * Require an explicit {@code incrementing.column.name} in query mode for modes that track an
-   * incrementing column.
+   * Require an explicit {@code incrementing.column.name} for {@code incrementing} /
+   * {@code timestamp+incrementing} modes in query mode, where the column cannot be auto-discovered
+   * (no {@code TableId}) and an empty value NPEs on every poll (INC-10982).
    *
-   * <p>In query mode the querier is constructed without a {@code TableId}, so the connector
-   * cannot auto-discover the incrementing column from table metadata the way it can in table
-   * mode. For mode {@code incrementing} / {@code timestamp+incrementing} with an empty
-   * {@code incrementing.column.name}, {@code TimestampIncrementingTableQuerier
-   * .findDefaultAutoIncrementingColumn} dereferences the null {@code TableId} and hits an
-   * unrecoverable {@link NullPointerException} on every poll (the task is killed and restarted
-   * indefinitely). Fail fast at validation time so the misconfiguration is caught at create /
-   * validate.
-   *
-   * <p>This is scoped to the incrementing column only. A missing {@code timestamp.column.name}
-   * (pure {@code timestamp} mode, or {@code timestamp+incrementing} with the incrementing column
-   * set) does <em>not</em> NPE — {@code TimestampTableQuerier} passes {@code null} for the
-   * incrementing column so the auto-discovery branch is skipped — so it is not rejected here.
+   * <p>Scoped to the incrementing column only: a missing {@code timestamp.column.name} does not
+   * NPE, so it is not rejected here.
    */
   private boolean validateQueryModeIncrementingColumnRequired() {
     if (!config.getQuery().isPresent() || !config.modeUsesIncrementingColumn()) {
--- a/src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
+++ b/src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
@@ -88,8 +88,7 @@ public class JdbcSourceConnectorValidation {
       boolean validationResult = validateMultiConfigs()
           && validateLegacyNewConfigCompatibility()
           && validateQueryConfigs()
-          && validateMappingConfigsNotUsedWithQuery()
-          && validateQueryModeColumnRequirements()
+          && validateQueryModeIncrementingColumnRequired()
           && validateQuerySemantics();
 
       if (validationResult && isUsingNewConfigs()) {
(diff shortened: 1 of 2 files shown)


sangeet259:
Hi @KGaneshDatta  `validateMappingConfigsNotUsedWithQuery` -> have we checked there are no running connectors  right now, where it is running with query mode, but customer has also provided the mapping configs ? because even though the mapping configs mean nothing in query mode, but right now, a connector can run in that combination, we need to make sure no customer is running such a connector. Otherwise it will transition them from running to validation error on roll.

sangeet259:
Also can you add that this is being validated as well to the PR title / description ? from the title and description it only seems that we are adding `validateQueryModeColumnRequirements`

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -361,6 +363,105 @@ private boolean validateQueryConfigs() {
...
+
+  /**
+   * Validate that, in query mode, the mode-dependent column names are explicitly provided.
+   *
+   * <p>In query mode the querier is constructed without a {@code TableId}, so the connector
+   * cannot auto-discover the incrementing column from table metadata the way it can in table
+   * mode. With mode {@code incrementing} / {@code timestamp+incrementing} and an empty
+   * {@code incrementing.column.name}, or mode {@code timestamp} / {@code timestamp+incrementing}
+   * and an empty {@code timestamp.column.name}, the connector hits an unrecoverable
+   * {@link NullPointerException} on every poll (the task is killed and restarted indefinitely).
```
sangeet259: The NPE only fires for an empty `incrementing.column.name`. 

In `TimestampIncrementingTableQuerier.findDefaultAutoIncrementingColumn` the guard is `incrementingColumnName != null && incrementingColumnName.isEmpty()`, and `TimestampTableQuerier` passes `null` for the incrementing column via `super(...)`, so for a pure `timestamp` mode that block is skipped and the NPE site is never reached.

That means two config shapes that this validator would now reject are actually running today, not crashing:
- `timestamp` + `query` + **NO** `timestamp.column.name`: runs, but never advances its offset (re-reads each poll).
- `timestamp+incrementing` + `query` with `incrementing.column.name` set but `timestamp.column.name` missing: runs and tracks correctly by the incrementing column.

So this is a stricter-than-before validation: on the next roll, `validate()` re-runs and these connectors would flip from running to a validation error without the user changing anything. Can we confirm against mothership that no fully-managed connector is currently in either state before we ship the hard rejection? 

Also the javadoc here says the timestamp case also NPEs but it does not, so we need to update that.

src/test/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidationTest.java
```diff
@@ -929,6 +929,297 @@ public void validate_withQueryAndTableFilteringConfigs_setsError() {
     );
   }
 
+  // ========== Query-Mode Column Requirement Tests ==========
+  // These guard against the runtime NPE (INC-10982) that occurs when a custom query is
```
sangeet259: nit: can you add the INC id to the PR title.

src/test/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidationTest.java
```diff
@@ -929,6 +929,297 @@ public void validate_withQueryAndTableFilteringConfigs_setsError() {
...
+    assertErrors(INCREMENTING_COLUMN_NAME_CONFIG, 1);
+    assertErrors(TIMESTAMP_COLUMN_NAME_CONFIG, 1);
+    assertErrorMatches(INCREMENTING_COLUMN_NAME_CONFIG,
+        ".*incrementing column cannot be auto-discovered in query mode.*");
+    assertErrorMatches(TIMESTAMP_COLUMN_NAME_CONFIG,
+        ".*must be provided when using mode.*with a custom query.*");
+  }
+
+  @Test
+  public void validate_withQueryAndIncrementingModeMissingColumn_setsError() {
```
sangeet259: This test does unset + whitespace + query.masked into one method with three props.clear() resets. Lets logically separate cases should be separate tests so a failure points at the case.

sangeet259: currently this means a failure in scenario 1 hides scenarios 2 to 4

sangeet259:
General recommendation on the added tests: split each scenario into its own `@Test` so `@Before` re-runs and you start clean, and each case becomes independently named in the report. Some of the preparation steps are already in the `beforeEach`.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
```diff
@@ -361,6 +363,105 @@ private boolean validateQueryConfigs() {
...
+        log.error(msg);
+        valid = false;
+      }
+    }
+
+    if (config.modeUsesTimestampColumn()) {
+      List<String> timestampColumnName = config.getTimestampColumnName();
+      boolean hasTimestampColumn = timestampColumnName != null
+          && timestampColumnName.stream()
+              .anyMatch(col -> col != null && !col.trim().isEmpty());
```
sangeet259: small readability nit: The incrementing-column emptiness check reads as a plain string test (`name == null || name.trim().isEmpty()`), while the timestamp-column one is a list test (`stream().anyMatch(col -> col != null && !col.trim().isEmpty())`). This is techincally correct since `incrementing.column.name` is a single String and `timestamp.column.name` is a List, so this isn't a bug as such.

That said, could we pull the list emptiness into a small `hasNonBlank(List<String>)` helper? It keeps the timestamp check readable and makes the intent, at least one non-blank column, obvious without the inline stream sitting next to the simpler string check.

KGaneshDatta:
Thanks for the review [Sangeet Mishra (@sangeet259)](https://github.com/sangeet259)
Yes, I agree that the NPE case is not reachable with the timestamp mode and should be scoped only in the incrementing mode. I have made the required changes and added only the required validation.

Also, on adding the validation for the query+incrementing.column.mapping mode - it is possible that few connectors with this configuration might fail. This is a backward incompatible change and hence, I feel it should not be added as of now. Hence, I have removed this validation.

src/test/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidationTest.java
KGaneshDatta: As this is public repo, hence wanted skipped to add INC no. to the title

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
KGaneshDatta: Agreed. I have removed the validation for `timestamp.column.name` and only added for `incrementing.column.name`.

src/test/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidationTest.java
sangeet259: by that reasoning we should also not mention the inc number in the comments as well right ?
to get an concrete answer you can check if there is a precedent of mentioning INC- or RCCA- numbers in PRs.

src/main/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidation.java
sangeet259: did you find any running connectors on those configs in MDB ? regardless I agree that removing the timestamp is the better approach here the same argument will be applicable for CP, and we cannot check the configs of all running connectors for CP.

But as a alternative we should add a warn log in that validate path for timestamp combo instead. That will atleast let the customers know that the configs they passed are not being used.

sangeet259:
for the validations we removed because they could be backwards incompatible for the CP customers, we can track it via a separate ticket so we include them in the next major version change like `11.x` series. It is fine in that case if a the connector goes from running to validate errors in a major version upgrade, they can check and fix it.

src/test/java/io/confluent/connect/jdbc/validation/JdbcSourceConnectorValidationTest.java
sangeet259: Hi @KGaneshDatta bump on this.

KGaneshDatta: Ack. I have added the three separate tests instead of the single test.

sangeet259:
Thanks @KGaneshDatta . LGTM!

sangeet259 (dismissed):

sangeet259 (approved):

PR #1656, opened 2026-06-29
INC-11851: Block JDBC URL injection and dangerous driver properties for MySQL/MariaDB

ahmedashfaq027:
## Summary

- **Security incident INC-11851**: tenants could smuggle dangerous MySQL Connector/J properties via two vectors that both reach `DriverManager.getConnection()` during `Connector.validate()`, which fires unconditionally on connector create and bypasses all template-level validators:
  1. **URL injection** — `connection.url=jdbc:mysql://host/db?allowLoadLocalInfile=true`
  2. **Property injection** — `connection.allowLoadLocalInfile=true` as a connector config (passed through `addConnectionProperties()` to the driver)
- Fixes both vectors for MySQL/MariaDB with a three-layer defence in `MySqlDatabaseDialect`:
  1. **`validateJdbcUrlParams(url)`** — rejects the 7 blocked params in the URL query string, plus `&` before `?` and `#` anywhere (connection.host metacharacter injection)
  2. **`getBlockedJdbcConnectionProperties()`** — throws `ConnectException` if any of the 7 blocked properties appear as `connection.*` connector configs
  3. **Safe-default pins** in `addConnectionProperties()` — forces `allowLoadLocalInfile=false`, `allowLocalInfile=false`, `allowUrlInLocalInfile=false`, `autoDeserialize=false` unconditionally, defeating old Connector/J 5.x driver defaults (which had `allowLoadLocalInfile=true` by default)
- Adds extensible hooks in `GenericDatabaseDialect` so other dialects can opt in; other dialect hardening tracked separately (Oracle JNDI, PostgreSQL socketFactory, etc.)
- 39 new tests: 36 in `MySqlDatabaseDialectTest` (URL rejection, property rejection, safe-pin assertions) + 3 in `GenericDatabaseDialectTest` (hook contracts)
- JIRA: https://confluentinc.atlassian.net/browse/INC-11851

## Blocked properties (MySQL/MariaDB)

| Property | Risk |
|---|---|
| `allowLoadLocalInfile` | Server-driven arbitrary file read on the worker |
| `allowLocalInfile` | Same (MariaDB Connector/J spelling) |
| `allowUrlInLocalInfile` | SSRF — server directs worker to read from arbitrary URLs |
| `autoDeserialize` | BLOBs auto-deserialized as Java objects → RCE (CVE-2019-2692) |
| `queryInterceptors` | Custom interceptor class loaded per-query; enables RCE chain with autoDeserialize |
| `statementInterceptors` | Connector/J 5.x name for queryInterceptors |
| `allowMultiQueries` | Permits `;`-separated statements, amplifying SQL injection |

## Test plan

- [ ] All 39 new security tests pass (`mvn test -Dtest="MySqlDatabaseDialectTest#should*,GenericDatabaseDialectTest#addConnection*,GenericDatabaseDialectTest#validateJdbc*" -Dcloud -Dcheckstyle.skip=true`)
- [ ] AppSec (Ritish Singh) to validate fix blocks the live exploit reproduced during incident
- [ ] No regression in existing tests (pre-existing Mockito `ClassImposterizer` failures are unrelated and pre-date this change)

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
@@ -102,6 +102,9 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
   static {
     // TreeSet with CASE_INSENSITIVE_ORDER so blocked.contains() matches regardless of casing
     Set<String> set = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+    // Known Connector/J aliases for each blocked property are listed explicitly.
+    // Connector/J also has a PropertyKey enum that can canonicalise unknown aliases; if
+    // new aliases are discovered in a future driver version, resolve them through that enum.
     set.addAll(Arrays.asList(
         // file read / SSRF
         "allowLoadLocalInfile",
@@ -334,6 +337,9 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
    */
   @Override
   protected void validateJdbcUrlParams(String url) {
+    if (url == null) {
+      return;
+    }
     // # is never valid in a JDBC URL
     if (url.indexOf('#') >= 0) {
       throw new ConnectException(
--- a/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
@@ -31,7 +31,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
@@ -88,7 +87,8 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
    * but a future Connector/J upgrade or new classpath dependency could change that.
    * Blocked now so the denylist does not silently regress.
    * <ul>
-   *   <li>{@code socketFactory} — arbitrary socket factory class → SSRF / traffic interception.</li>
+   *   <li>{@code socketFactory} — arbitrary socket factory class → SSRF / traffic
+   *       interception.</li>
    *   <li>{@code authenticationPlugins} / {@code defaultAuthenticationPlugin} — custom auth
    *       class loaded on every connect.</li>
    *   <li>{@code clientInfoProvider} — custom client-info class loaded on connect.</li>
@@ -98,6 +98,7 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
    * </ul>
    */
   private static final Set<String> MYSQL_BLOCKED_JDBC_PROPERTIES;
+
   static {
     // TreeSet with CASE_INSENSITIVE_ORDER so blocked.contains() matches regardless of casing
     Set<String> set = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
@@ -137,6 +138,7 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
    * even if the blocklist above is somehow bypassed.
    */
   private static final Map<String, String> MYSQL_SAFE_PROPERTY_PINS;
+
   static {
     Map<String, String> pins = new HashMap<>();
     pins.put("allowLoadLocalInfile", "false");
--- a/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
@@ -356,7 +356,13 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
               + "possible metacharacter injection in connection host.");
     }
 
-    // Check authority section property bags: (host=h,port=p,prop=val,...)
+    checkAuthorityBags(preQuery);
+    if (queryStart >= 0) {
+      checkQueryString(url.substring(queryStart + 1));
+    }
+  }
+
+  private void checkAuthorityBags(String preQuery) {
     Matcher bagMatcher = AUTHORITY_PROPERTY_BAG_PATTERN.matcher(preQuery);
     while (bagMatcher.find()) {
       for (String prop : bagMatcher.group(1).split(",")) {
@@ -366,16 +372,15 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
         }
       }
     }
+  }
 
-    // Check standard ?key=value&key=value query string
-    if (queryStart >= 0) {
-      for (String param : url.substring(queryStart + 1).split("&")) {
-        if (param.isEmpty()) {
-          continue;
-        }
-        int eq = param.indexOf('=');
-        checkBlockedKey(eq > 0 ? param.substring(0, eq) : param);
+  private void checkQueryString(String queryString) {
+    for (String param : queryString.split("&")) {
+      if (param.isEmpty()) {
+        continue;
       }
+      int eq = param.indexOf('=');
+      checkBlockedKey(eq > 0 ? param.substring(0, eq) : param);
     }
   }
 
--- a/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
@@ -67,18 +67,34 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
    * via {@code PropertyKey.normalizeCase()} before applying them, so {@code ALLOWLOADLOCALINFILE}
    * and {@code allowLoadLocalInfile} are treated identically by the driver.
    *
+   * <p><b>Currently-exploitable properties:</b>
    * <ul>
    *   <li>{@code allowLoadLocalInfile} / {@code allowLocalInfile} — server-driven
    *       {@code LOAD DATA LOCAL INFILE} reads arbitrary worker files.</li>
    *   <li>{@code allowUrlInLocalInfile} — same mechanism via arbitrary URLs → SSRF.</li>
-   *   <li>{@code allowLoadLocalInfileInPath} — restricts file read to a path; setting to
-   *       {@code /} is equivalent to {@code allowLoadLocalInfile=true}.</li>
+   *   <li>{@code allowLoadLocalInfileInPath} — path-restricted variant;
+   *       setting to {@code /} is equivalent to {@code allowLoadLocalInfile=true}.</li>
    *   <li>{@code autoDeserialize} — BLOBs auto-deserialized as Java objects; a malicious
    *       server can craft a gadget chain → RCE (see CVE-2019-2692).</li>
    *   <li>{@code queryInterceptors} / {@code statementInterceptors} — custom interceptor
-   *       classes run per query; combined with {@code autoDeserialize} → RCE on connect.</li>
+   *       classes loaded per query; combined with {@code autoDeserialize} → RCE on connect.</li>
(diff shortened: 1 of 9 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -199,6 +252,69 @@ public String buildUpsertQueryStatement(
...
+    if (queryStart == -1) {
+      return;
+    }
+    String queryString = url.substring(queryStart + 1);
+    for (String param : queryString.split("&")) {
+      if (param.isEmpty()) {
+        continue;
+      }
+      String key = param.contains("=") ? param.substring(0, param.indexOf('=')) : param;
+      if (MYSQL_BLOCKED_JDBC_PROPERTIES.contains(key)) {
```
ritishkumarsingh: The driver matches property names case-insensitively (PropertyKey.caseInsensitiveValues + normalizeCase(), used by ConnectionUrl.collectProperties). This check is case-sensitive, so ?ALLOWLOADLOCALINFILE=true (or AutoDeserialize, etc.) bypasses it but is still honored. Normalize the key before comparing (e.g. key.toLowerCase(Locale.ROOT)) and make the blocklist case-insensitive. Also URL-decode the key first — the driver decodes %xx, so allowLoad%4cocalInfile-style encodings evade a raw string match.

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -199,6 +252,69 @@ public String buildUpsertQueryStatement(
...
+   * We reject any {@code &} that appears before the {@code ?} query separator, and any {@code #}
+   * fragment identifier, neither of which has a legitimate place in a MySQL JDBC URL.
+   */
+  @Override
+  protected void validateJdbcUrlParams(String url) {
+    int queryStart = url.indexOf('?');
+    // & must not appear before the query-string delimiter — that would mean a metacharacter
+    // was injected into the host/path portion (e.g. connection.host="host&allowLoadLocalInfile=true")
+    String preQuery = queryStart >= 0 ? url.substring(0, queryStart) : url;
+    if (preQuery.indexOf('&') >= 0) {
```
ritishkumarsingh, line 353: Good for the &-smuggled host, but the authority key-value form uses , and () as separators, not &, so this won't catch (host=…,allowLoadLocalInfile=true). Pair it with the authority parsing above.

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -49,6 +57,51 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
...
+   *   <li>{@code allowUrlInLocalInfile} — same mechanism via arbitrary URLs → SSRF.</li>
+   *   <li>{@code autoDeserialize} — BLOBs auto-deserialized as Java objects; a malicious
+   *       server can craft a gadget chain → RCE (see CVE-2019-2692).</li>
+   *   <li>{@code queryInterceptors} / {@code statementInterceptors} — custom interceptor
+   *       classes run per query; combined with {@code autoDeserialize} → RCE on connect.</li>
+   *   <li>{@code allowMultiQueries} — permits multiple {@code ;}-separated statements,
+   *       amplifying any SQL-injection impact.</li>
+   * </ul>
+   */
+  private static final Set<String> MYSQL_BLOCKED_JDBC_PROPERTIES =
```
ritishkumarsingh: Make this a case-insensitive set (e.g. new TreeSet<>(String.CASE_INSENSITIVE_ORDER) wrapped unmodifiable, or compare on normalized keys) so both the addConnectionProperties check and validateJdbcUrlParams are case-proof in one place.

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -49,6 +57,51 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
...
+      Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+          // file read / SSRF
+          "allowLoadLocalInfile",
+          "allowLocalInfile",       // MariaDB Connector/J spelling
+          "allowUrlInLocalInfile",
```
MananPopat7: We should also add allowLoadLocalInfileInPath

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -49,6 +57,51 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
...
+   * user-provided connection.* props are processed, so a driver default can never be dangerous
+   * even if the blocklist above is somehow bypassed.
+   */
+  private static final Map<String, String> MYSQL_SAFE_PROPERTY_PINS;
+  static {
+    Map<String, String> pins = new HashMap<>();
+    pins.put("allowLoadLocalInfile", "false");
+    pins.put("allowLocalInfile", "false");
+    pins.put("allowUrlInLocalInfile", "false");
+    pins.put("autoDeserialize", "false");
```
MananPopat7, line 150: I think we should also get it reviewed by the SMEs (@KGaneshDatta), since it changes CP behaviour.

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -199,6 +252,69 @@ public String buildUpsertQueryStatement(
...
+    return MYSQL_BLOCKED_JDBC_PROPERTIES;
+  }
+
+  @Override
+  protected Properties addConnectionProperties(Properties properties) {
+    // Super checks user-provided connection.* props against the blocklist and passes safe ones.
+    Properties result = super.addConnectionProperties(properties);
+    // Pin safe values last — guards against old Connector/J 5.x driver defaults
+    // (allowLoadLocalInfile was true by default pre-8.x) even if the URL or blocklist
+    // is somehow bypassed. URL params take precedence over Properties, so validateJdbcUrlParams
```
MananPopat7: Isn't it vice-versa? Properties override the URL

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -49,6 +61,67 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
...
+   *       {@code /} is equivalent to {@code allowLoadLocalInfile=true}.</li>
+   *   <li>{@code autoDeserialize} — BLOBs auto-deserialized as Java objects; a malicious
+   *       server can craft a gadget chain → RCE (see CVE-2019-2692).</li>
+   *   <li>{@code queryInterceptors} / {@code statementInterceptors} — custom interceptor
+   *       classes run per query; combined with {@code autoDeserialize} → RCE on connect.</li>
+   *   <li>{@code allowMultiQueries} — permits multiple {@code ;}-separated statements,
+   *       amplifying any SQL-injection impact.</li>
+   * </ul>
+   */
+  private static final Set<String> MYSQL_BLOCKED_JDBC_PROPERTIES;
```
ritishkumarsingh, line 100: **Defense in Depth:**

Denylist omits the class-loading properties: socketFactory, authenticationPlugins / defaultAuthenticationPlugin, clientInfoProvider, serverRSAPublicKeyFile, propertiesTransform. On mysql-connector-j 8.0.33 these are not currently weaponizable (the driver loads them via Util.getInstance with forName(initialize=false) + isAssignableFrom before construction, and there's no third-party gadget on the worker classpath — verified), so this is not an exploit today. But a denylist is inherently incomplete: a Connector/J upgrade, a new classpath dependency, or simply a property you didn't enumerate reopens it.

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -49,6 +60,94 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
...
+    Set<String> set = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+    set.addAll(Arrays.asList(
+        // file read / SSRF
+        "allowLoadLocalInfile",
+        "allowLocalInfile",             // MariaDB Connector/J spelling
+        "allowUrlInLocalInfile",
+        "allowLoadLocalInfileInPath",   // path-restricted variant; / == full access
+        // RCE via Java deserialization
+        "autoDeserialize",
+        "queryInterceptors",
```
ritishkumarsingh, line 116: The blocklist is a hand-maintained set of name spellings. You already had to add two alias pairs manually (statementInterceptors↔queryInterceptors, allowLocalInfile↔allowLoadLocalInfile) — which is exactly the failure mode: if any blocked property has another accepted alias you didn't enumerate, a tenant uses the alias and bypasses both the blocklist and validateJdbcUrlParams, while the driver still honors it. Case is handled, aliases are not. More robust: resolve each key through the driver's own canonicalization (PropertyKey.normalizeCase() / PropertyKey.fromValue(key).getKeyName()) and block on the canonical key — that collapses case and aliases in one place, so you don't have to chase spellings. Low priority (no known unlisted alias for the dangerous props today), but it removes the maintenance trap.

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -199,6 +298,99 @@ public String buildUpsertQueryStatement(
...
+   *
+   * <p>Keys are URL-decoded and compared case-insensitively because MySQL Connector/J normalises
+   * property names via {@code PropertyKey.normalizeCase()} and decodes {@code %xx} sequences
+   * before applying them — so {@code ALLOWLOADLOCALINFILE} and {@code allow%4CoadLocalInfile}
+   * are treated identically by the driver.
+   */
+  @Override
+  protected void validateJdbcUrlParams(String url) {
+    // # is never valid in a JDBC URL
+    if (url.indexOf('#') >= 0) {
```
ritishkumarsingh, line 344: url.indexOf('#') will NPE if jdbcUrl is ever null. Add a null guard (cheap robustness).

src/main/java/io/confluent/connect/jdbc/dialect/MySqlDatabaseDialect.java
```diff
@@ -49,6 +60,94 @@ public class MySqlDatabaseDialect extends GenericDatabaseDialect {
...
+        // SQL injection amplification
+        "allowMultiQueries",
+        // class-loading / file-read (defence-in-depth; not weaponisable today but
+        // could regress with a Connector/J upgrade or new worker classpath entry)
+        "socketFactory",
+        "authenticationPlugins",
+        "defaultAuthenticationPlugin",
+        "clientInfoProvider",
+        "propertiesTransform",
+        "serverRSAPublicKeyFile"
```
ritishkumarsingh, line 127: serverRSAPublicKeyFile in the blocklist: fine to keep, but note it's a weak vector — the driver reads it as an RSA public key to encrypt the password, it doesn't return the file content to the server, so it isn't a real exfil primitive. Harmless inclusion.

ritishkumarsingh (approved):

ritishkumarsingh (changes requested):
The validation/blocklist is implemented only in MySqlDatabaseDialect (which serves both mysql and mariadb subprotocols). Every other dialect — Postgres, SQL Server, Oracle, SQLite, DB2, etc. — inherits the no-op base:

```
GenericDatabaseDialect.validateJdbcUrlParams(url) → empty body
GenericDatabaseDialect.getBlockedJdbcConnectionProperties() → Collections.emptySet()
```

So for the non-MySQL dialects, this PR does not validate the connection URL and does not block any driver property. They still flow through the same getConnection() chokepoint — the guard there just does nothing for them.

PR #1657, opened 2026-07-01
Add opt-in, admin-configured JDBC URL parameter blocklist

KGaneshDatta:
## Problem

The user-provided `connection.url` is passed verbatim to `DriverManager.getConnection(...)`, so a user can make use of dangerous JDBC driver parameters through it. For example:

- `allowLoadLocalInfile=true` / `allowUrlInLocalInfile=true` → arbitrary worker file read / SSRF via `LOAD DATA LOCAL INFILE`
- `autoDeserialize=true` → deserialization RCE

Unlike Confluent Cloud, self-managed CP should **not** block these parameters by default, since customers may legitimately use them. It is required to restrict them by opt-in and controlled by the cluster admin and not by the user creating a new connector with exploitable params.

## Solution

Add an **opt-in, admin-configured blocklist** of JDBC URL parameter names, enforced entirely within the connector.

- **Opt-in:** unset/empty ⇒ no restriction, so default behaviour is unchanged.
- **Admin-controlled, not user-controllable:** the blocklist is read from the `jdbc.connection.url.blocked.params` JVM system property (or the `JDBC_CONNECTION_URL_BLOCKED_PARAMS` environment variable), which the worker operator sets at JVM startup. A connector plugin cannot read worker config, and this value cannot be supplied through a connector's REST config — so a tenant can neither see nor change it. Setting it identically on every worker makes it apply cluster-wide.
- **Connector-only, no runtime changes:** the check runs in `GenericDatabaseDialect.getConnection()` — the single point where the URL is handed to the driver — so it covers both config validation and task runtime, for every dialect.

Enforcement behaviour:
- A **new** connector whose `connection.url` contains a blocked parameter is rejected at validation time (HTTP 400, never created).
- A **pre-existing** connector fails its tasks on the next connection attempt once the policy is enabled.

Example worker configuration:

```
KAFKA_OPTS="-Djdbc.connection.url.blocked.params=allowLoadLocalInfile,allowUrlInLocalInfile,autoDeserialize"
```

## Testing

**Unit tests** (`GenericDatabaseDialectTest`):
- `validateJdbcUrlParamsIsNoOpWhenAdminHasNotOptedIn` — no property set ⇒ nothing blocked (including a URL carrying `allowLoadLocalInfile`) and `null` URL is safe.
- `validateJdbcUrlParamsBlocksConfiguredParams` — with a blocklist configured: clean URL allowed; blocked param rejected via query-string, semicolon, and parenthetical-bag syntaxes; case-insensitive and URL-encoded (`%61llow...`) variants rejected; a param not on the list still allowed.

**Manual end-to-end on local Confluent Platform**

| Scenario | Blocklist | Result |
|---|---|---|
| Source, `?allowLoadLocalInfile=true` | off (default) | Created / not blocked — opt-in confirmed ✅ |
| Source, clean URL | on | HTTP 201, RUNNING ✅ |
| Source, `?allowLoadLocalInfile=true` | on | Rejected (HTTP 400): "…blocked connection parameter 'allowLoadLocalInfile'…" ✅ |
| Source, `?%61llowLoadLocalInfile=true` (URL-encoded) | on | Rejected — decoded and caught ✅ |
| Sink, clean URL | on | HTTP 201, RUNNING ✅ |
| Sink, `?autoDeserialize=true` | on | Rejected ✅ |

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/GenericDatabaseDialect.java
@@ -34,6 +34,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.math.BigDecimal;
 import java.net.URL;
+import java.net.URLDecoder;
 import java.nio.ByteBuffer;
 import java.sql.Blob;
 import java.sql.Clob;
@@ -62,8 +63,11 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Queue;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import io.confluent.connect.jdbc.dialect.DatabaseDialectProvider.FixedScoreProvider;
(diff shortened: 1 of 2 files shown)


sangeet259 (approved):
Looks good to me.

PR #1659, opened 2026-07-07
CC-41849: fill JDBC source test-suite gaps 

sangeet259:
## What

Fills the test-coverage gaps catalogued in the CC-41849 testing-gaps audit. All test-only; no connector production code is changed.

New source integration tests (each extends the existing `BaseConnectorIT` / testcontainers pattern):

- `MySqlJdbcSourceConnectorIT`, `MsSqlJdbcSourceConnectorIT` — run the shared source mode matrix (bulk, incrementing, timestamp, timestamp+incrementing) on MySQL (embedded MariaDB) and SQL Server, closing the gap where only Postgres exercised it.
- `PostgresNumericMappingIT` — asserts the emitted Connect schema type per `numeric.mapping` mode.
- `PostgresSchemaPatternIT` — `schema.pattern` narrowing plus the duplicate-unqualified-name failure.
- `PostgresTableLifecycleIT` — a table created mid-run is discovered and streamed.
- `PostgresSchemaEvolutionIT` — a mid-stream `ALTER TABLE ADD COLUMN` appears in later records.
- `PostgresQueryModeIT` — query mode end to end, including a `query.suffix` WHERE clause that must actually filter.
- `PostgresMultiTaskIT` — `tasks.max=2` distributes two tables and both stream.
- `PostgresOffsetResumeIT` — restarts the connector and asserts committed rows are not re-read, the end-to-end guard for the INC-11312 duplication class.

Plus a unit guard, `testKnownGoodConfigsPassValidation` in `JdbcSourceConnectorTest`, that pins representative documented configs (including query mode) as passing `validate()`, guarding the regression class behind GitHub issue 1591.

## Why

Follow-on to INC-11312, where the documented config path was never tested, so a driver metadata change broke the fleet while CI stayed green. This closes the sibling gaps the audit found.

## Why safe

Additive test code only; no production classes modified. Validated green on the full CI suite (whole test set plus the JaCoCo coverage gate).

## Notes

- Supersedes #1658. That PR's branch name contained a `-f` substring (`...-fill-...`) that tripped a CI push permission glob; this branch is renamed to avoid it. Same content.
- Incrementing-mode ITs assert exact record counts (over-production guard); bulk-mode ITs assert `>=` because bulk re-emits the row set every poll by design.

Ticket: CC-41849


src/test/java/io/confluent/connect/jdbc/integration/PostgresSchemaPatternIT.java
```diff
@@ -0,0 +1,176 @@
...
+    // Narrowing to app1 leaves a single "orders" table, so the connector runs and streams app1's rows.
+    props.put(JdbcSourceConnectorConfig.SCHEMA_PATTERN_CONFIG, SCHEMA_ONE);
+    connect.kafka().createTopic(TOPIC, 1);
+
+    connect.configureConnector(CONNECTOR_NAME, props);
+    waitForConnectorToStart(CONNECTOR_NAME, 1);
+
+    ConsumerRecords<byte[], byte[]> records = connect.kafka().consume(2, CONSUME_TIMEOUT_MS, TOPIC);
+    assertTrue("Should stream app1's two rows once schema.pattern narrows discovery",
+        records.count() >= 2);
```
KGaneshDatta: The comment says "app2 must never reach the topic when narrowed to app1," but the assertion is only records.count() >= 2 — it never checks that app2's rows (a2_*) are absent. 
Can you check whether we can drain a couple of polls and assert no value contains a2_

KGaneshDatta (commented):
Thanks for the PR @sangeet259 
Few points can you check.
1. There are few features like db.timezone, quote.sql.identifiers etc. which are included. Please check if we can add them in the tests also.
2. I see the Oracle Dialect based test cases are not included. If there any reason for it.
3. Most of the feature tests are for Postgres. Do we need to add the same for other MSSql Server and MySQL dialects also ? If required please add them otherwise please ignore.

src/test/java/io/confluent/connect/jdbc/integration/PostgresSchemaPatternIT.java
sangeet259: Addressed. The test now inspects every consumed record and explicitly asserts that no  `a2_`  row reaches the topic.

sangeet259:
@KGaneshDatta , Thanks for the review. I added coverage for `db.timezone`, `quote.sql.identifiers`, and all four Oracle source modes. I kept generic feature tests on PostgreSQL to avoid duplicating identical behavior across MySQL and SQL Server, the dialect-specific cases can be added where behavior differs.

KGaneshDatta (approved):

PR #1660, opened 2026-07-09
Add PostgreSQL JSON/JSONB complex type support

KGaneshDatta:
Builds on the complex-types foundation (#1651) to map PostgreSQL `json`/`jsonb` columns.

## Problem

`json`/`jsonb` columns arrive on the topic as untagged STRING literals. The text itself is correct and lossless, but nothing on the topic says it is JSON, so every downstream consumer has to know out of band to re-parse it — and on the way back out a sink has no way to tell such a value apart from ordinary text, so it provisions `TEXT` and the document lands in a column that cannot be queried as JSON.

## Solution

When `sql.complex.types.enable=true`, a `json`/`jsonb` column maps to a **lossless logical JSON STRING** — `io.confluent.connect.jdbc.data.Json`, aligned with Debezium's `io.debezium.data.Json` — instead of a plain STRING. The raw document text is unchanged; only the schema gains the tag. On the sink that tag is what lets the value round-trip into a native `jsonb` column, bound as text through a `::jsonb` cast.

- **Source:** `addFieldToSchema` resolves the column's schema through `jsonSchema()`, which returns `Json.optionalSchema()` or `Json.schema()` according to the column's nullability. `columnConverterFor` is deliberately **untouched**: the value is the document's raw text either way, so only the schema changes.
- **Sink:** the `Json` logical type maps to `JSONB` in `getSqlType`. That mapping lives on **#1651** together with the `::jsonb` cast, since #1661 needs it too for `hstore` in `json` mode; this PR is what makes a `json`/`jsonb` column produce the tag in the first place.

When the feature is disabled (the default), `json`/`jsonb` remain a plain STRING and the sink provisions `TEXT` — unchanged behaviour.

`json`/`jsonb` is always represented as a STRING, aligned with Debezium. There is no Map representation and no `json.handling.mode` config.

### Mixed configurations

Both directions are intentional and covered:

- **Source enabled, sink disabled:** the sink ignores the tag and writes `text`, so the document is preserved verbatim but is not queryable as JSON. Promoting it to `jsonb` on a sink whose operator never opted in was the rejected alternative.
- **Source disabled, sink enabled:** the value is an untagged STRING, indistinguishable from text, so it also lands in `text`.

- [ ] yes
- [x] no

## Test Strategy

Unit tests cover the source schema for both column nullabilities — asserting the type, the logical name and the optional flag together — that the value stays the document's raw text rather than a re-serialized or projected form, and that the sink maps the `Json` tag to `JSONB`.

Integration tests run against a real PostgreSQL: `json` and `jsonb` emitted verbatim, non-object documents (arrays, scalars, `null`) preserved, a SQL NULL column, a `NOT NULL` column emitting a non-optional schema, and the flag-off plain-STRING case. Source→Kafka→sink round-trips cover several document shapes landing in `jsonb`, the flag-off case landing in `text`, both mixed-configuration directions above, a malformed JSON document being reported rather than silently written, and a plain STRING written into an existing `jsonb` column.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Cut from and merges into the base branch (#1651); ships with that foundation. Default-off (`sql.complex.types.enable=false`), so the change is backward compatible and safe to roll back.


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -387,6 +388,44 @@ private boolean complexTypesEnabled() {
...
+    if (jsonAsString()) {
+      SchemaBuilder jsonBuilder = Json.builder();
+      if (optional) {
+        jsonBuilder.optional();
+      }
+      return jsonBuilder.build();
+    }
+    SchemaBuilder mapBuilder = SchemaBuilder.map(
+        Schema.STRING_SCHEMA, Schema.OPTIONAL_STRING_SCHEMA);
+    if (optional) {
```
akanimesh7: Dug into where the MAP mode comes from. The [PRD (INIT-11203) §5.1](https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/5127471188/Product+Requirements+-+Support+additional+datatypes+with+JDBC+Source+and+Sink+Connectors#5.-Technical-Requirements-&-Database-Specifics) mentions JSON → Map, but as `Map<String, Any>` for a *dynamic* schema (and a `Struct` when the schema is known) — not `Map<String,String>`. And the driving use case (§3A, IoT sensor data) explicitly wants a **Struct** so ksqlDB can filter on **nested** fields like `sensor_data.temperature > 50`.

This `Map<String,String>` can't do that: with the Level-1 projection, nested objects become escaped JSON strings and scalars turn into strings (`42` → `"42"`) — which the [design doc §2.2](https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/5624889731/Support+additional+datatypes+with+JDBC+Source+and+Sink+Connectors#2.2-JSON-/-JSONB) "Limitation" section itself calls out. So it matches neither what the PRD asked for (`Struct` / `Map<String,Any>`) nor the use case it exists for, yet it still adds a `json.handling.mode` config. Debezium also maps json/jsonb to `io.debezium.data.Json` (a STRING).

What's the reasoning for adding the MAP mode here?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -387,6 +388,44 @@ private boolean complexTypesEnabled() {
     return false;
   }
 
+  /**
+   * Whether PostgreSQL json/jsonb columns should be emitted as a logical JSON STRING (mode
+   * {@code string}) rather than a Connect Map (mode {@code map}, the default). Only the source
+   * connector exposes this.
+   */
+  private boolean jsonAsString() {
+    if (config instanceof JdbcSourceConnectorConfig) {
```
akanimesh7, line 466: why this check ? this code path would come only in source connectors right ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -387,6 +388,44 @@ private boolean complexTypesEnabled() {
...
+      return JdbcSourceConnectorConfig.JSON_HANDLING_MODE_STRING
+          .equalsIgnoreCase(config.getString(JdbcSourceConnectorConfig.JSON_HANDLING_MODE_CONFIG));
```
akanimesh7: config.getString is an anti-pattern it should be config.xyzFeatureEnabled()

akanimesh7 (commented):
`JsonConverter.jsonStringToMap` (from base PR #1651, so noting here rather than inline): to be re-considered after we decide whether MAP is needed.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Right. I have removed this check.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack.
I have added the new method jsonHandlingModeIsString() for getting checking the config value. But, as the config is individually defined for both Sink and Source connectors, added the separate reference.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Currently, by default, a Postgres Source json/jsonb column is published to the topic as a Connect MAP<STRING,STRING>.

But this currently handles only the top Level-1 projection. In cases of nested JSON, those are by default mapped as STRINGS and hence while sinking they are being treated as STRINGS only instead of JSON.

As, this has been implemented with new logical type, it can be only known to the JDBC Sink connectors.

But there might be few cases where users have only Level-1 projection and wanted to sink the data to other sources like KSqlDB and Flink or other sink connectors, they should have the common schema which can be achieved by this MAP mode. 

Hence, it is required to add the MAP mode and the Product is also aligned with this.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +788,79 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+    Struct value = new Struct(schema).put("a", 1);
+
+    sinkDialect().bindField(statement, 3, schema, value, colDef, "field");
+
+    verify(statement).setString(3, "{\"a\":1}");
+  }
+
+  // ----- complex-type test helpers -----
+
+  private PostgreSqlDatabaseDialect complexTypesDialect(String... extraProps) {
```
akanimesh7: both usages of this don't pass any inputs. In that case, let's simplify this ?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +788,79 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+    System.arraycopy(extraProps, 0, props, 2, extraProps.length);
+    return new PostgreSqlDatabaseDialect(sourceConfigWithUrl("jdbc:postgresql://something", props));
+  }
+
+  private PostgreSqlDatabaseDialect sinkDialect() {
+    return new PostgreSqlDatabaseDialect(sinkConfigWithUrl(
+        "jdbc:postgresql://something", JdbcSinkConfig.SQL_COMPLEX_TYPES_ENABLE, "true"));
+  }
+
+  private Schema sourceFieldSchema(
```
akanimesh7: function used only once, why not make it inline then ? Also, this looks like a useful function, please check if some variation of this is already present in the code.

akanimesh7 (approved):

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -408,6 +405,23 @@ private boolean complexTypesEnabled() {
     return false;
   }
 
+  /**
+   * Build the Connect schema for a PostgreSQL json/jsonb column. When complex types are disabled
+   * the column stays a plain STRING; when enabled it is a logical JSON STRING (raw text, aligned
+   * with Debezium's {@code io.debezium.data.Json}).
+   */
+  private Schema jsonSchema(ColumnDefinition columnDefn) {
```
akanimesh7, line 477: nit: `Json` already exposes the factories — `Json.optionalSchema()` and `Json.schema()` — and that's what #1661's `hstoreSchema()` and #1662's `hstore[]`/`json[]` element cases both call. Can this use them too, so the sites read the same way ?

```java
private Schema jsonSchema(ColumnDefinition columnDefn) {
  boolean optional = columnDefn.isOptional();
  if (!complexTypesEnabled()) {
    return optional ? Schema.OPTIONAL_STRING_SCHEMA : Schema.STRING_SCHEMA;
  }
  return optional ? Json.optionalSchema() : Json.schema();
}
```

_🤖 flagged by Claude_

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -797,7 +801,74 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+
+  @Test
+  public void shouldMapComplexTypesToSqlTypes() {
+    // Scalar logical JSON STRING -> native JSONB, on a sink with complex types enabled.
+    assertEquals("JSONB",
+        sinkDialect().getSqlType(new SinkRecordField(Json.schema(), "col", false)));
+  }
+
+  @Test
+  public void shouldBindMapValueAsJsonStringForJsonbColumn() throws Exception {
```
akanimesh7: nit: this binds a `MAP<STRING,STRING>` through `maybeBindJson` — both `maybeBindJson` and `isStringToStringMap` are added by #1651, not here — and the shape it exercises is hstore's, not json/jsonb. Nothing in it touches what this PR changes. Would it sit better on #1651 where the code lives, or #1661 whose feature produces that shape ?

Same thing more mildly for `shouldMapComplexTypesToSqlTypes`: `Json.LOGICAL_NAME` -> JSONB is also #1651's. It's defensible here as a round-trip contract check for this PR's schema change — worth saying so in the comment if that's the intent.

_🤖 flagged by Claude_

akanimesh7 (commented):
Approach looks good to me — the main-code change is minimal and `columnConverterFor` correctly stays untouched, since the raw text is the same either way and only the tag changes. Tests are thorough too: `testJsonPreservesNonObjectDocuments`, `testSinkReportsMalformedJsonText` and both mixed-config directions (`testSourceEnabledSinkDisabledLandsInTextWithoutDataLoss` / `testSourceDisabledSinkEnabledLandsInText`) are exactly the cases I'd have asked for.

One thing to fix before merge — the description has drifted, mostly from STRUCT support being dropped:

| Description says | Actual |
|---|---|
| "`addFieldToSchema` / **`columnConverterFor`** emit the `Json` logical STRING via `jsonSchema()`" | `columnConverterFor` isn't touched by this PR and doesn't call `jsonSchema()` |
| "`getSqlType` maps the `Json` logical type (**and STRUCT/MAP**) to `JSONB`" | There's no `case STRUCT:` — only `case MAP:`, gated on `isStringToStringMap` |
| Test Strategy cites **`shouldBindStructValueAsJsonStringForJsonbColumn`** | That test doesn't exist; it's `shouldBindMapValueAsJsonStringForJsonbColumn` now |
| Problem: "a Connect MAP/**STRUCT** cannot be written into a native JSON/JSONB sink column" | STRUCT still can't, by design now — so the problem statement promises something this PR doesn't deliver |

akanimesh7 (approved):

PR #1661, opened 2026-07-09
Add PostgreSQL HSTORE complex type support

KGaneshDatta:
Builds on the complex-types foundation (#1651) to map PostgreSQL `hstore` columns per `hstore.handling.mode`.

## Problem

`hstore` columns are silently dropped from the topic today — the connector emits a WARN and produces no field for them, so customers using hstore columns receive no data at all, typically discovered only through downstream data drift.

## Solution

When `sql.complex.types.enable=true`, an `hstore` column maps per `hstore.handling.mode`, following Debezium's config of the same name:

- **`map`** (default): a Connect `MAP<STRING,STRING>` whose values are optional, since an hstore value may be NULL.
- **`json`**: a `Json`-tagged STRING carrying the hstore as a JSON object.

**Both representations are provisioned as a native `jsonb` column on the sink**, so the mode changes the on-topic shape and nothing about the destination. `jsonb` rather than a native `hstore` column because `hstore` is a non-core extension that may not exist — or may not be visible — on the destination, whereas `jsonb` is always available.

Details:

- **Source:** `addFieldToSchema` and `columnConverterFor` resolve the representation through the shared `hstoreSchema(boolean optional)` helper on the base branch, so the scalar column path here and the array element path in #1662 cannot drift apart.
- **Sink:** a `MAP<STRING,STRING>` is serialized to JSON text and bound through the `::jsonb` cast; a `Json` STRING is bound as text through the same cast. Generic STRUCT values and other map shapes are rejected on both the DDL and the bind path.
- **Unexpected driver values:** any non-`Map` value follows Debezium's `handleUnknownData` — a nullable column degrades to null with a WARN, a NOT NULL column fails, and the value itself is never logged in case it is sensitive.

When the feature is disabled (the default), an `hstore` column is **skipped and produces no field**, exactly as today — unchanged behaviour.

### Known limitation

`hstore` is recognised by its unqualified type name, so the schema owning the extension must be on the connection's `search_path`. Off the `search_path` the driver reports the type as `"schema"."hstore"`, the column is skipped like any unsupported type, and a WARN naming the `search_path` requirement is logged once per column so the cause is actionable from the worker log. Documented on `hstore.handling.mode`.

- [ ] yes
- [x] no

## Test Strategy

Unit tests cover the source schema for both modes and both column nullabilities, case-insensitive type-name matching, that a non-hstore `Types.OTHER` type is not captured by the hstore branch, the flag-off skip, a SQL NULL column in both modes, the unexpected-value degrade and failure paths, the off-`search_path` warning including its warn-once-per-column dedupe, and on the sink the `MAP<STRING,STRING>` → `jsonb` DDL and bind plus the rejection of every other map shape on both paths.

Integration tests run against a real PostgreSQL with the extension installed: single and multiple pairs, NULL values, special characters and quoting, an empty hstore, a SQL NULL column, the driver's `Map` contract, the flag-off drop, and a column whose type is off the `search_path`. Source→Kafka→sink round-trips cover both handling modes landing in `jsonb`, and the flag-off case.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Cut from and merges into the base branch (#1651); ships with that foundation. Default-off (`sql.complex.types.enable=false`), so the change is backward compatible and safe to roll back.


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -330,6 +336,18 @@ protected ColumnConverter columnConverterFor(
           return rs -> rs.getString(col);
         }
 
+        if (complexTypesEnabled() && isHstoreType(columnDefn)) {
```
akanimesh7: What does the JDBC driver actually return for an HSTORE column - a Map, a String, or a PGobject ? pgjdbc only gives a `Map` when it can resolve the hstore OID (it's an extension type with a dynamic OID); otherwise it's a `PGobject` carrying the raw `"k"=>"v"` text. Our only test mocks `getObject` as a Map, so we're assuming the answer rather than establishing it.

Also worth noting the two modes fail very differently in that case: json mode guards it (`connectValueToJson` throws a clear `DataException` for a non-Map), but map mode has no guard - `rs.getObject(col)` passes the value straight through against a declared `MAP<STRING,STRING>` schema, so it blows up deeper inside Connect as an opaque schema/value mismatch that never names hstore.

Can we add an IT against a real Postgres with the extension installed to pin the driver contract ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -396,6 +418,36 @@ private boolean complexTypesEnabled() {
...
+  }
+
+  /**
+   * Build the Connect schema for a PostgreSQL hstore column, honoring
+   * {@code hstore.handling.mode}: a JSON-object STRING when {@code json}, otherwise a
+   * Map&lt;String,String&gt;.
+   */
+  private Schema hstoreSchema(ColumnDefinition columnDefn) {
+    if (hstoreAsJson()) {
+      return columnDefn.isOptional() ? Schema.OPTIONAL_STRING_SCHEMA : Schema.STRING_SCHEMA;
```
akanimesh7: Not using the `Json` class here - how will the sink handle it ? A plain untagged STRING means `getSqlType` sees `schemaName() == null`, skips the logical switch and falls to `case STRING:` -> `TEXT`. So json mode lands in TEXT while map mode lands in JSONB.

Two reasons I think this should be `Json`:
- Debezium tags this as `io.debezium.data.Json` - per section 2.4 of our own design doc, which is the behaviour we say we're following.
- After #1660, a `json`/`jsonb` column always emits a `Json` STRING -> JSONB on the sink. So the same "STRING carrying JSON" concept gets two different sink column types across the two PRs.

The design doc is also self-contradictory here: section 2.3's mode table says `json` -> `TEXT`, but section 2.3's prose says an HSTORE value "lands in a JSONB column ... since the on-topic schema is a generic MAP/STRING" (i.e. both shapes -> JSONB), and section 2.4's summary table says HSTORE -> JSONB unconditionally. Two of the three places say JSONB. Which is intended ?

`Json.optionalSchema()` would make it consistent - `getSqlType` already maps `Json.LOGICAL_NAME` -> JSONB.

One more thing: the rationale in the test comment ("not JSONB, since hstore is a non-core extension that may not exist on the destination") is the doc's reason for using JSONB **instead of a native hstore column** - it doesn't explain TEXT-instead-of-JSONB.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +789,108 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+  @Test
+  public void hstoreMapModeShouldPassThroughDriverMap() throws Exception {
+    // In map mode (the default) the driver's Map is emitted as-is for the Connect MAP schema.
+    Map<String, String> hstore = Collections.singletonMap("env", "prod");
+    assertEquals(hstore,
+        hstoreConverter(complexTypesDialect()).convert(hstoreResultSet(hstore)));
+  }
+
+  @Test
+  public void hstoreShouldMapToSinkSqlTypesPerMode() {
```
akanimesh7: This asserts generic `MAP` -> JSONB and generic `STRING` -> TEXT. Both are pre-existing `getSqlType` behaviour from the base PR - the test doesn't exercise a single line this PR adds (no `hstoreSchema`, no `isHstoreType`, no `hstoreAsJson`) and would still pass with this PR's main-code changes reverted.

Can we make it cover the actual new sink surface - the MAP value binding as jsonb (the `::jsonb` cast path, not just the DDL), plus an hstore round-trip IT ?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +789,108 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
 
 
-  // validateQuery behaviour is inherited from GenericDatabaseDialect and exercised in
-  // GenericDatabaseDialectTest; no PostgreSQL-specific override exists to test here.
+  // validateQuery is inherited from GenericDatabaseDialect; tested in GenericDatabaseDialectTest.
+
+  // ========== Complex SQL types (sql.complex.types.enable) ==========
+
+  @Test
+  public void hstoreHandlingModeShouldSelectSourceSchema() {
```
akanimesh7, line 820: Two cases missing here:
- A complex-types-**disabled** test for hstore (the arrays PR has `shouldDropArraysWhenComplexTypesDisabled`), so we know the default `false` still preserves today's drop-with-WARN behaviour.
- A NULL hstore column for both modes - json mode has an explicit `value == null` branch and map mode relies on `getObject` returning null, neither is covered.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: I  reproduced it against a real Postgres database where the driver returns a `java.util.Map` whenever `hstore` is visible on the connection's `search_path`, and a `PGObject` only when it isn't — in which case `getColumnTypeName` also reports "ext"."hstore", so `isHstoreType()` declines and the column is skipped before the value is ever read, making both failure modes you described unreachable. 

I've still added the guard, in both modes and similar to Debezium's `handleUnknownData` (nullable → null with a warning, NOT NULL → DataException, value never logged).

Also, added ITs against a real `hstore` column and other negative cases also.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -283,6 +287,11 @@ public String addFieldToSchema(
           return fieldName;
         }
 
+        if (complexTypesEnabled() && isHstoreType(columnDefn)) {
```
akanimesh7: Uber level: what's the behavior if the driver returns `ext.hstore`, i.e. hstore isn't enabled on the search_path ?

Traced it: `columnDefn.typeName()` comes back as `"ext"."hstore"`, so `isHstoreType()` returns false, we fall through to `super.addFieldToSchema`, hit the `default:` branch, and the column is dropped with the generic `JDBC type 1111 ("ext"."hstore") not currently supported` WARN. `shouldNotTreatOtherOtherTypesAsHstore` already pins that (asserts null for `"ext"."hstore"`) and `testHstoreOutsideSearchPathIsSkipped` covers it end to end — so the behavior is correct and intentional.

The gap is the diagnostic. A customer in this state loses the whole column, and the only thing in the log is "type not currently supported" — nothing hints that the extension just needs to be visible on the connection's search_path. Meanwhile the message that explains exactly that (`HSTORE_UNEXPECTED_VALUE_WARNING`) sits on a branch this case never reaches — see my other comment. Can we log the search_path hint here, where the column actually gets dropped ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -89,6 +89,10 @@ public DatabaseDialect create(AbstractConfig config) {
   static final String UUID_TYPE_NAME = "uuid";
   static final String HSTORE_TYPE_NAME = "hstore";
 
+  private static final String HSTORE_UNEXPECTED_VALUE_WARNING =
```
akanimesh7: This message tells the operator "The hstore type must be visible on the connection's `search_path` for the driver to return a map" — but that's the one cause that can't reach this branch. When hstore is off the search_path the driver reports `"ext"."hstore"`, so `isHstoreType()` declines and the column is dropped in `addFieldToSchema` long before `hstoreValue()` runs, which is what `testHstoreOutsideSearchPathIsSkipped` pins. So the actionable hint is on the wrong branch: whatever can actually fire this warning has some other, unnamed cause, and the path that really is a search_path problem logs a generic "type not supported". (The unit test's comment repeats the same claim, so the two contradict each other.)

Also the `DataException` on the NOT NULL path drops the hint entirely — just `class=X`. If either message should carry the "what do I do about it" text, it's the one that fails the task.

_🤖 flagged by Claude_

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -797,7 +811,246 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+        mock(PreparedStatement.class), 1, intValued, Collections.singletonMap("n", 1)));
+
+    Schema intKeyed = SchemaBuilder.map(Schema.INT32_SCHEMA, Schema.STRING_SCHEMA)
+        .optional().build();
+    assertThrows(ConnectException.class, () -> sinkDialect().bindField(
+        mock(PreparedStatement.class), 1, intKeyed, Collections.singletonMap(1, "v")));
+  }
+
+  @Test
+  public void shouldNotTreatOtherOtherTypesAsHstore() {
```
akanimesh7: nit: `shouldNotTreatOtherOtherTypesAsHstore` — the doubled "Other" reads like a typo. If it means "other `Types.OTHER` types", maybe `shouldNotTreatNonHstoreOtherTypesAsHstore` ?

_🤖 flagged by Claude_

akanimesh7 (commented):
Two things that don't anchor to a line.

**1. The description has drifted from the code** — worth fixing before merge, since it now documents the opposite of what ships:

| Description says | Code does |
|---|---|
| "`json`: a JSON-object STRING -> `TEXT` sink column" | `Json` logical STRING -> **JSONB** |
| "lands in `JSONB` (map) or `TEXT` (json)" | JSONB in both modes |
| "via `isHstoreType()`, `hstoreSchema()`, and `hstoreAsJson()`" | `hstoreAsJson()` no longer exists — it's `hstoreHandlingMode()` + the enum |
| "When the feature is disabled, hstore is **emitted as STRING** — unchanged behavior" | **Dropped.** `testHstoreDroppedWhenComplexTypesDisabled` asserts `assertFieldAbsent(row, "hs")` |

The Test Strategy section also still names only `hstoreHandlingModeShouldSelectSourceSchema`, with no mention of the 17 ITs, the guard tests or the disabled-path coverage — which undersells the work in here.

**2. Design doc section 2.3 needs the same correction.** Its mode table still says `json` -> `TEXT` with the escaped-string example value. The code now matches section 2.3's prose and section 2.4's summary (both JSONB), so that one table row is the last thing still asserting TEXT. Worth fixing so the doc has a single answer.

Otherwise the approach looks good to me, and all four of my earlier comments are addressed in code — thanks for the thorough turnaround on the ITs especially.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Instead of supporting, I have updated the WARN message to clearly indicate about the other schema HSTORE extensions are not supported.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: Done. Added the test cases.

akanimesh7:
Please add a ut/it for the exception path on unresolved Hstore case.

akanimesh7 (approved):

PR #1662, opened 2026-07-09
Add PostgreSQL native ARRAY complex type support

KGaneshDatta:
## Problem

PostgreSQL native array columns are **silently dropped** by the JDBC connectors today — the source emits a WARN and produces no field for them, so consumers receive no data at all for those columns and typically discover the gap only through downstream data drift. On the sink, a Connect `ARRAY` cannot be written into a native PostgreSQL array column.

## Solution

Builds on the complex-types foundation (#1651) to map PostgreSQL native arrays end to end, gated behind `sql.complex.types.enable` (default `false`, so existing pipelines are unchanged). A supported array column becomes a Connect `ARRAY` of the mapped element type on the topic, and maps back to the corresponding native `type[]` column on the sink.

- **Element types:** text/varchar/char, int2/int4/int8, float4/float8, bool, numeric, json/jsonb, hstore, and the temporals date/time/timestamp/timestamptz. Unsupported element types are skipped, preserving today's behaviour.
- **Temporal elements behave exactly like their scalar counterparts** — they honour `db.timezone`, `timestamp.granularity` (including the micro/nano precision modes) and `date.calendar.system`, on both the read and the write path. `timestamptz[]` drops the zone and is stored as a plain timestamp, matching scalar `timestamptz`, so a column and its array form round-trip with the same fidelity.
- **`numeric[]`** elements are carried as the `VariableScaleDecimal` logical type, because JDBC reports an array's element typmod as `-1` — there is no declared per-element scale, and each value may carry its own.
- **Multi-dimensional arrays** (e.g. `int[][]`) have no faithful representation in a one-dimensional Connect `ARRAY`. Rather than failing the task or dropping the column, one `null` is emitted per top-level element so the outer length is preserved — matching the Debezium-based `PostgresCdcSourceV2` connector for the same unsupported case. Dimensionality is only detectable per value, never from JDBC metadata, so the warning is logged once per column and at DEBUG thereafter; array fields are always optional so a `NOT NULL` source column still accepts the nulled value.
- **Sink:** array values are bound as native `type[]` arrays. `MAP<STRING,STRING>` elements — the shape an `hstore` column takes on the topic — are serialized to their JSON text and bound as `jsonb[]`, symmetric with the type the DDL advertises. Generic `STRUCT` and other map shapes are rejected on both the DDL and the bind path.

Scope is confined to the PostgreSQL dialect; no other dialects are affected.

- [ ] yes
- [x] no

## Test Strategy

Unit tests cover the source element-schema mapping for every supported type, value decoding (including `db.timezone`, `timestamp.granularity` and `date.calendar.system`), the null-per-element multi-dimensional handling and its warn-once-per-column logging, the unsupported-type skips, the always-optional array field, and the sink DDL and binding for each element type — including that DDL and bind agree, so auto-create never provisions a column that cannot be written.

Integration tests run against a real PostgreSQL: the primitive element matrix (with `char(n)` padding preserved), per-element `numeric` scale, `json`/`jsonb`, the temporals, the three kinds of nothing (SQL NULL column, empty array, NULL element), multi-dimensional values, unsupported-type skips, and the flag-off drop. Source→Kafka→sink round-trips cover every supported element type, both `hstore.handling.mode` values, each `timestamp.granularity`, a non-UTC `db.timezone`, both `date.calendar.system` values against a pre-1582 date, and `numeric.mapping`.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Cut from and merges into the base branch (#1651), and ships with that foundation via the standard weekly cc-connect release cycle. The feature is default-off (`sql.complex.types.enable=false`), so the change is backward compatible and safe to roll back.


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -293,6 +313,17 @@ public String addFieldToSchema(
 
         break;
       }
+      case Types.ARRAY: {
+        if (complexTypesEnabled()) {
+          Schema elementSchema = arrayElementSchemaFor(columnDefn);
+          if (elementSchema != null) {
+            // Optional even for NOT NULL: multi-dim arrays skip to null (same type name as 1-D).
```
akanimesh7: above comment is confusing to the reader

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -347,6 +385,216 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+        // Raw JSON text via the Json logical type; round-trips to a native jsonb[] sink column.
+        return Json.optionalSchema();
+      case "date":
+        return Date.builder().optional().build();
+      case "time":
+        return Time.builder().optional().build();
+      case "timestamp":
+        return Timestamp.builder().optional().build();
+      case "timestamptz":
+        // ZonedTimestamp STRING (built-in Timestamp drops the zone); round-trips to timestamptz[].
```
akanimesh7: would prefer aligning it with the non-array case

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -347,6 +385,216 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+              VariableScaleDecimal.fromLogical(elementSchema, (BigDecimal) value));
+        case Json.LOGICAL_NAME:
+          // Element is the raw JSON text; toString() covers both String and PGobject.
+          return rs -> readJdbcArray(rs, col, Object::toString);
+        case ZonedTimestamp.LOGICAL_NAME:
+          return rs -> readZonedTimestampArray(rs, col);
+        case Date.LOGICAL_NAME:
+        case Time.LOGICAL_NAME:
+        case Timestamp.LOGICAL_NAME:
+          return rs -> readTemporalArray(rs, col, elementName);
```
akanimesh7: readTemporalArray and readZonedTimestampArray are using UTC. They should rather honor the config db.timezone

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -347,6 +385,216 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+        default:
+          break;
+      }
+    }
+    // Primitive elements pass through directly: the driver already returns the matching Java type.
+    return rs -> readJdbcArray(rs, col, value -> value);
```
akanimesh7: readJdbcArray, readZonedTimestampArray and readTemporalArray have a lot of duplication which can be refactored. Please check.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -347,6 +385,216 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+      return null;
+    }
+    try {
+      Object raw = arr.getArray();
+      if (raw == null) {
+        return null;
+      }
+      Object[] elements = (Object[]) raw;
+      if (isMultiDimensional(elements)) {
+        log.warn(MULTI_DIMENSIONAL_ARRAY_WARNING, col);
```
akanimesh7: What's the motivation for this. Is it possible in postgres ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -615,69 +869,242 @@ protected boolean maybeBindPrimitive(
...
-            break;
-          default:
-            break;
-        }
+  /**
+   * Bind a Connect ARRAY value to a native PostgreSQL array parameter by dispatching on the element
+   * type. Each {@code maybeBind...Array} helper handles a single element type and returns true once
+   * it binds; returns false if the element type is not handled here.
+   */
+  private boolean bindArray(PreparedStatement statement, int index, Schema schema, Object value)
```
akanimesh7, line 1133: this maybeBind*Array cascade is very lengthy and can be reduced a lot. Make it a switch on the element type, in line with legacy maybeBindPrimitive in GenericDatabaseDialect — that drops the per-method type check (temporalArrayType folds into the switch too). And every maybeBind*Array does the same createArrayOf(...) + setArray; only building the Object[] from values differs. Extract that and pass the array builder in as a functional interface.

akanimesh7 (commented):
A couple of notes on the source-config check below — this code comes from the base PR #1651, so leaving them here rather than inline:

- `io.confluent.connect.jdbc.source.JdbcSourceConnectorConfig` is too long. Just use `JdbcSourceConnectorConfig`.
- `config.sqlTypesEnabled()` should hold this logic.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: `timestamptz[]` (→ `ZonedTimestamp` → native `TIMESTAMPTZ[]`), but a scalar `timestamptz` column still falls through to the generic `Types.TIMESTAMP` path → built-in `Timestamp` (zone dropped) → sink `TIMESTAMP`. So a `timestamptz` and a `timestamptz[]` column round-trip with different fidelity.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack. Improved it.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack.
- Replaced the cascade with a switch on element type (`arrayElementBinding`) that returns the pg-type + an `ArrayValuesBuilder` functional interface.
-  bindArray now does the single shared createArrayOf+setArray, and temporalArrayType folded into the switch.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Yes, currently the dimension of the arrays is not known while reading and building the schema. Only while processing the records it is known and as it is not supported, skipping it with the warning. Added the detailed comment for the method signature.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Done. 
Refactored the code as required. Please take a look and let me know if any further can be done.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Done. I have also added the support of using the db.timezone in timestampz[] temporal arrays.

akanimesh7: Can you add a test case/UT/IT whichever seems appropriate for this ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
akanimesh7: Exactly my point -- a `timestamptz` and a `timestamptz[]` column should not round-trip with different fidelity

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -348,6 +387,208 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+    try {
+      Object raw = arr.getArray();
+      if (raw instanceof Object[] && isMultiDimensional((Object[]) raw)) {
+        log.warn(MULTI_DIMENSIONAL_ARRAY_WARNING, col);
+        return null;
+      }
+      if (isTemporalElement(elementSchema.name())) {
+        return readTemporalElements(arr, elementSchema.name());
+      }
+      if (raw == null) {
```
akanimesh7: This `raw == null` check could move up right after `Object raw = arr.getArray()` as an early return.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -348,6 +387,208 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+
+  /**
+   * Decode a single temporal element from the array's element {@link ResultSet} (value in column 2,
+   * index in column 1). timestamptz is rendered to an ISO-8601 {@code ZonedTimestamp} string; the
+   * other temporals return their Connect logical Date/Time/Timestamp value.
+   */
+  private static Object decodeTemporalElement(
+      ResultSet elementRs, String elementName, Calendar dateCal, Calendar timeCal)
+      throws SQLException {
+    if (Date.LOGICAL_NAME.equals(elementName)) {
```
akanimesh7: Check GenericDatabaseDialect for how these are handled and follow the same preceedence.
For example, rather than directly using UTC for date, use dateTimeZoneId.
date.calendar.system is not considered here. Please include that.
see if some of the logic can be re-used by some refactoring.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -614,69 +861,205 @@ protected boolean maybeBindPrimitive(
...
+              TIME_TYPE_NAME, values -> temporalArrayFor(Time.LOGICAL_NAME, values));
+        case Timestamp.LOGICAL_NAME:
+          return new ArrayElementBinding(
+              TIMESTAMP_TYPE_NAME, values -> temporalArrayFor(Timestamp.LOGICAL_NAME, values));
+        default:
+          break;
       }
+    }
+    Schema.Type type = elementSchema.type();
+    if (type == Schema.Type.STRUCT || type == Schema.Type.MAP) {
```
akanimesh7: According to the narrative till now, MAP and STRING were mapped to JSON/JSONB. Where did STRUCT come from ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -614,69 +861,205 @@ protected boolean maybeBindPrimitive(
...
+  /**
+   * Render each temporal element (a {@code java.util.Date} carrying a logical Date/Time/Timestamp)
+   * as a UTC text literal for binding into a native date[]/time[]/timestamp[] column. Formatting in
+   * UTC keeps the value timezone-safe irrespective of the JVM default zone. Nulls are preserved.
+   */
+  private static Object[] temporalArrayFor(String elementName, Collection<?> valueCollection) {
+    DateTimeFormatter fmt = temporalFormatter(elementName);
+    return valueCollection.stream()
+        .map(o -> o == null ? null
+            : fmt.format(((java.util.Date) o).toInstant().atZone(ZoneOffset.UTC)))
```
akanimesh7: zone fixed to UTC. This is wrong. Use member vars zoneId and dateTimeZoneId

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -614,69 +861,205 @@ protected boolean maybeBindPrimitive(
...
+        .toArray(String[]::new);
+  }
+
+  /**
+   * Return the UTC text format for a temporal logical element name (date, time, or timestamp).
+   */
+  private static DateTimeFormatter temporalFormatter(String elementName) {
+    if (Date.LOGICAL_NAME.equals(elementName)) {
+      return DateTimeFormatter.ofPattern("yyyy-MM-dd");
```
akanimesh7: these patterns are duplicated in date time utils. Re-use them.

akanimesh7 (commented):
Some pointers on the ARRAY solution
- Please implement and account for `date.calendar.system`
- I'm still not convinced on the disparity between handling of timestamptz and a timestamptz[].
- Sink implementation for temportal types still doesnt account for db.timezone and date.calendar.system.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: STRUCT never appears in the source narrative. On the source side these Postgres types map to Json STRING (json/jsonb), MAP<STRING,STRING> (hstore), and ARRAY<element> (native arrays)

 The sink consumes arbitrary upstream topics with STRUCT shape, where a nested record naturally arrives as a Connect STRUCT (e.g. an Avro/Protobuf record, or an array of objects). `getSqlType` already maps JSONB for STRUCT — that mapping comes from the base PR (case STRUCT: case MAP: → JSONB), and because the ARRAY branch recurses into its element type, an ARRAY<STRUCT> already resolved to JSONB[].

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -348,13 +384,226 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+    return base.toLowerCase();
+  }
+
+  /**
+   * Select the value converter for a PostgreSQL array column. Returns null when the feature is
+   * disabled or the element type is not supported; otherwise {@link #readArray} handles element
+   * decoding based on the schema resolved by {@link #arrayElementSchemaFor}.
+   */
+  private ColumnConverter arrayColumnConverter(ColumnDefinition columnDefn, int col) {
+    if (!complexTypesEnabled()) {
```
akanimesh7: this check should not be a part of this method. It should be done before calling this method.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -348,13 +384,226 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+  }
+
+  /**
+   * Whether a PostgreSQL array element type is temporal, and therefore decoded through the array's
+   * element {@link ResultSet} so an explicit calendar can be applied. Routing on the PostgreSQL
+   * type (rather than the Connect schema name) keeps working under the precision modes where the
+   * timestamp element schema is an unnamed INT64/STRING.
+   */
+  private static boolean isTemporalElementType(String elementType) {
+    return DATE_TYPE_NAME.equals(elementType)
```
akanimesh7, line 564: No need for static here ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -348,13 +384,226 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+    }
+  }
+
+  /**
+   * Map elements the driver already returns as the target Java type into a Connect list (nulls
+   * preserved). Used for primitive, Json and VariableScaleDecimal elements.
+   */
+  private static List<Object> readMappedElements(Object[] elements, Schema elementSchema) {
+    List<Object> out = new ArrayList<>(elements.length);
+    for (Object element : elements) {
```
akanimesh7, line 619: elementSchema is not required to be passed in this method. Its required only in mapArrayElement, where it can be obtained using VariableScaleDecimal.optionalSchema()

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -614,69 +866,207 @@ protected boolean maybeBindPrimitive(
...
+   * dedicated PostgreSQL array type (json, timestamptz, numeric, the temporals, and STRUCT/MAP as
+   * jsonb) are resolved by {@link #arrayElementBinding(Schema)} and bound via
+   * {@code createArrayOf}; any other (primitive) element type falls back to
+   * {@link #maybeBindPrimitiveArray}. Returns false if the element type is not handled here.
+   */
+  private boolean bindArray(PreparedStatement statement, int index, Schema schema, Object value)
+      throws SQLException {
+    Collection<?> values = arrayValueCollection(value);
+    Schema elementSchema = schema.valueSchema();
+    ArrayElementBinding binding = arrayElementBinding(elementSchema);
```
akanimesh7: This doesn't check `complexTypesEnabled()` at all, unlike `addFieldToSchema`'s and `arrayColumnConverter`'s ARRAY cases (and `getSqlType`'s STRUCT/MAP case). With the flag off (default), an incoming `ARRAY<Struct>`/`ARRAY<Map>` - from any producer, not just this connector - would still get bound as `jsonb[]` here. Doesn't that contradict "when the feature is not enabled, every type behaves exactly as it does today" ?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +800,481 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+    assertArrayElement("_date", Type.INT32, Date.LOGICAL_NAME);
+    assertArrayElement("_time", Type.INT32, Time.LOGICAL_NAME);
+    assertArrayElement("_timestamp", Type.INT64, Timestamp.LOGICAL_NAME);
+    // timestamptz drops the zone and maps to the same timestamp schema as scalar timestamptz.
+    assertArrayElement("_timestamptz", Type.INT64, Timestamp.LOGICAL_NAME);
+  }
+
+  @Test
+  public void timestampArraysHonorPrecisionMode() {
+    // The element schema follows timestamp.granularity, exactly like scalar timestamp/timestamptz:
```
akanimesh7: test name should be granurality, precision mode may not be correct. Also, this test would be better if we took STRING as the value. INT64 is the schema type by default too.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +800,481 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+        Arrays.asList("{\"k\":\"v\"}", "{\"a\":1}"),
+        "jsonb",
+        new Object[]{"{\"k\":\"v\"}", "{\"a\":1}"});
+  }
+
+  @Test
+  public void shouldBindStructArrayAsNativeJsonbArray() throws Exception {
+    // ARRAY<STRUCT> -> jsonb[]; each element is serialized to its JSON text (schema field order).
+    Schema element = SchemaBuilder.struct().optional()
+        .field("a", Schema.INT32_SCHEMA)
```
akanimesh7: Which case is `ARRAY<STRUCT>` solving for, now that `json.handling.mode` is removed ? See the top-level comment on this review re: getSqlType's STRUCT/MAP case - same question.

akanimesh7 (commented):
Two things anchor in `getSqlType`'s `ARRAY`/`STRUCT`/`MAP` cases, which are base-branch (#1651) code so I can't comment inline here, but they're directly relevant to the array work in this PR:

- `getSqlType`'s `ARRAY` case recurses through `field.name()` unchanged - if an `ARRAY<INT64>`/`ARRAY<STRING>` field's name is in `timestamp.fields.list`, this emits `TIMESTAMP[]` DDL, but `primitiveArrayFor` (added in this PR) has no `timestamp.fields.list` awareness and sends a plain `Long[]`/`String[]`. Is this combination handled anywhere?
- Which case is the `STRUCT`/`MAP` -> `JSONB` branch actually solving for? With `json.handling.mode` removed (#1660), our source will never emit a STRUCT/MAP for a json/jsonb column anymore. Is this still needed for STRUCT/MAP values from a different producer/topic, or is it dead now? (Same question applies to the `ARRAY<STRUCT>`/`ARRAY<MAP>` bind tests in this PR - flagged inline below.)

Separately: how does Debezium handle multi-dimensional arrays? Per our own docs for the sibling `PostgresCdcSourceV2` (Debezium-based) connector, it doesn't drop the whole column - it preserves the outer array's shape/length and nulls each element instead (e.g. a `bigint[][]` column with value `{{1001,1002},{1003,1004}}` is published as `{"array": [null, null]}`). This PR instead nulls the whole field when it hits a multi-dimensional array. Not saying one is wrong, but wanted to flag the fidelity difference between the two connectors for the same unsupported case - was this considered?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +800,481 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+        new Object[]{"{\"a\":1,\"b\":\"x\"}", "{\"a\":2,\"b\":\"y\"}"});
+  }
+
+  @Test
+  public void shouldBindMapArrayAsNativeJsonbArray() throws Exception {
+    // ARRAY<MAP> -> jsonb[]; each element is serialized to its JSON object text.
+    Schema element = SchemaBuilder.map(Schema.STRING_SCHEMA, Schema.STRING_SCHEMA)
+        .optional().build();
+    verifyArrayBind(
+        element,
```
akanimesh7: Same question for `ARRAY<MAP>` - which case is this solving for now that `json.handling.mode` is removed ?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +800,481 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+
+  @Test
+  public void arrayConverterRoutesByElementLogicalType() throws Exception {
+    // numeric[] -> each element decoded into a VariableScaleDecimal struct (per-value scale).
+    Object num = ((List<?>) arrayColumnConverter("_numeric")
+        .convert(arrayResultSet(new Object[]{new BigDecimal("1.5")}))).get(0);
+    assertEquals(0, new BigDecimal("1.5").compareTo(
+        VariableScaleDecimal.toLogical((Struct) num)));
+
+    // jsonb[] -> each element carried as its raw JSON text.
```
akanimesh7, line 1810: None of the read-side or bind-side array tests cover a null element mixed among non-null values, even though `readMappedElements`/`mapArrayElement`, `bigDecimalArrayFor`, `jsonbArrayFor`, and `temporalArrayFor` all explicitly branch on `o == null`. Can we add a case with a null element for at least numeric/json/temporal ?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -781,7 +800,481 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+
+  @Test
+  public void shouldReturnNullWhenArrayHasNoContents() throws Exception {
+    // getArray() returning null short circuits before any element decoding.
+    ResultSet resultSet = mock(ResultSet.class);
+    Array array = mock(Array.class);
+    when(resultSet.getArray(1)).thenReturn(array);
+    when(array.getArray()).thenReturn(null);
+
+    assertNull(arrayColumnConverter("_int4").convert(resultSet));
```
akanimesh7, line 1902: This only covers `getArray()` returning null (SQL NULL array). Is there coverage for a genuinely empty, non-null array ('{}', 0 elements) ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack. Moved it to before method.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: I kept it because, most of other methods in the class are already having it.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack. Removed it and updated all the referenced methods. Please take a look.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Yes. You are right. Missed this. 

I have added the check for gating here.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: Ack. Renamed to `timestampArraysHonorTimestampGranularity` and switched the assertion to
`micros_string`

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: ARRAY<STRUCT> is something which we might get it from other producers and now these complex types are supported on Sink connector which is independently apart from JDBC Source, so we should support IMO. Also now gated behind `sql.complex.types.enable`

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: MAP values from other producers, and hstore `map` mode on #1661 which puts a `MAP<STRING,STRING>` on the topic.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: You're Right. 
Added `arrayReadPreservesNullElements` and
`arrayBindPreservesNullElements`, each exercising a null among non-nulls for numeric, jsonb and
temporal on both paths.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: Added `shouldReturnEmptyListWhenArrayHasNoElements` — `'{}'` decodes to an empty list, as distinct
from a SQL NULL array which stays null.

KGaneshDatta:
Thanks for the review @akanimesh7 

> getSqlType's ARRAY case recurses through field.name() unchanged - if an ARRAY<INT64>/ARRAY<STRING> field's name is in timestamp.fields.list, this emits TIMESTAMP[] DDL, but primitiveArrayFor (added in this PR) has no timestamp.fields.list awareness and sends a plain Long[]/String[]. Is this combination handled anywhere?

IIUC, here you mean that when the table column is in timestamp.fields.list then it emits TIMESTAMP[] DDL but later as it is ARRAY<INT64> it sends as BIGINT[]. 
This behavior already exists in the base 10.9.x branch and hence I kept the same behavior. If such case happens, then the customer must pre-create the table. 
Also, now we support the TIMESTAMP[] types, this should not be an issue with this case.

> Which case is the STRUCT/MAP -> JSONB branch actually solving for? With json.handling.mode removed (https://github.com/confluentinc/kafka-connect-jdbc/pull/1660), our source will never emit a STRUCT/MAP for a json/jsonb column anymore. Is this still needed for STRUCT/MAP values from a different producer/topic, or is it dead now? (Same question applies to the ARRAY<STRUCT>/ARRAY<MAP> bind tests in this PR - flagged inline below.)

STRUCT/MAP -> JSONB happens on the case in HSTORE i.e. MAP -> JSONB, but remove the support for STRUCT types.

> how does Debezium handle multi-dimensional arrays?

We wanted to keep this out of scope as aligned with the Product.

akanimesh7 (commented):
All 10 of my previous comments are addressed in the code — thanks, and the ITs are genuinely thorough now (25 array ITs covering the null matrix, granularity modes, non-UTC `db.timezone`, and proleptic vs legacy calendar round-trips). The Debezium note from my last review was acted on too: multi-dimensional arrays now preserve the outer length with one null per element instead of dropping the field. One consequence of that change is worth a look — see my comment on the log level.

Two housekeeping items:

**1. The description has drifted from the code.**

| Description says | Code does |
|---|---|
| "Multi-dimensional arrays … are **skipped with a warning**" | Neither — they emit one null per element, logged at **debug** |
| "**Sink:** `STRUCT`/`MAP` elements are serialized to their JSON text and bound as `jsonb[]`" | Both dropped in `be33de7`; only the hstore-shaped `MAP<STRING,STRING>` still maps to `jsonb[]` |
| Element types list: text/int/float/bool/numeric/json/temporals | Omits `hstore`, added in `042c915` |

**2. Two comment threads are now misleading.** Your replies on `ARRAY<STRUCT>` and `ARRAY<MAP>` argue for keeping them ("we should support IMO"), but `be33de7` dropped both and the tests now assert the opposite (`shouldNotBindStructArray`, `shouldNotMapStructArrayToSqlType`). Can you resolve those two threads so anyone reading back gets the current story ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -350,6 +389,256 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+        return Schema.OPTIONAL_BOOLEAN_SCHEMA;
+      case "numeric":
+      case "decimal":
+        // Per-value scale carried via VariableScaleDecimal (array typmod is -1 in JDBC metadata).
+        return VariableScaleDecimal.optionalSchema();
+      case "json":
+      case "jsonb":
+        // Raw JSON text via the Json logical type; round-trips to a native jsonb[] sink column.
+        return Json.optionalSchema();
+      case HSTORE_TYPE_NAME:
```
akanimesh7, line 489: Is an array of hstore something we need to support ? I don't see `hstore[]` in the design doc's array element table — it's in neither the supported list (text/int/float/bool/numeric/json/temporals) nor the unsupported one (uuid[], inet[], money[], interval[], varbit[], xml[], geometric, multi-dimensional). So this looks like scope we picked up along the way rather than something that was asked for.

And if we do want it, why is the code here rather than in the hstore PR (#1661) ? Two concrete problems with the current split:

1. **This branch has no scalar hstore support at all** — `isHstoreType` doesn't exist here, #1661 is what adds it. So if #1662 merges before #1661, `hstore[]` columns get read while plain `hstore` columns are still silently dropped. That's a strange intermediate state to ship.
2. **The mode -> schema decision is now duplicated.** #1661 has `hstoreSchema()` doing exactly this Json-vs-`MAP<STRING,STRING>` switch; here it's re-inlined for the element case, and the two already differ — that one derives optionality from `columnDefn.isOptional()`, this one is always optional. When both land we'll have two copies that can drift.

If it stays, can the element case call a shared helper ? `hstoreHandlingMode()` and `HSTORE_TYPE_NAME` are already in the base, so a single `hstoreSchema(boolean optional)` could live there and serve both.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -350,6 +389,256 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+    if (arr == null) {
+      return null;
+    }
+    try {
+      Object raw = arr.getArray();
+      if (raw == null) {
+        return null;
+      }
+      if (raw instanceof Object[] && isMultiDimensional((Object[]) raw)) {
+        log.debug(MULTI_DIMENSIONAL_ARRAY_MESSAGE, columnId);
```
akanimesh7: This used to be a `log.warn`. Now that we emit one null per element rather than dropping the column, the customer silently gets nulls where their data was — and at DEBUG nobody sees it under default logging. That's data loss with no signal.

I'd guess the downgrade was to avoid flooding, since this fires per row per poll. But the fix for flooding shouldn't be to hide it entirely — can we warn once per column instead (a `Set<ColumnId>` of already-warned columns) ? Warning at schema-build time isn't an option since multi-dimensionality is only detectable from values, as the `isMultiDimensional` javadoc says, so one-shot per column is the closest we get.

_🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -350,6 +389,256 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
...
+      return out;
+    }
+  }
+
+  /**
+   * Decode one hstore array element. The element {@link ResultSet} applies the driver's own hstore
+   * decoding, so the value arrives as a Map; json mode then serializes it.
+   */
+  private Object decodeHstoreElement(ResultSet elementRs) throws SQLException {
+    Object value = elementRs.getObject(2);
```
akanimesh7: Where does the magical number 2 come from ? It's the JDBC contract for `Array.getResultSet()` — column 1 is the index, column 2 the value — and the caller spells that out at line 576, but this method repeats the bare `2` with no comment. There are 4 of these element-column reads across `decodeTemporalElement` and `decodeHstoreElement`. Worth a named constant (`ELEMENT_VALUE_COLUMN = 2`) so the meaning travels with every use ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -408,26 +696,38 @@ private boolean complexTypesEnabled() {
...
+    switch (schemaName) {
+      case Decimal.LOGICAL_NAME:
+        return "DECIMAL";
+      case Date.LOGICAL_NAME:
+        return "DATE";
+      case Time.LOGICAL_NAME:
+        return "TIME";
+      case Timestamp.LOGICAL_NAME:
+        return "TIMESTAMP";
+      case Json.LOGICAL_NAME:
```
akanimesh7, line 874: Check if returning null here is backward compatible ?

Traced it and I believe yes, but the mechanism is worth stating because it's easy to get wrong. The pre-flag pipeline — a Postgres json column emitted as a plain STRING and written to a `TEXT` column — is untouched, because on this branch `addFieldToSchema` emits an **untagged** `STRING` for json/jsonb (the `Json` tagging is #1660's change, not this branch's). An untagged STRING has no `schemaName`, so `logicalSqlType` returns null on the very first check and it lands in `case STRING:` -> `TEXT`, exactly as before. It never reaches this branch.

So this `null` only governs a mixed config: a `Json`-tagged value — from #1660's source with its flag on, or a `json[]`/`hstore[]` element — arriving at a sink with the flag off, where it degrades to `TEXT` / `TEXT[]`. `gatedLogicalTypesFallBackWhenComplexTypesDisabled` pins that (TEXT, TEXT[], and the `VariableScaleDecimal` throw). Can you confirm the degrade-to-TEXT behavior for that mixed case is intended and not just a side effect ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -638,69 +938,207 @@ protected boolean maybeBindPrimitive(
...
+
+    Object[] toArray(Collection<?> values) {
+      return valuesBuilder.build(values);
+    }
+  }
+
+  /**
+   * Resolve the native array binding for a Connect element schema, or null if the element has no
+   * dedicated PostgreSQL array type (a primitive, handled by {@link #maybeBindPrimitiveArray}).
+   * Dispatch is by logical type name first, then STRUCT/MAP -> jsonb kept last, since some logical
```
akanimesh7: This javadoc is stale two ways now: there's no `STRUCT/MAP -> jsonb` branch left (dropped in `be33de7`), and the string-to-string map check is the **first** thing in the method, not last. The stated reason for "kept last" — that VariableScaleDecimal is itself a STRUCT — no longer applies either, since `isStringToStringMap` only matches MAP. It reads as if the ordering is load-bearing for a case that's gone.

_🤖 flagged by Claude_

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -797,7 +820,832 @@ private <T> void verifyArrayFormatting(PostgreSqlDatabaseDialect dialect, Expres
...
+    List<?> value = Collections.singletonList(new Struct(element).put("a", 1));
+
+    assertThrows(ConnectException.class, () -> dialect.bindField(mock(PreparedStatement.class), 1,
+        schema, value, mock(ColumnDefinition.class), "field"));
+    assertThrows(ConnectException.class, () -> complexTypesDialect().bindField(
+        mock(PreparedStatement.class), 1, schema, value, mock(ColumnDefinition.class), "field"));
+  }
+
+  @Test
+  public void shouldNotMapStructArrayToSqlType() {
```
akanimesh7, line 1481: Good that this pins DDL/bind agreement for `ARRAY<STRUCT>`. Two cases still have exactly the disagreement it guards against, both because `getSqlType`'s `case ARRAY:` recurses on the element type while the bind path has no matching branch:

- **Nested arrays.** `ARRAY<ARRAY<INT32>>` recurses to `INT[][]`, which Postgres accepts — so auto-create succeeds, then every write fails: `arrayElementBinding` returns null, `primitiveArrayFor` has no `ARRAY` case, `bindArray` returns false, and `super.maybeBindPrimitive` falls through to `ConnectException: Unsupported source data type: ARRAY`. (That also answers my earlier question on `bindArray` that didn't get a reply — it's deterministic, not an NPE, so no correctness bug; just a column that can never be written.)
- **`timestamp.fields.list`.** An `ARRAY<INT64>`/`ARRAY<STRING>` field whose name is in that list recurses through the `INT64`/`STRING` cases and comes out `TIMESTAMP[]`, but `primitiveArrayFor` binds a plain `Long[]`/`String[]`. I raised this in the body of my last review and it wasn't picked up.

Can we extend this test's principle to both — either reject them in `getSqlType` like STRUCT, or handle them in the bind path ?

Separately, that `ConnectException` message is weak for a sink failure: it says *source* data type, it reports only `ARRAY` so a nested array is indistinguishable from `ARRAY<uuid>` or any other unsupported element, and it drops the field name though `fieldName` is in scope. It lives in `GenericDatabaseDialect` so it's base-branch, but worth improving while we're here.

_🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Support for HSTORE[] array types is not requested explicitly. But this has been supported in the Debezium code base. Hence, only support for ARRAYS is requested, hence added this HSTORE[] support also. 

Added the `hstoreSchema(boolean optional)` in the base.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack. 
Added the constant - `ELEMENT_VALUE_COLUMN = 2`

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: I have added the `Set<ColumnId>` based approach in one of the commit but this would not viable when the worker restarted due to rebalances, hence moved it back to DEBUG log. But as you mentioned it would not be visible when there are NULL, hence added this `Set<ColumnId>` based approach only.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Updated the doc.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Confirmed intended. The Json gate in base, added because `testSourceEnabledSinkDisabledLandsInTextWithoutDataLoss` on [#1660](https://github.com/confluentinc/kafka-connect-jdbc/pull/1660) failed with `expected:<text>` but `was:<jsonb>` against the ungated mapping — the test was left unchanged and base was gated to satisfy it. Promoting a tagged value to jsonb on a flag-off sink was the rejected alternative; this PR adds the matching gate for VariableScaleDecimal beside it.

Also right, that `json[]`/`hstore[]` elements are the route by which it's reachable here, and the degrade differs by the element's underlying type: Json is STRING-based so it falls to the primitive bind path and lands in text[] with documents intact, while VariableScaleDecimal is a STRUCT and is rejected. Now pinned by `gatedArrayElementsDegradeByUnderlyingTypeWhenComplexTypesDisabled` and `testJsonArrayRoundTripSourceEnabledSinkDisabled`, the end-to-end equivalent of #1660's scalar test.

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: Worked on the first option — `getSqlType` now rejects both, plus `ARRAY<BYTES>` which has the same disagreement, through a single `isUnbindableArrayElement` guard that routes to `super.getSqlType` exactly as `ARRAY<STRUCT>` does; on `timestamp.fields.list` I rejected rather than emitting `BIGINT[]` to match the bind, since quietly storing raw epoch values for someone who explicitly asked for timestamp semantics would be worse than failing. `shouldNotMapUnbindableArrayElementsToSqlType` and `shouldNotMapTimestampFieldsListArrayToSqlType` cover all three (including that a Decimal element still maps, since it is BYTES-based but does bind), and I'll take the ConnectException message on #1651 where it is written.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -350,6 +401,256 @@ protected boolean isJsonType(ColumnDefinition columnDefn) {
     return JSON_TYPE_NAME.equalsIgnoreCase(typeName) || JSONB_TYPE_NAME.equalsIgnoreCase(typeName);
   }
 
+  /**
+   * Map a PostgreSQL array column's element JDBC type to a Connect element {@link Schema}.
+   * Returns null when the element type isn't covered by this MVP (callers fall through to the
+   * generic dialect, which will skip the column with a WARN).
+   */
+  private Schema arrayElementSchemaFor(ColumnDefinition columnDefn) {
```
akanimesh7, line 457: `bytea[]` is missing here, and I don't think it was a deliberate exclusion.

Scalar `bytea` has always been supported — the generic dialect maps `BINARY`/`VARBINARY`/`BLOB`/`LONGVARBINARY` to `BYTES`. But `bytea` isn't in this switch, so a `bytea[]` column is silently dropped. And it appears in **neither** list in the design doc: not in the supported array element types (text/varchar/char, int2/int4/int8, float4/float8, bool, numeric, json/jsonb, temporals), and not in the unsupported one either (`uuid[]`, `inet[]`, `cidr[]`, `macaddr[]`, `macaddr8[]`, `money[]`, `interval[]`, `varbit[]`, `xml[]`, geometric, multi-dimensional). It looks like it was just never considered — pgjdbc does return `byte[][]` for it, so it is readable.

What makes me fairly confident it's an oversight rather than a decision: `isUnbindableArrayElement` now explicitly rejects `ARRAY<BYTES>` on the sink DDL side. So the sink has a named opinion about a shape the source can never produce. The two halves were reasoned about separately.

So either:
- support it — `case "bytea": return Schema.OPTIONAL_BYTES_SCHEMA;` and drop it from the sink's unbindable set, or
- decide it's out of scope and add `bytea[]` (and plain `bit[]`, which is in the same position) to the design doc's unsupported list, so the omission is a documented choice rather than a gap.

Same question for `bit[]`: the Postgres dialect handles a scalar `bit` specially (`numBits <= 1` -> BOOLEAN, `<= 8` -> INT8, else BYTES), `varbit[]` is documented as unsupported, but plain `bit[]` is listed nowhere.

_🤖 flagged by Claude_

akanimesh7 (approved):

PR #1666, opened 2026-07-17
Add shadow logging for query + incrementing modes with extra clauses in query config

KGaneshDatta:
> **This PR is shadow logging only. It does not reject any configuration and does not change
> connector behaviour.** The goal is to measure, from CCloud logs, how many real configs would be
> affected before an enforcing validation is proposed in a follow-up.

## Problem
In an incremental query mode (`incrementing`, `timestamp`, or `timestamp+incrementing`) with a custom `query`, the connector builds its statement by concatenating its own criteria onto the **end** of the user's query:

- `incrementing` → ` WHERE <inc> > ? ORDER BY <inc> ASC`
- `timestamp` → ` WHERE <ts> > ? AND <ts> < ? ORDER BY <ts> ASC`
- `timestamp+incrementing` → ` WHERE <ts> < ? AND ((<ts> = ? AND <inc> > ?) OR <ts> > ?) ORDER BY <ts>,<inc> ASC`

This only produces valid SQL when the query's **outermost** `SELECT` is "bare". If the outermost `SELECT` already contains a top-level `WHERE`, `ORDER BY`, `GROUP BY`, `HAVING`, `LIMIT`/`OFFSET`/`FETCH`, or is a set operation (`UNION`/`INTERSECT`/`EXCEPT`), the appended `WHERE … ORDER BY …` yields invalid SQL (e.g. `… ORDER BY x WHERE … ORDER BY …`). The connector accepts the config at create time and only fails **at runtime on every poll** with a driver `SQLException` — on Confluent Cloud this surfaces after creation as a `FAILED` task, so the misconfiguration isn't caught up front.

The documented, supported way to filter in these modes is to wrap the query in a sub-select so the outermost `SELECT` stays bare (see [Custom WHERE clauses with query modes](https://docs.confluent.io/kafka-connectors/jdbc/current/source-connector/overview.html#specifying-a-where-clause-with-query-modes)), e.g. `SELECT * FROM (SELECT … WHERE … ORDER BY …) sub`. Today nothing steers users toward that pattern before the failure.

## Solution
Before proposing an enforcing validation, this PR adds a **shadow (log-only) detector** so the
fleet-wide blast radius can be measured from real CCloud configs.

`JdbcSourceConnectorValidation.logQueryAppendedCriteriaCompatibility()` emits a single `INFO` line
tagged `[query-appended-criteria-shadow]` when a custom query is configured **and** the mode is
incremental **and** the query's outermost `SELECT` carries a clause that would collide with the
appended criteria. `bulk` is exempt (it never appends a criteria). It **does not** add a config
error, so nothing is rejected.

Safety properties, so the shadow cannot affect any connector:
- it is invoked **outside** the `&&` chain in `validate()`, so it can neither be short-circuited away nor influence the returned `Config`;
- the whole body is wrapped in `try/catch (Exception)`, so it can never throw;
- **the query itself is never logged** (it may contain customer data) — only the config key in use (`query` vs `query.masked`) and the mode.

The check is backed by a new helper, `SqlParser.outerSelectHasTailClauses(sql)` (reusing the existing jsqlparser dependency), which:
- inspects **only the outermost `SELECT`** — a top-level `WHERE`, `ORDER BY`, `GROUP BY`, `HAVING`, `LIMIT`/`OFFSET`/`FETCH`, or a set operation is flagged;
- **allows** clauses nested inside a `FROM` sub-select — i.e. the documented `SELECT * FROM (<your query>) sub` pattern is not flagged;
- **fails open** (returns `false`) on empty, unparseable, or non-`SELECT` input, so exotic/dialect-specific SQL is never wrongly flagged.

### How to read the resulting data
Two limitations matter when interpreting the logs, both of which cause **under**-counting:

1. **Only inbound configs are covered.** `Connector.validate()` is called on connector create/update and by the UI validate path — *not* on connector start (`JdbcSourceConnector.start()` builds its config directly). Connectors already running with a colliding query will not emit the line until someone edits their config. **Absence of the log line does not mean the fleet is clean.**
2. **Known fail-open blind spots**, all of which are real runtime breakages the detector does not flag: a trailing semicolon on an otherwise bare query (`SELECT * FROM t;`), a bare parenthesized select (`(SELECT * FROM t)`), and `FOR UPDATE`. These should be closed before any enforcing version ships.

There is no backward-compatibility risk: nothing is rejected, and the only added work is one jsqlparser
parse on the validate path.

- [ ] yes
- [x] no

## Test Strategy
- `SqlParserTest` — `outerSelectHasTailClauses`: plain select / joins / the documented sub-select pattern (not flagged); top-level `WHERE`, `ORDER BY`, `GROUP BY`+`HAVING`, `LIMIT`/`OFFSET`, set operation, and a sub-select whose *outer* select still carries `WHERE`+`ORDER BY` (flagged); null / blank / unparseable / non-`SELECT` (fail-open).
- `JdbcSourceConnectorValidationTest` — asserts both halves of the shadow contract, i.e. that the detector fires **and** that no config error is added: `incrementing` + top-level `ORDER BY`, `query.masked` + `timestamp` + top-level `WHERE`, and `timestamp+incrementing` + `UNION` are all flagged with no errors; the sub-select pattern, `bulk` + `ORDER BY`, and table-mode (no custom query) are not flagged, also with no errors.
- Manual: the parser was exercised against dialect-specific shapes to check for false positives — `SELECT TOP 10`, `WITH (NOLOCK)`, optimizer hints, quoted identifiers, CTEs with clauses inside the CTE, and `SELECT * FROM (a UNION b) sub` are all correctly left unflagged.
- Full unit suite green (1441 tests, 0 failures) and checkstyle clean.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/util/SqlParser.java
+++ b/src/main/java/io/confluent/connect/jdbc/util/SqlParser.java
@@ -59,6 +59,10 @@ import net.sf.jsqlparser.parser.CCJSqlParserUtil;
 import net.sf.jsqlparser.schema.Column;
 import net.sf.jsqlparser.statement.Statement;
 import net.sf.jsqlparser.statement.select.OrderByElement;
+import net.sf.jsqlparser.statement.select.ParenthesedSelect;
+import net.sf.jsqlparser.statement.select.PlainSelect;
+import net.sf.jsqlparser.statement.select.Select;
+import net.sf.jsqlparser.statement.select.SetOperationList;
 import net.sf.jsqlparser.util.deparser.ExpressionDeParser;
 import net.sf.jsqlparser.util.deparser.SelectDeParser;
 import net.sf.jsqlparser.util.deparser.StatementDeParser;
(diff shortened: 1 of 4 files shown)


Kalla-Shubham (approved):
lgtm

PR #1676, opened 2026-08-10
Write PostgreSQL hstore columns from Connect maps, including off-search_path extensions

KGaneshDatta:
## Problem

**1. An hstore column could not be written back as an hstore column.**

#1661 taught the source to read `hstore` as a Connect `MAP<STRING,STRING>` (map mode) or a logical JSON string (json mode). The sink, however, mapped `MAP<STRING,STRING>` to `jsonb`, so a Postgres-to-Postgres pipeline turned an `hstore` column into a `jsonb` one. The values survived, but the target type did not, and none of the hstore operators worked at the destination.

Debezium does not have this gap: its PostgreSQL sink dialect registers `MapToHstoreType`, which maps the Connect MAP type to `hstore` with `cast(? as hstore)`. The decision is made by the dialect, with no config and no metadata on the record — the MySQL dialect registers `MapToJsonType` for the same Connect type.

**2. An hstore installed outside the connection's `search_path` was refused outright.**

`CREATE EXTENSION hstore SCHEMA ext` is a common way to segregate extensions. When the extension is not on the `search_path`, pgjdbc cannot resolve its type OID, and two things follow: `getColumnTypeName` returns the qualified `"ext"."hstore"`, and values arrive as a `PGobject` carrying raw `"k"=>"v"` text rather than a decoded `Map`.

The source recognised that state and **failed the whole task**, with a message telling the operator to change their `search_path` or set `hstore.handling.mode=none`. The sink had no way to name the type at all.

Debezium handles this transparently: `TypeRegistry.get(String)` strips the schema and quotes before its catalog lookup, and `convertHstoreToMap` parses the text form with the driver's own `HStoreConverter.fromString`. There is no reason the extension's location should be the operator's problem.

**3. Selecting hstore with no extension installed had no clear failure**, and a `MAP` bound into a column that is not an hstore failed with a raw PostgreSQL parse error that never mentioned hstore.

## Solution

**1. A Connect `MAP<STRING,STRING>` now maps to `hstore`**, matching `MapToHstoreType`. Values are serialized to hstore text and bound as text for a `::hstore` cast to parse server-side, as Debezium's `cast(? as hstore)` does.

Two guards are deliberately kept rather than copying Debezium exactly:

- The mapping stays behind `sql.complex.types.enable`. Debezium has no equivalent gate only because it had no prior MAP behaviour to preserve; we do, so upgrading alone must not change an existing pipeline.
- Only `MAP<STRING,STRING>` is accepted. `MapToHstoreType` casts to `Map<String,String>` unconditionally and would `ClassCastException` on, say, `MAP<STRING,INT32>`; the shape check turns that into the ordinary "no mapping to the SQL database column type" failure at DDL time.

json mode is unchanged and still lands in `jsonb`. A `Json` string is a JSON document that happens to have come from hstore, and that is how Debezium routes it too. So `hstore[]` round trips to `hstore[]` in map mode and to `jsonb[]` in json mode.

**2. An hstore is now recognised wherever the extension is installed, on both sides.**

- *Source*: `isHstoreType` compares the local type name, so `"ext"."hstore"` is the same type as `hstore` — the same normalisation Debezium applies before its catalog lookup. Values that arrive as raw text are parsed into the map both handling modes require, and `hstore[]` elements are normalised identically. The refusal path is gone: `isUnresolvedHstoreType`, `isUnresolvedHstoreArrayType`, `hstoreOffSearchPathError` and its message are deleted.
- *Sink*: the extension's location is resolved once per dialect instance — `to_regtype('hstore')` first, then `pg_extension` joined to `pg_namespace` — and cached alongside the maximum identifier length, which already resolves that way. The answer is held as a single `HstoreType` value (`UNRESOLVED`, `ABSENT`, or installed under a name) rather than a name plus a flag, so the three states are named and one volatile reference publishes them together. DDL and the array element type use the resulting name, so an extension in `ext` produces `"ext".hstore`, and the cast follows the type name the driver reports for the column, giving `?::"ext"."hstore"` for a pre-existing one — requoted rather than interpolated, since pgjdbc builds that qualified name without escaping an embedded quote.

**3. Both failure modes now name their cause.** With no extension installed anywhere, writing a map fails with *"the hstore extension is not installed on this database. Install it with CREATE EXTENSION hstore, in any schema."* Writing a map into a column that is not an hstore fails with *"column type is jsonb, not hstore. Recreate the column as hstore, or set sql.complex.types.enable=false"* rather than PostgreSQL's `invalid input syntax for type json`. The remediation names the type the operator would actually create, so the array form reads `column type is jsonb[], not hstore[]. Recreate the column as hstore[]` — pgjdbc reports those as `_jsonb`/`_hstore`, which is not what anyone writes in DDL. This adds no capability — `MAP` still maps to `hstore` only — it just makes the required failure legible.

Both are raised as `ConnectException`, which fails the task. Neither is a DDL failure — one throw site is `bindField` — and both conditions are properties of the database or the target table rather than of a record, so no batch split can find a record that fits and reporting each one in turn only drains a stream into a DLQ while the task reports healthy. This matches `GenericDatabaseDialect`, which raises `ConnectException` for any value it cannot bind.

A failed catalog read is deliberately distinct from an absent extension: `resolveHstoreTypeName` propagates its `SQLException` rather than returning null, so an unreadable catalog leaves the type *unresolved*, falls back to the bare name and lets PostgreSQL report the real problem. Only a confirmed absence produces the connector's own message. Because the type stays unresolved, the next connection retries the lookup, so a transient failure heals rather than pinning the bare name for the life of the task — the dialect's `getConnection()` runs when the cached connection is opened or replaced, not per batch. The resolution is skipped entirely for source tasks, which never write hstore.

### Changes from code review

Four behavioural corrections after review, none in the hstore mapping itself:

- **Both refusals are `ConnectException`.** An earlier revision raised them as `TableAlterOrCreateException` to gain the writer's rollback and the task's error-reporter routing. That was reverted: neither refusal is DDL, and both are deterministic for the whole stream, so the batch-splitting the DDL type triggers finds nothing to salvage and a DLQ would quietly absorb every record while the task looked healthy. Failing loudly is the intended outcome; the rollback and DLQ routing are given up deliberately. `TableAlterOrCreateException` no longer appears in the dialect at all.
- **The extension lookup is no longer latched off after a failure.** An earlier revision set a flag so a failed catalog read was never retried, on the premise that `getConnection()` runs per batch. It does not: `CachedConnectionProvider` calls the dialect only when the connection is opened or replaced, so the lookup already ran once per physical connection. The latch bought nothing and blocked recovery — one transient failure pinned the bare type name for the life of the task. Removed, so the next connection retries.
- **An unparseable `hstore[]` element degrades to null** instead of failing the task, matching what the scalar path already did for an optional column. Array element schemas are always optional, so the strict behaviour contradicted the mapping's own policy.
- **The `::hstore` cast requotes the driver-reported type name.** pgjdbc assembles `"schema"."type"` without escaping (its `TypeInfoCache` carries a `TODO: escaping !?`), so a schema containing a quote produced invalid DML. `resolveHstoreTypeName` already doubled quotes; this makes the two paths symmetric.

A later review round produced four more, none changing the mapping:

- **The refusal remediation names a creatable type.** The array branch fed pgjdbc's `_hstore` into the message, so it read *"column type is `_jsonb`, not `_hstore`. Recreate the column with the hstore type"* — which, followed literally, produces a scalar column that fails the next batch. It now renders `hstore[]` and names the expected type in the remediation.
- **The lookup state is one value.** `hstoreTypeName` plus `hstoreTypeResolved` made a reader reconstruct three states from two fields, where `null` meant either "not read" or "absent" depending on the flag — the source of two separate review questions. Replaced by an `HstoreType` value with named states and a `fromTypeName` factory, which also retires the hand-rolled write ordering the old field comment had to explain. `requireHstoreTypeName` became `hstoreTypeNameOrFail`, since it both returns a name and throws.
- **The unchecked cast in the array read is gone.** `asStringMap` existed only to host a `@SuppressWarnings("unchecked")` for a cast to `Map<String,String>` that erasure never verified — and the neighbouring `catch (ClassCastException)` could not fire for the same reason. Declaring the local as `Map<?, ?>` needs no cast at all: `connectMapToJson` takes `Object` and the method returns `Object`, so the type arguments were never load-bearing.
- **The `::hstore` cast is behind the flag.** It was ungated, so a STRING field carrying hstore text reached a pre-existing hstore column with the feature off, where the uncast parameter was previously rejected. The same early return also drops the local-name match that let a schema-qualified type reach `CAST_TYPES`, so with the flag off the method is the pre-feature expression exactly.
- **The `_` array prefix is a named constant.** `ARRAY_TYPE_PREFIX` replaces two unrelated literals — one building the prefix, one stripping it.

### Why a local `HstoreConverter` rather than pgjdbc's

pgjdbc ships `org.postgresql.util.HStoreConverter`, but it is not available to us: the driver is declared `runtime` scope in `pom.xml`, which is what packages it into the connector archive while keeping it off the compile classpath, so no dialect compiles against driver classes. Widening the scope to make one helper reachable would trade that property away; `provided` would additionally drop the driver from the shipped archive.

It is a separate class rather than private methods on the dialect for three reasons: `JsonConverter` is the same thing for JSON in the same package with the same single consumer; Debezium made the identical call, with a 74-line `HstoreConverter` in its `dialect/postgres` package rather than inlining it; and it is a pure text codec with no dialect state, which does not belong inside an already 1,500-line stateful class.

The parser accepts quoted pairs — what `hstore_out` emits (`"key"=>"value"` separated by `", "`, both sides quoted apart from a bare `NULL`), plus the surrounding whitespace and trailing comma that `hstore_in` also allows. Anything that is not a quoted-pair sequence did not come from PostgreSQL and is rejected rather than guessed at.

### Two overrides worth explaining

`statementBinder(…, TableDefinition, …)` is overridden **only** to carry the table definition into the binder. The interface default silently discards it, which leaves every `ColumnDefinition` null at bind time; `SqlServerDatabaseDialect`, `OracleDatabaseDialect` and `SybaseDatabaseDialect` all already override it for the same mechanical reason, so this closes a pre-existing gap rather than introducing a pattern — though they read the definition to choose a better bind (`setNString`, `setBlob`) rather than to reject one, and Sybase does not read it at all. It is safe: `colDef` is otherwise unused in the generic bind path, so the only consumer of the now-non-null value is the hstore column check.

`bindField(…, ColumnDefinition, String)` is overridden because it is the only bind entry point given the column definition — `maybeBindPrimitive`, where the hstore bind happens, does not receive it.

- [ ] yes
- [x] no

##### If yes, where?

`hstore` is a PostgreSQL extension type, so this is contained in `PostgreSqlDatabaseDialect`.

## Test Strategy

**Unit** — the hstore text form in both directions: quoting, escaping, a NULL value versus the string `"NULL"`, delimiters inside keys and values, the empty map, a serialize/parse round trip, and rejection of anything that is not a quoted-pair sequence. Plus `isHstoreType` across schemas, the DDL and bind for both handling modes, the schema-qualified type name for an extension in another schema, the cast following the driver-reported column type, unresolved-is-not-absent, and both new failures naming their field and column type.

The hstore text codec is covered by its own `HstoreConverterTest` in `util`, alongside `JsonConverterTest` for its sibling class. Those assertions previously lived in the dialect test; moving them adds key-order preservation (which `Map.equals` cannot show), the whitespace and trailing comma `hstore_in` tolerates, and the position named in a malformed-text failure.

**Integration** — auto-create producing a real `hstore` column and reading it back through the hstore operators; the same against an extension installed in `ext` and off the `search_path`; writing into a hand-created `ext.hstore` column; the task failing when the extension is missing; the task failing when the target column is not an hstore; source reads of a qualified `ext.hstore` and `ext.hstore[]` in map and json modes; and map-mode and json-mode round trips for both `hstore` and `hstore[]`, covering a single pair, several pairs, a NULL hstore *value*, special characters, an empty hstore, a SQL NULL column and a numeric-looking value that must stay text.

The cast gate is covered in both directions end to end: with the flag off a STRING into a pre-existing hstore column is refused by PostgreSQL, and with it on the same record lands as a real hstore, queryable through `->`.

One integration test covers the deliberate degradation rather than a success path: when the catalog read fails, the type name stays unresolved and the bare `hstore` is assumed, which is only safe because PostgreSQL then rejects it. `testUnresolvedHstoreAssumesTheBareNameAndTheWriteIsRejected` pins both routes the guess reaches — the DDL fails with `type "hstore" does not exist`, and the array bind fails in `createArrayOf` with *Unable to find server array type for provided name hstore* — against a database with the extension in `ext`. It reaches the unresolved state via an unreachable URL rather than a refused catalog read; the state and both failures are identical, and forcing a real permission error needs a restricted role plus a `REVOKE` on a system catalog.

**End to end** — `testHstoreRoundTripWithExtensionOutsideSearchPath` runs the whole feature against an extension in its own schema: a dedicated database with `CREATE EXTENSION hstore SCHEMA ext`, the source reading `ext.hstore` and `ext.hstore[]`, and the sink provisioning and writing native `hstore` and `_hstore` columns, verified through `OPERATOR(ext.->)` and `ext.akeys`. This is the case that previously failed the source task outright.

Run locally against the embedded Postgres:

- `PostgreSqlDatabaseDialectTest` — 126/126 and `HstoreConverterTest` — 8/8, so 134 unit tests for the dialect and the codec at `ecc2781a`
- `PostgresDatatypeIT` — 74/74 passing at `ecc2781a` (full class, not just the new cases; 920s against embedded Postgres)
- Semaphore green on `ecc2781a` — **1731 tests passed**
- Full unit suite locally — the only error is the pre-existing `JdbcSourceTaskLifecycleTest` Derby teardown flake (`DirectoryNotEmptyException`), which passes 10/10 in isolation and is in a file this PR does not touch
- `mvn -o validate` checkstyle clean

**Not run:** the Docker-based Postgres IT classes, including `PostgreSqlSourceDialectIT` and `PostgresSinkDialectIT`. They cannot start on a current Docker Engine: `testcontainers.version` is 1.17.3, whose docker-java speaks Engine API 1.32, while Docker 29 reports `MinAPIVersion 1.40` and rejects the client outright. That blocks 15 of the 17 Postgres IT classes for anyone on Docker 29 and is unrelated to this change — worth its own `testcontainers.version` bump. `PostgresDatatypeIT` is unaffected because it uses embedded Postgres rather than Testcontainers.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Targets `10.9.x`, to be forward-merged to `master` with the rest of the complex-types work. No new configuration.

Behaviour changes to call out in the release notes, all under `sql.complex.types.enable=true`:

- A Connect `MAP<STRING,STRING>` that previously auto-created a `jsonb` column now creates an `hstore` one, and the hstore extension becomes a requirement for any such field — including maps that did not originate from an hstore column.
- Writing a `MAP` into a **pre-existing** non-hstore column now fails with a message naming the required type — and this is a genuine behaviour change, not only a better diagnostic. Before, a `MAP` was serialized to JSON, which **every** candidate column accepted: `jsonb` and `json` via their casts, and `text`/`varchar` with no cast at all. All of those writes succeeded. They now fail. For `jsonb`/`json` the write would fail regardless once the value became hstore text (`invalid input syntax for type json`), so there the check only improves the message; for `text`/`varchar` the check itself is what converts a success into a failure. That is deliberate: silently storing `"k"=>"v"` in a text column changes the on-disk format without telling anyone. Only reachable with `sql.complex.types.enable=true`, which has never shipped.
- A source that previously **failed the task** on an hstore outside the `search_path` now maps the column. This is a fix, but it is a behaviour change: the old `search_path` error no longer exists, and such columns start flowing where they were previously fatal.

Nothing is visible with the flag at its default. That includes the `::hstore` cast: it takes the destination column's type, so it would otherwise have applied to any pipeline writing a STRING into a pre-existing hstore column, turning a rejection into a success without the flag being touched. It is gated, so a flag-off pipeline behaves exactly as on `10.9.x` — verified by running the cast over 52 column types on both and diffing, identical. The scalar `::json`, `::jsonb` and `::uuid` casts pre-date the feature and are untouched.

The complex types have not shipped, so no released pipeline is affected.

Reverting is safe and self-contained — no config, schema or offset format changes.

Note for whoever merges second: this PR and #1677 both touch `PostgreSqlDatabaseDialect.java` and `PostgresDatatypeIT.java`. I performed the merge locally to verify it — **four** conflicts, all union merges:

| Conflict | Resolution |
|---|---|
| `bindArray` javadoc | keep `bytea` in the element list **and** the `arrayElementBinding(Schema, String)` signature |
| `byteArrayFor`/`bytesOf` vs `hstoreArrayFor` | keep all three; `hstoreArrayFor` keeps its own javadoc |
| `valueTypeCast` | union on the local name — hstore first (returning the reported qualified name), then `CAST_TYPES`, then the `_`-prefixed array form |
| `testAutoCreateArrayColumnTypes` | keep both expectations: `a_hstore → _hstore` and `a_bytes → _bytea` |

That merge was verified when this branch was at `950c70db`; both branches have since moved, so the conflict list is a guide rather than a guarantee — re-run the merge before relying on it. Note that both PRs add the same `fieldName` parameter to `bindArray`/`arrayElementBinding`, so that part merges as an identical change rather than a divergence.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -85,6 +85,10 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
    * The type name to write hstore columns with, resolved once per connection: {@code hstore} when
    * the extension is on the search_path, {@code "schema".hstore} when it is installed elsewhere,
    * and null when it is not installed at all. Visible for testing.
+   *
+   * <p>{@link #hstoreTypeResolved} is written last, inside the {@code synchronized} block in
+   * {@link #getConnection()}. Both are volatile, so a reader that sees it {@code true} is
+   * guaranteed to see the name written before it, and needs no lock of its own.
    */
   volatile String hstoreTypeName;
   volatile boolean hstoreTypeResolved;
@@ -210,9 +214,18 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       if (maxIdentifierLength <= 0) {
         maxIdentifierLength = computeMaxIdentifierLength(result);
       }
-      if (!hstoreTypeResolved && complexTypesEnabled()) {
-        hstoreTypeName = resolveHstoreTypeName(result);
-        hstoreTypeResolved = true;
+      // Only the sink writes hstore columns, so a source task never pays for the lookup.
+      if (!hstoreTypeResolved && complexTypesEnabled() && config instanceof JdbcSinkConfig) {
+        try {
+          hstoreTypeName = resolveHstoreTypeName(result);
+          // Written last: see the ordering note on hstoreTypeResolved.
+          hstoreTypeResolved = true;
+        } catch (SQLException e) {
+          // Leave it unresolved rather than report the extension as missing, which would name the
+          // wrong cause. A write then assumes the bare type name and lets PostgreSQL say why.
+          log.warn("Unable to determine where the hstore extension is installed; hstore columns "
+              + "will be written as the unqualified type", e);
+        }
       }
     }
     return result;
@@ -221,9 +234,10 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
   /**
    * Locate the hstore extension: the bare name when {@code search_path} resolves it, otherwise the
    * schema it was installed into, so {@code CREATE EXTENSION hstore SCHEMA ext} is usable without
-   * changing the connection's search_path. Returns null when the extension is not installed.
+   * changing the connection's search_path. Returns null when the extension is not installed, and
+   * throws when the catalog cannot be read — the two are distinct and must not be conflated.
    */
-  static String resolveHstoreTypeName(Connection connection) {
+  static String resolveHstoreTypeName(Connection connection) throws SQLException {
     try (Statement stmt = connection.createStatement()) {
       try (ResultSet rs = stmt.executeQuery("SELECT to_regtype('hstore') IS NOT NULL")) {
         if (rs.next() && rs.getBoolean(1)) {
@@ -233,20 +247,15 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       try (ResultSet rs = stmt.executeQuery(
           "SELECT n.nspname FROM pg_extension e JOIN pg_namespace n ON n.oid = e.extnamespace"
               + " WHERE e.extname = 'hstore'")) {
-        if (rs.next()) {
-          return "\"" + rs.getString(1) + "\"." + HSTORE_TYPE_NAME;
-        }
+        return rs.next() ? "\"" + rs.getString(1) + "\"." + HSTORE_TYPE_NAME : null;
       }
-    } catch (SQLException e) {
-      log.warn("Unable to determine whether the hstore extension is installed", e);
     }
-    return null;
   }
 
   /**
-   * The type name for an hstore column, or a failure naming the field when the extension is not
-   * installed. Before any connection exists — unit tests, and validation before the first write —
-   * the bare name is assumed, since there is nothing yet to check it against.
+   * The type name for an hstore column, or a failure naming the field when the extension is known
+   * to be absent. Where it is merely unresolved — before any connection exists, or after a catalog
+   * read failed — the bare name is assumed, since there is nothing to contradict it.
    */
   private String hstoreTypeName(String fieldName) {
     if (hstoreTypeResolved && hstoreTypeName == null) {
@@ -1158,7 +1167,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
   ) throws SQLException {
     switch (schema.type()) {
       case ARRAY:
-        if (bindArray(statement, index, schema, value)) {
+        if (bindArray(statement, index, schema, value, fieldName)) {
           return true;
         }
         break;
@@ -1176,16 +1185,21 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
   /**
    * Bind a Connect ARRAY value to a native PostgreSQL array parameter. When complex types are
    * enabled, element types with a dedicated PostgreSQL array type (json, numeric and the
-   * temporals) are resolved by {@link #arrayElementBinding(Schema)} and bound via
+   * temporals) are resolved by {@link #arrayElementBinding(Schema, String)} and bound via
    * {@code createArrayOf}; any other (primitive) element type falls back to the pre-existing
    * {@link #maybeBindPrimitiveArray}. Returns false if the element type is not handled here.
    */
-  private boolean bindArray(PreparedStatement statement, int index, Schema schema, Object value)
-      throws SQLException {
+  private boolean bindArray(
+      PreparedStatement statement,
+      int index,
+      Schema schema,
+      Object value,
+      String fieldName
+  ) throws SQLException {
     Collection<?> values = arrayValueCollection(value);
     Schema elementSchema = schema.valueSchema();
     ArrayElementBinding binding =
-        complexTypesEnabled() ? arrayElementBinding(elementSchema) : null;
+        complexTypesEnabled() ? arrayElementBinding(elementSchema, fieldName) : null;
     if (binding == null) {
       return maybeBindPrimitiveArray(statement, index, elementSchema, values);
     }
@@ -1224,15 +1238,16 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
    * Resolve the native array binding for a Connect element schema, or null if the element has no
    * dedicated PostgreSQL array type (a primitive, handled by {@link #maybeBindPrimitiveArray}).
    * Only the hstore-shaped {@code MAP<STRING, STRING>} and the logical types below are mapped.
+   * The field name is carried only so a missing hstore extension can name the column it fails on.
    */
-  private ArrayElementBinding arrayElementBinding(Schema elementSchema) {
+  private ArrayElementBinding arrayElementBinding(Schema elementSchema, String fieldName) {
     if (elementSchema == null) {
       return null;
     }
     if (isStringToStringMap(elementSchema)) {
       // Each element serialized to hstore text -> hstore[].
       return new ArrayElementBinding(
-          hstoreTypeName(null), PostgreSqlDatabaseDialect::hstoreArrayFor);
+          hstoreTypeName(fieldName), PostgreSqlDatabaseDialect::hstoreArrayFor);
     }
     String elementName = elementSchema.name();
     if (elementName != null) {
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -28,6 +28,7 @@ import io.confluent.connect.jdbc.util.ColumnId;
 import io.confluent.connect.jdbc.util.DateTimeUtils;
 import io.confluent.connect.jdbc.util.ExpressionBuilder;
 import io.confluent.connect.jdbc.util.ExpressionBuilder.Transform;
+import io.confluent.connect.jdbc.util.HstoreConverter;
 import io.confluent.connect.jdbc.util.IdentifierRules;
 import io.confluent.connect.jdbc.util.JdbcCredentials;
 import io.confluent.connect.jdbc.util.JsonConverter;
@@ -53,6 +54,7 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -78,6 +80,14 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
 
   // Visible for testing
   volatile int maxIdentifierLength = 0;
+
+  /**
+   * The type name to write hstore columns with, resolved once per connection: {@code hstore} when
+   * the extension is on the search_path, {@code "schema".hstore} when it is installed elsewhere,
+   * and null when it is not installed at all. Visible for testing.
+   */
+  volatile String hstoreTypeName;
+  volatile boolean hstoreTypeResolved;
 
   /**
    * The provider for {@link PostgreSqlDatabaseDialect}.
@@ -124,6 +134,10 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
           + "not on this connection's search_path. Add the schema owning the extension to the "
           + "search_path, for example with ?currentSchema=ext,public on connection.url, or set "
           + "hstore.handling.mode=none to skip hstore columns.";
+
+  private static final String HSTORE_EXTENSION_MISSING =
+      "Cannot write %s to an hstore column: the hstore extension is not installed on this "
+          + "database. Install it with CREATE EXTENSION hstore, in any schema.";
 
   /**
    * Define the PG datatypes that require casting upon insert/update statements.
@@ -196,8 +210,50 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       if (maxIdentifierLength <= 0) {
         maxIdentifierLength = computeMaxIdentifierLength(result);
       }
+      if (!hstoreTypeResolved && complexTypesEnabled()) {
+        hstoreTypeName = resolveHstoreTypeName(result);
+        hstoreTypeResolved = true;
+      }
     }
     return result;
+  }
+
+  /**
+   * Locate the hstore extension: the bare name when {@code search_path} resolves it, otherwise the
+   * schema it was installed into, so {@code CREATE EXTENSION hstore SCHEMA ext} is usable without
+   * changing the connection's search_path. Returns null when the extension is not installed.
+   */
+  static String resolveHstoreTypeName(Connection connection) {
+    try (Statement stmt = connection.createStatement()) {
+      try (ResultSet rs = stmt.executeQuery("SELECT to_regtype('hstore') IS NOT NULL")) {
+        if (rs.next() && rs.getBoolean(1)) {
+          return HSTORE_TYPE_NAME;
+        }
+      }
+      try (ResultSet rs = stmt.executeQuery(
+          "SELECT n.nspname FROM pg_extension e JOIN pg_namespace n ON n.oid = e.extnamespace"
+              + " WHERE e.extname = 'hstore'")) {
+        if (rs.next()) {
+          return "\"" + rs.getString(1) + "\"." + HSTORE_TYPE_NAME;
+        }
+      }
+    } catch (SQLException e) {
+      log.warn("Unable to determine whether the hstore extension is installed", e);
+    }
+    return null;
+  }
+
+  /**
+   * The type name for an hstore column, or a failure naming the field when the extension is not
+   * installed. Before any connection exists — unit tests, and validation before the first write —
+   * the bare name is assumed, since there is nothing yet to check it against.
+   */
+  private String hstoreTypeName(String fieldName) {
+    if (hstoreTypeResolved && hstoreTypeName == null) {
+      throw new ConnectException(String.format(HSTORE_EXTENSION_MISSING,
+          fieldName == null ? "a map value" : "field " + fieldName));
+    }
+    return hstoreTypeName == null ? HSTORE_TYPE_NAME : hstoreTypeName;
   }
 
   static int computeMaxIdentifierLength(Connection connection) {
@@ -721,7 +777,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
 
   /**
    * Whether the schema is a {@code MAP<STRING, STRING>}, the only Connect container mapped to a
-   * native {@code jsonb} column. That is the shape a PostgreSQL {@code hstore} column takes on the
+   * native {@code hstore} column. That is the shape a PostgreSQL {@code hstore} column takes on the
    * topic; STRUCT values and other map shapes are not supported.
    */
   private static boolean isStringToStringMap(Schema schema) {
@@ -730,26 +786,20 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
         && schema.valueSchema().type() == Schema.Type.STRING;
   }
 
-  private boolean isJsonBindCandidate(Schema schema) {
-    return isStringToStringMap(schema) && complexTypesEnabled();
-  }
-
-  private boolean maybeBindJson(
+  /**
+   * Bind a Connect map as hstore text; the {@code ::hstore} cast from {@link #valueTypeCast} parses
+   * it server-side. Returns false when the schema is not the supported map shape.
+   */
+  private boolean maybeBindHstore(
       PreparedStatement statement,
       int index,
       Schema schema,
       Object value
   ) throws SQLException {
-    if (!isJsonBindCandidate(schema)) {
+    if (!isStringToStringMap(schema) || !complexTypesEnabled()) {
       return false;
     }
-    String json = JsonConverter.connectMapToJson(value);
-    if (json == null) {
-      statement.setNull(index, Types.OTHER);
-    } else {
-      // Bind as text; the ::jsonb cast (see valueTypeCast) parses it into jsonb server-side.
-      statement.setString(index, json);
-    }
+    statement.setString(index, HstoreConverter.connectMapToHstore(value));
     return true;
   }
 
@@ -937,7 +987,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
         return getSqlType(childField) + "[]";
       case MAP:
         if (isStringToStringMap(field.schema()) && complexTypesEnabled()) {
-          return JSONB_TYPE_NAME.toUpperCase();
+          return hstoreTypeName(field.name());
         }
         return super.getSqlType(field);
       default:
@@ -1113,7 +1163,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
         }
         break;
       case MAP:
-        if (maybeBindJson(statement, index, schema, value)) {
+        if (maybeBindHstore(statement, index, schema, value)) {
           return true;
         }
         break;
@@ -1180,8 +1230,9 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       return null;
     }
     if (isStringToStringMap(elementSchema)) {
-      // hstore map mode -> each element serialized to JSON text -> jsonb[].
-      return new ArrayElementBinding(JSONB_TYPE_NAME, PostgreSqlDatabaseDialect::jsonArrayFor);
+      // Each element serialized to hstore text -> hstore[].
+      return new ArrayElementBinding(
+          hstoreTypeName(null), PostgreSqlDatabaseDialect::hstoreArrayFor);
     }
     String elementName = elementSchema.name();
     if (elementName != null) {
@@ -1255,12 +1306,12 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
   }
 
   /**
-   * Serialize each map element to JSON text for binding into a native {@code jsonb[]} column, as an
-   * hstore column in map mode produces. Null elements are preserved.
+   * Serialize each map element to hstore text for binding into a native {@code hstore[]} column.
+   * Null elements are preserved.
    */
-  private static Object[] jsonArrayFor(Collection<?> valueCollection) {
+  private static Object[] hstoreArrayFor(Collection<?> valueCollection) {
     return valueCollection.stream()
-        .map(o -> o == null ? null : JsonConverter.connectMapToJson(o))
+        .map(o -> o == null ? null : HstoreConverter.connectMapToHstore(o))
         .toArray();
   }
 
@@ -1366,13 +1417,15 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
   protected String valueTypeCast(TableDefinition tableDefn, ColumnId columnId) {
     if (tableDefn != null) {
       ColumnDefinition defn = tableDefn.definitionForColumn(columnId.name());
-      if (defn != null) {
+      if (defn != null && defn.typeName() != null) {
         String typeName = defn.typeName(); // database-specific
-        if (typeName != null) {
-          typeName = typeName.toLowerCase();
-          if (CAST_TYPES.contains(typeName)) {
-            return "::" + typeName;
-          }
+        String localName = localTypeName(typeName).toLowerCase();
+        if (HSTORE_TYPE_NAME.equals(localName)) {
+          // The reported name, so an extension off the search_path stays schema-qualified.
+          return "::" + typeName;
+        }
+        if (CAST_TYPES.contains(localName)) {
+          return "::" + localName;
         }
       }
     }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -22,6 +22,7 @@ import io.confluent.connect.jdbc.sink.JdbcSinkConfig;
 import io.confluent.connect.jdbc.sink.JdbcSinkConfig.InsertMode;
 import io.confluent.connect.jdbc.sink.JdbcSinkConfig.PrimaryKeyMode;
 import io.confluent.connect.jdbc.sink.PreparedStatementBinder;
+import io.confluent.connect.jdbc.sink.TableAlterOrCreateException;
 import io.confluent.connect.jdbc.sink.metadata.FieldsMetadata;
 import io.confluent.connect.jdbc.sink.metadata.SchemaPair;
 import io.confluent.connect.jdbc.sink.metadata.SinkRecordField;
@@ -92,10 +93,6 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
    */
   volatile String hstoreTypeName;
   volatile boolean hstoreTypeResolved;
-
-  /** Set when the catalog is unreadable, so the lookup runs once. Not an absence
-   * and on every connection */
-  private volatile boolean hstoreTypeLookupFailed;
 
   /**
    * The provider for {@link PostgreSqlDatabaseDialect}.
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -196,10 +213,82 @@ public Connection getConnection() throws SQLException {
...
+  void maybeResolveHstoreType(Connection connection) {
+    if (hstoreTypeResolved) {
+      return;
+    }
+    try {
+      hstoreTypeName = resolveHstoreTypeName(connection);
+      hstoreTypeResolved = true;
+    } catch (SQLException e) {
+      // Unresolved, not absent: a write assumes the bare name and lets PostgreSQL say why.
+      log.warn("Could not locate the hstore extension; assuming the bare type name. Writes fail "
```
akanimesh7: Wasn't it a better design choice to fail when the extension isn't enabled, rather than warn and move on?

Having read it again I think you've already got that covered — absent returns null and `requireHstoreTypeName` throws, and `hstore.handling.mode` is a source config anyway so it isn't what gates this path. So this branch is only the catalog-read failure. Can we make the log say that? "Could not locate the hstore extension" reads like "it isn't installed", which is the one thing it doesn't mean. _🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -196,10 +213,82 @@ public Connection getConnection() throws SQLException {
...
+    }
+  }
+
+  /** Quote an identifier, doubling embedded quotes: a schema name may legally contain one. */
+  private static String quoteIdentifier(String identifier) {
+    return "\"" + identifier.replace("\"", "\"\"") + "\"";
+  }
+
+  /** Requote a reported type name, doubling any embedded quote the driver left as-is. */
+  private static String requoteTypeName(String typeName) {
```
akanimesh7, line 316: Not readable — this is quote-and-unquote and it's unclear what it's for. Please add an example in the javadoc.

On tests: `shouldCastHstoreColumnsByTheirReportedTypeName` already covers the qualified name, so the gap is narrower than I first thought — what's untested is the case this function exists for, a schema name containing a `"`. pgjdbc concatenates `"schema"."type"` without escaping, so that's the input that would otherwise produce broken SQL. Worth one test. _🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -662,17 +736,33 @@ private List<Object> readElementResultSet(Array arr, String elementType) throws
...
+    }
     return hstoreHandlingMode() == HstoreHandlingMode.JSON
-        ? JsonConverter.connectMapToJson(value)
-        : value;
+        ? JsonConverter.connectMapToJson(map)
+        : map;
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Map<String, String> asStringMap(Object value) {
```
akanimesh7: This function looks redundant.

Also `catch (DataException | ClassCastException e)` above it — erasure means the cast only checks `Map`, which `value instanceof Map` already guaranteed, so the `ClassCastException` arm can't fire. Not a live bug (pgjdbc always gives us `Map<String,String>`), but it reads like a guard that's doing something. Drop the helper and the CCE arm together? _🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -730,26 +820,79 @@ private static boolean isStringToStringMap(Schema schema) {
...
+    );
   }
 
-  private boolean maybeBindJson(
+  /**
+   * {@inheritDoc} A map is written as hstore text, so the column must be an hstore. Checked here
+   * as the only bind entry point given the column definition.
+   */
+  @Override
+  public void bindField(
```
akanimesh7, line 906: To me this looks more like a validation. Based on the schema and the value we have expectations about the column this data goes into, and if the column definition doesn't align, that's what we're validating. Could this move up into the generic dialect — a validate hook before any bind work, empty for other dialects, implemented for Postgres?

`colDef` is currently unused throughout `bindFieldInternal` in `GenericDatabaseDialect`, so it's threaded through the generic path with no consumer there; a hook would give it one and avoid overriding all of `bindField`. Happy to be told the extra method surface isn't worth it. _🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -730,26 +820,79 @@ private static boolean isStringToStringMap(Schema schema) {
...
+      Schema schema,
+      Object value,
+      ColumnDefinition colDef,
+      String fieldName
+  ) throws SQLException {
+    if (value != null && complexTypesEnabled()) {
+      if (isStringToStringMap(schema)) {
+        requireHstoreColumn(colDef, fieldName, HSTORE_TYPE_NAME);
+      } else if (schema.type() == Schema.Type.ARRAY && isStringToStringMap(schema.valueSchema())) {
+        requireHstoreColumn(colDef, fieldName, "_" + HSTORE_TYPE_NAME);
```
akanimesh7: `"_" + HSTORE_TYPE_NAME` should be defined as another constant — the `_` prefix convention also shows up at line 613 with no shared name.

Related: this branch feeds `_hstore` into `HSTORE_COLUMN_REQUIRED`, so the array failure reads *"column type is _jsonb, not _hstore. Recreate the column with the hstore type"*. `_hstore` is pgjdbc's internal name — the type a user would actually write is `hstore[]`. Can the message use the array form for this branch? Nothing asserts the text today. _🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -730,26 +820,79 @@ private static boolean isStringToStringMap(Schema schema) {
...
+
+  /** Fail unless an existing column can hold hstore text; unknown column or type is skipped. */
+  private void requireHstoreColumn(ColumnDefinition colDef, String fieldName, String expected) {
+    String actual = colDef == null ? null : localTypeName(colDef.typeName());
+    if (actual == null) {
+      return;
+    }
+    if (!expected.equalsIgnoreCase(actual)) {
+      throw new TableAlterOrCreateException(
+          String.format(HSTORE_COLUMN_REQUIRED, fieldName, actual, expected));
```
akanimesh7, line 937: Exception type looks incorrect. Every other `TableAlterOrCreateException` is thrown from `DbStructure`, i.e. DDL. This one is at bind time and it inherits that type's handling in `JdbcSinkTask`: straight to `unrollAndRetry`, skipping `maxRetries`/`retryBackoffMs`.

That policy fits "the table can't be adapted" — split the batch to find which records don't fit. Here it's deterministic for every record sharing the schema, so the whole batch fails anyway and we pay a `closeQuietly()` + `initWriter()` per record before DLQ'ing each one, having bypassed the retry policy the user configured.

Line 292 is fine — that one comes from `getSqlType` during DDL, so it genuinely is a table-create failure. Only this one. Use a type whose handling we want, or add a dedicated exception with its own catch arm? Neither existing arm is a clean fit, so worth deciding rather than inheriting DDL semantics by accident.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -196,10 +213,82 @@ public Connection getConnection() throws SQLException {
...
+      // Only the sink writes hstore columns, so a source task never pays for the lookup.
+      if (complexTypesEnabled() && config instanceof JdbcSinkConfig) {
+        maybeResolveHstoreType(result);
+      }
     }
     return result;
   }
 
+  /**
+   * Resolve the type name once per connection. A failed read leaves it unresolved, not absent, so
```
akanimesh7: "once per connection" — `hstoreTypeResolved` is instance state, so it's once per dialect instance; a reconnect on the same dialect keeps the first answer. Narrow in practice, and `initWriter()` builds a fresh dialect on the retry paths anyway. Comment fix, not a code fix. _🤖 flagged by Claude_

akanimesh7 (commented):
Overall this is solid — the unresolved-vs-absent distinction and the requoting of pgjdbc's unescaped qualified name are both good catches that I'd have missed. Description and test plan are thorough, CI is green. Comments below are mostly readability and placement; two are worth a real answer (the `catch` log message, and the exception type at the hstore column check). Nothing blocking a merge once those are addressed.

One out-of-scope observation while reading this: `ExpressionBuilder.appendIdentifierQuoted` appends the raw name between quotes with no escaping at all, so any table or column name containing a `"` builds broken SQL. That's repo-wide and predates this PR — this is actually the first place that handles the doubling correctly. Worth its own ticket rather than folding in here.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Done. Updated the doc.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Done. Added the helpful to make it as `hstore[]`

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Updated the exception as ConnectException similar to how it is being handled in other scenarios.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: The base implementation was never meant to be the consumer. colDef isn't consumer-less: Oracle , SQL Server and Sybase each read it inside their own bindField overrides. The parameter exists so dialect overrides can consume it and doing the same in Postgres also

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -730,26 +832,93 @@ private static boolean isStringToStringMap(Schema schema) {
...
+      throw new ConnectException(String.format(HSTORE_COLUMN_REQUIRED,
+          fieldName, writableTypeName(actual), writableTypeName(expected)));
+    }
+  }
+
+  /**
+   * The form an operator would write in DDL: pgjdbc's {@code _hstore} becomes {@code hstore[]}.
+   * Both name the same type, but only the latter is what they would type to create the column.
+   */
+  private static String writableTypeName(String typeName) {
```
akanimesh7, line 945: Seems unnecessary. What was the issue with `_jsobb` ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -662,17 +753,28 @@ private List<Object> readElementResultSet(Array arr, String elementType) throws
...
-   * decoding, so the value arrives as a Map; json mode then serializes it.
+   * Decode one hstore array element. The element {@link ResultSet} yields a Map or raw text, so
+   * both are normalised as the scalar path does; json mode then serializes the result.
    */
   private Object decodeHstoreElement(ResultSet elementRs) throws SQLException {
     Object value = elementRs.getObject(ELEMENT_VALUE_COLUMN);
     if (value == null) {
       return null;
     }
+    Map<?, ?> map;
```
akanimesh7, line 803: Why not Map<String,String> ?

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
```diff
@@ -945,22 +950,40 @@ public void shouldDropHstoreWhenComplexTypesDisabled() {
...
+   * input and must decode to the same map a resolved column would give.
+   */
   @Test
-  public void hstoreValueThatIsNotAMapShouldFollowColumnNullability() throws Exception {
-    // Defence in depth for any non-Map shape the driver might hand back. Off the search_path the
-    // type name is qualified, so the column is skipped before this runs; whatever else could reach
-    // here has no known cause. Follows Debezium's handleUnknownData: a nullable column degrades to
-    // null, a NOT NULL column fails because null would breach its schema anyway.
-    ResultSet rawText = hstoreResultSet("\"env\"=>\"prod\"");
+  public void hstoreRawTextShouldDecodeLikeADriverMap() throws Exception {
```
akanimesh7, line 957: Looks better suited to a separate test file for HstoreConverter

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -196,10 +217,93 @@ public Connection getConnection() throws SQLException {
...
+  private static String unquoteOuter(String part) {
+    return part.length() > 1 && part.startsWith("\"") && part.endsWith("\"")
+        ? part.substring(1, part.length() - 1)
+        : part;
+  }
+
+  /** Type name, or a rollback-and-reportable failure when absent. Unresolved assumes bare. */
+  private String requireHstoreTypeName(String fieldName) {
+    if (hstoreTypeResolved && hstoreTypeName == null) {
+      throw new TableAlterOrCreateException(String.format(HSTORE_EXTENSION_MISSING,
```
akanimesh7: Why is this still `TableAlterOrCreateException` ?

akanimesh7: Also this path is not in any of the tests. Is this reproducible ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -196,10 +217,93 @@ public Connection getConnection() throws SQLException {
...
+   * reconnect keeps the first answer. A failed read leaves it unresolved rather than absent, so the
+   * next connection on this instance retries and a transient failure heals. Visible for testing.
+   */
+  void maybeResolveHstoreType(Connection connection) {
+    if (hstoreTypeResolved) {
+      return;
+    }
+    try {
+      hstoreTypeName = resolveHstoreTypeName(connection);
+      hstoreTypeResolved = true;
```
akanimesh7: hstoreTypeName might be null here. But still hstoreTypeResolved is set to true. Later on, in `requireHstoreTypeName` if hstoreTypeName is null, it returns an exception. Can't that exception be returned here itself ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: There is no issue with `_jsonb` ?

This has came as a previous comment above. As, I felt that json[] is more readable and user can understand it easily, I had added this method.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: I wanted to keep that the table create or alter failures with this exception. But as all others are ConnectException hence made it to ConnectException. Also, added an IT here for this to get the scenario - `testMapArrayFailsWhenExtensionIsNotInstalled`

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Definitely, hstoreTypeName might be null but if there is a database which doesn't have the HSTORE extension and it is not required by the user as there are no HSTORE values, then when hstoreTypeName is NULL, it would throw the exception even if there is no requirement.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: This has been done as work-around to this comment - https://github.com/confluentinc/kafka-connect-jdbc/pull/1676#discussion_r3765291436

src/test/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialectTest.java
KGaneshDatta: Done added the new HstoreConverter Test class.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
akanimesh7: nit -- I feel this is redundant.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
akanimesh7: Ah i see. It runs unconditionally, since it's triggered from `getConnection`. 

That's not right. It should ideally run only when hstore coulumn is being dealth with. That way, the code is more readable and less error prone. Current behaviour is very confusing.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -295,10 +294,14 @@ private static String unquoteOuter(String part) {
         : part;
   }
 
-  /** Type name, or a rollback-and-reportable failure when absent. Unresolved assumes bare. */
+  /**
+   * Type name, or a failure when the extension is absent; unresolved assumes bare. Absence is a
+   * property of the database, not of a record, so no batch split could find a record that fits:
+   * this fails the task rather than carrying the DDL exception's per-record handling.
+   */
   private String requireHstoreTypeName(String fieldName) {
```
akanimesh7: This function does two things — returns the type name and throws when the extension is absent — and the name conveys neither, nor that the name can come back schema-qualified.

More broadly, encoding availability across hstoreTypeResolved + hstoreTypeName makes the reader reconstruct a tri-state from two fields. Can we model it as one value instead — an enum or a small holder that says on-path / qualified / absent / unresolved — so the call sites read directly? Resolution can stay in getConnection() where the connection exists; it's the representation I'd like clearer, not the timing.

akanimesh7 (commented):
Name of function should be changed.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack. Updated it.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Now it's one value instead of two fields: HstoreType is either UNRESOLVED, ABSENT, or INSTALLED. Call sites just ask isAbsent() or nameOrBare(). I renamed the method to hstoreTypeNameOrFail so the name says what it does. I didn't split on-path and qualified — no code checks the difference.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: As, you suggeted later - Timing stays in `getConnection()`. The confusing part is two fields and is now one value, `HstoreType`.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -86,12 +86,48 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
...
+    private final State state;
+    private final String name;
+
+    private HstoreType(State state, String name) {
+      this.state = state;
+      this.name = name;
+    }
+
+    /** The catalog answered; a null name means the extension is not installed. */
+    static HstoreType installedAs(String name) {
```
akanimesh7: better name -- fromTypeName

akanimesh7 (dismissed):

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -196,10 +252,97 @@ public Connection getConnection() throws SQLException {
...
+    if (hstoreType.isResolved()) {
+      return;
+    }
+    try {
+      hstoreType = HstoreType.installedAs(resolveHstoreTypeName(connection));
+    } catch (SQLException e) {
+      // Unresolved, not absent: a write assumes the bare name and lets PostgreSQL say why.
+      log.warn("Could not read the catalog to find where the hstore extension is installed — this "
+          + "does not mean it is absent, which is reported separately when a write needs it. "
+          + "Assuming the bare type name; writes fail if it turns out to live off the search_path.",
```
akanimesh7, line 278: The unresolved path returns the bare `hstore` and relies on PostgreSQL failing the write later. Nothing tests that — the two unit tests only assert the bare name comes back, and every off-search_path IT runs with resolution already succeeded.

Please add a test pinning it: unresolved, extension in `ext`, then a write.

akanimesh7 (approved):

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
akanimesh7: @KGaneshDatta Please check this comment before merging.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -1366,13 +1561,20 @@ protected Transform<ColumnId> columnValueVariables(TableDefinition defn) {
...
+          // The hstore cast and the local-name match are both new, so off is the pre-feature form.
+          String rawName = typeName.toLowerCase();
+          return CAST_TYPES.contains(rawName) ? "::" + rawName : "";
+        }
+        String localName = localTypeName(typeName).toLowerCase();
+        if (HSTORE_TYPE_NAME.equals(localName)) {
+          // Requoted, not interpolated: pgjdbc builds the qualified name without escaping.
+          return "::" + requoteTypeName(typeName);
+        }
+        if (CAST_TYPES.contains(localName)) {
```
akanimesh7, line 1576: Rather than duplicating this, isn't the below better ?

```
String typeName = defn.typeName();
String rawName = typeName.toLowerCase();
if (CAST_TYPES.contains(rawName)) {
  return "::" + rawName;
}

String localName = localTypeName(typeName).toLowerCase();
if (complexTypesEnabled() && HSTORE_TYPE_NAME.equals(localName)) {
  // Requoted, not interpolated: pgjdbc builds the qualified name without escaping.
  return "::" + requoteTypeName(typeName);
}

return "";

```

PR #1677, opened 2026-08-10
Fix gated array element failures and add bytea[]/uuid[] support for PostgreSQL

KGaneshDatta:
## Problem

Two related gaps in the PostgreSQL native array support added by #1651.

**1. A gated array element failed opaquely instead of being refused.**

`logicalSqlType` gates `Json` and `VariableScaleDecimal` behind `sql.complex.types.enable`, but not `Decimal`, `Date`, `Time` or `Timestamp`. With the flag off, a `Timestamp` array therefore still advertised a `timestamp[]` column, which auto-create happily built and no insert could ever populate: the bind fell through to the primitive path, which tried to store `java.util.Date` values into a `Long[]` and raised a bare `java.lang.ArrayStoreException`.

So the operator saw a JVM error with no indication that a config was responsible, against a column the connector had itself just created. `Decimal` had the milder form of the same defect — the DDL succeeded and every insert then failed.

This is pre-existing behaviour on `master`, not a regression introduced by the complex-types stack.

**2. `bytea[]` and `uuid[]` were skipped entirely.**

Neither was in the supported element set, so the columns were dropped with a WARN even though the scalar `bytea` and `uuid` types are both mapped.

## Solution

**1.** A single gate, raised from both the DDL path and the bind path, so an operator gets the same actionable message either way:

> Cannot write an array of `org.apache.kafka.connect.data.Timestamp` into column `a_ts` while `sql.complex.types.enable` is false; set it to true to map complex column types.

`Json` deliberately keeps degrading to `text[]` — it is STRING-based, so the fallback is lossless and useful for a partially upgraded pipeline. Only the elements that genuinely cannot bind are refused.

**2.** `bytea` elements map to Connect `BYTES`, `uuid` elements to canonical text (matching the scalar `uuid` mapping, which has no Connect logical type).

Making `bytea[]` work required fixing the multi-dimensional guard first. `getArray()` returns `byte[][]` for a `bytea[]` column, and the guard asked `element.getClass().isArray()` — true for every `byte[]` — so every `bytea[]` would have been read as multi-dimensional and nulled out. It now asks `element instanceof Object[]` instead: `byte[]` has a primitive component so it is not an `Object[]`, while a real `bytea[][]` arrives as `byte[][][]` whose elements are. That one question is correct for every element type, because `bytea` is the only one whose 1-D value is itself an array — `org.postgresql.jdbc.ArrayDecoding` declares a reference-array decoder for every type with the single exception of `ArrayDecoder<byte[][]> BYTE_ARRAY_ARRAY` — so no element-type special case is needed.

On the sink, `bytea` elements bind through `createArrayOf`, accepting both `byte[]` and `ByteBuffer` since Connect permits either. `uuid` has no logical type and travels as `ARRAY<STRING>`, so `valueTypeCast` now also casts the array form of a cast type — letting a `text[]` parameter reach a pre-existing `uuid[]` column via `::uuid[]`.

That match deliberately does **not** normalise a schema-qualified type name. Only extension types are ever reported qualified, and `json`, `jsonb` and `uuid` are all built in, so the case cannot arise; an earlier revision handled it and the handling was removed as speculative.

### Changes from code review

Five corrections after review. The mapping itself is unchanged; all five are about refusing correctly and pinning what matters.

- **The gate is derived, not enumerated.** `isComplexArrayElement` now asks `arrayElementBinding(element) != null` instead of consulting a hand-maintained set of logical names, with `Json` as the one deliberate exception. That closes the gap where `VariableScaleDecimal` — which *does* have a binding — was refused by the DDL path but fell through the bind path to a generic *"Unsupported source data type: ARRAY"*, defeating the "same actionable message either way" goal above. The enumerated set is gone, so the two can no longer drift.
- **A BYTES element no binding claims is no longer advertised.** `isUnbindableArrayElement` now also refuses it, so a named logical BYTES type the dialect cannot write (`io.debezium.data.Bits`, or any custom one) is skipped with a WARN instead of auto-creating a `bytea[]` column every insert then fails on — the exact failure mode item (a) of this PR exists to remove. Pre-existing, but this PR re-homed that predicate, so it belongs here.
- **Unbindable is checked before the flag gate.** Previously a field in `timestamp.fields.list` with a temporal element was told to set `sql.complex.types.enable=true`, and doing so produced a second, less helpful failure. Enabling the feature cannot help an element nothing can write, so that advice is no longer given.
- **The array form of `valueTypeCast` is behind the flag.** It was ungated, which made a previously-rejected `ARRAY<STRING>` into a pre-existing `uuid[]`/`jsonb[]` column start succeeding on default config. Gated so a flag-off pipeline is byte-identical to `10.9.x`; the scalar cast above it is untouched.
- **`complexArrayElementDisabledError` raises `ConnectException`.** An earlier revision used `TableAlterOrCreateException` to gain the writer's rollback and the task's error-reporter routing. That was reverted on review: the refusal is not a DDL failure — one throw site is `bindArray` — and `sql.complex.types.enable=false` holds for the whole connector, so no record can succeed and splitting the batch to report each in turn finds nothing to salvage. It now fails the task, as `GenericDatabaseDialect` does for any value it cannot bind, matching #1676's refusals. The rollback and DLQ routing are given up deliberately, not overlooked.
- **The `bytea[]` bind tests now pin the runtime component type.** They compared elements but never the array's component type, which is the whole correctness condition — pgjdbc rejects a `byte[]` nested in an `Object[]`. Proven by mutation: changing `.toArray(byte[][]::new)` to `.toArray()` previously left all 120 unit tests green and now fails two.

Two further corrections after a later review pass:

- **The multi-dimensional guard drops its element-type parameter.** `isMultiDimensional(Object[], String)` becomes `isMultiDimensional(Object[])`, reverting a signature change this PR had introduced. See the guard's description above — the general form is correct for all element types, so the `bytea` branch was redundant rather than wrong. Confirmed by mutation: `shouldReadByteaArrayElements` and `shouldSkipMultiDimensionalByteaArrayElements` both stay green, while removing the check entirely fails the first with `expected:<[[B@…, null, [B@…]> but was:<[null, null, null]>`.
- **`complexArrayElementDisabled` renamed to `complexArrayElementDisabledError`.** It reads as a predicate but returns an exception; the file's own convention for that shape is `hstoreOffSearchPathError`.

### Second commit

`01f68672` is refinement only, no behaviour change:

- the `uuid` element folds into the existing `json`/`jsonb` case in `mapArrayElement`, since all three take the raw text
- `bindArray` keeps its original body and gains only a three-line guard, rather than restructuring the branch that resolves the element binding
- `isMultiDimensional` drops a hoisted local that bought nothing, as the loop returns on the first non-null element
- documentation corrections in the same code: the `@link` to `isMultiDimensional` names its current signature, the element-type set no longer claims its values are `Struct`s and says why `VariableScaleDecimal` is excluded, and `readMappedElements`, `mapArrayElement` and `valueTypeCast` describe the `bytea` and `uuid` elements they now handle

- [ ] yes
- [x] no

##### If yes, where?

PostgreSQL is the only dialect that maps native arrays, so both gaps are contained in `PostgreSqlDatabaseDialect`.

## Test Strategy

Unit tests cover the schema mapping for `_bytea`/`_uuid`, the multi-dimensional guard firing on `bytea[][]` but not `bytea[]`, the `bytea` bind for both `byte[]` and `ByteBuffer`, the array form of `valueTypeCast`, and the gate refusing each affected element type on both the DDL and bind paths.

Integration tests cover reading `bytea[]`/`uuid[]` including empty, NULL element and NULL column; a driver-contract test pinning that pgjdbc returns `byte[][]` and `UUID[]`, which the mocked unit tests assume; auto-create producing `_bytea`; the gated temporal array failing with the config named and no table created; and source-to-sink round trips for both `bytea[]` and `uuid[]`.

Run locally:

- `PostgreSqlDatabaseDialectTest` — 120/120 passing at `47aa28dc`
- `PostgresDatatypeIT` — 72/72 passing (full class, not just the new cases). Last full local run predates the two later review items above, both of which are covered by unit tests; Semaphore green on `47aa28dc`
- Full unit suite — 1567 tests, no failures attributable to this change. `JdbcSourceTaskLifecycleTest` is flaky in this repo independently of this PR — it has produced both a Derby `DirectoryNotEmptyException` teardown error and an EasyMock *"Unexpected method calls: CachedConnectionProvider.getConnection()"* on unmodified branches, and passes 10/10 in isolation. Neither involves a Postgres dialect.
- `mvn -o validate` checkstyle clean

**Not run:** the Docker-based Postgres IT classes, including `PostgreSqlSourceDialectIT` and `PostgresSinkDialectIT`, which cover the *scalar* `bytea` and `uuid` paths. They cannot start on a current Docker Engine: `testcontainers.version` is 1.17.3, whose docker-java speaks Engine API 1.32, while Docker 29 reports `MinAPIVersion 1.40` and rejects the client outright. That blocks 15 of the 17 Postgres IT classes for anyone on Docker 29 and is unrelated to this change — worth its own `testcontainers.version` bump. `PostgresDatatypeIT` is unaffected because it uses embedded Postgres rather than Testcontainers.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

Targets `10.9.x`, to be forward-merged to `master` with the rest of the complex-types work. No new configuration. `bytea[]`/`uuid[]` support is only reachable with the flag on.

One change is visible with `sql.complex.types.enable` at its default, and nothing that previously worked stops working:

- A temporal or decimal array now fails with a clear message naming the config, instead of an `ArrayStoreException`, and no unwritable column is created. Where the column already exists the failure stays at bind time and only the message improves; with `auto.create` it moves to DDL, so the unwritable column is no longer created at all.

Everything else is behind the flag, including the cast. An earlier revision left the array form of `valueTypeCast` ungated, on the grounds that it completes cast logic that pre-dates the feature: an `ARRAY<STRING>` into a pre-existing `uuid[]` column previously failed with *column "u" is of type uuid[] but expression is of type character varying[]* and would have started succeeding via `?::uuid[]`. That was gated on review — a failure turning into a success is still a flag-off behaviour change, and an upgrade is easier to reason about when the default config behaves exactly as it did before. `uuid[]` writes into a pre-existing column therefore require the flag, which is consistent with the rest of the feature.

Verified by running the same cast over 52 column types on this branch with the flag off and on `10.9.x`, and diffing: identical. With the flag on, the only additions are `_json`, `_jsonb` and `_uuid`; the scalar `::json`, `::jsonb` and `::uuid` casts fire either way, since those pre-date the feature.

Reverting is safe and self-contained — no config, schema or offset format changes.

Note for whoever merges second: this PR and #1676 both touch `PostgreSqlDatabaseDialect.java` and `PostgresDatatypeIT.java`. I performed the merge locally to verify it — **four** conflicts, all union merges:

| Conflict | Resolution |
|---|---|
| `bindArray` javadoc | keep `bytea` in the element list **and** the `arrayElementBinding(Schema, String)` signature |
| `byteArrayFor`/`bytesOf` vs `hstoreArrayFor` | keep all three; `hstoreArrayFor` keeps its own javadoc |
| `valueTypeCast` | union on the local name — hstore first (returning the reported qualified name), then `CAST_TYPES`, then the `_`-prefixed array form |
| `testAutoCreateArrayColumnTypes` | keep both expectations: `a_hstore → _hstore` and `a_bytes → _bytea` |

That merge was verified when both branches were at their earlier heads; both have since moved, so treat the conflict list as a guide and re-run the merge before relying on it.

One trap worth naming: only `bindArray` gains `fieldName` in both PRs. `arrayElementBinding` does **not** — this PR keeps `arrayElementBinding(Schema)` while #1676 changes it to `arrayElementBinding(Schema, String)`. Resolving the `bindArray` javadoc hunk by "keeping both" can therefore preserve the 1-arg form and silently drop #1676's field naming from its missing-extension error, or leave two overloads. This PR also carries three `{@link #arrayElementBinding(Schema)}` references that become dangling signatures once #1676 lands, and need updating in the same pass.

### DIFF
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -580,7 +580,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
 
   /**
    * Read a PostgreSQL array column into a Connect list. The shared handling — SQL NULL, skipping
-   * multi-dimensional columns (see {@link #isMultiDimensional(Object[], String)}), and freeing the
+   * multi-dimensional columns (see {@link #isMultiDimensional(Object[])}), and freeing the
    * array — lives here. Temporal elements are decoded via the element {@link ResultSet} so each
    * can honor the configured {@code db.timezone}; all other elements come straight from
    * {@code getArray()}.
@@ -596,7 +596,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       if (raw == null) {
         return null;
       }
-      if (raw instanceof Object[] && isMultiDimensional((Object[]) raw, elementType)) {
+      if (raw instanceof Object[] && isMultiDimensional((Object[]) raw)) {
         if (multiDimensionalWarnedColumns.add(columnId)) {
           log.warn(MULTI_DIMENSIONAL_ARRAY_MESSAGE, columnId);
         } else {
@@ -724,14 +724,15 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
    * Postgres allows these, but JDBC reports the same type name as 1-D, so they are only
    * detectable from values; with no 1-D Connect ARRAY schema, callers skip them, returning null.
    *
-   * <p>A {@code bytea} element is legitimately a {@code byte[]}, so for that type only a nested
-   * {@code Object[]} — {@code byte[][][]}, from a {@code bytea[][]} column — is multi-dimensional.
+   * <p>Tested as an object array rather than any array: {@code bytea} is the one element type whose
+   * value is itself a {@code byte[]}, and its primitive component keeps it out of {@code Object[]}
+   * while a real {@code bytea[][]} still nests. Every other decoder yields a reference array, so
+   * the same question is correct for all of them.
    */
-  private static boolean isMultiDimensional(Object[] elements, String elementType) {
+  private static boolean isMultiDimensional(Object[] elements) {
     for (Object element : elements) {
       if (element != null) {
-        return BYTEA_TYPE_NAME.equals(elementType)
-            ? element instanceof Object[] : element.getClass().isArray();
+        return element instanceof Object[];
       }
     }
     return false;
@@ -889,7 +890,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
    * Shared by the DDL and bind paths for one actionable message. Not a DDL failure: the flag is
    * off for the whole connector, so no record can succeed and a batch split would find nothing.
    */
-  private static ConnectException complexArrayElementDisabled(
+  private static ConnectException complexArrayElementDisabledError(
       Schema element, String fieldName) {
     return new ConnectException(String.format(COMPLEX_ARRAY_ELEMENT_DISABLED,
         element.name() == null ? element.type() : element.name(),
@@ -976,7 +977,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
           return super.getSqlType(field);
         }
         if (!complexTypesEnabled() && isComplexArrayElement(element)) {
-          throw complexArrayElementDisabled(element, field.name());
+          throw complexArrayElementDisabledError(element, field.name());
         }
         SinkRecordField childField = new SinkRecordField(
             element,
@@ -1191,7 +1192,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
     Schema elementSchema = schema.valueSchema();
     if (!complexTypesEnabled() && isComplexArrayElement(elementSchema)) {
       // The primitive fallback would fail on the element's Java type rather than say why.
-      throw complexArrayElementDisabled(elementSchema, fieldName);
+      throw complexArrayElementDisabledError(elementSchema, fieldName);
     }
     ArrayElementBinding binding =
         complexTypesEnabled() ? arrayElementBinding(elementSchema) : null;
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -1454,8 +1454,9 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
           if (CAST_TYPES.contains(typeName)) {
             return "::" + typeName;
           }
-          // The array of a cast type needs the same cast, e.g. text[] into a uuid[] column.
-          if (typeName.startsWith("_") && CAST_TYPES.contains(typeName.substring(1))) {
+          // Behind the flag: the array form is new, the scalar cast above pre-dates the feature.
+          if (complexTypesEnabled()
+              && typeName.startsWith("_") && CAST_TYPES.contains(typeName.substring(1))) {
             return "::" + typeName.substring(1) + "[]";
           }
         }
--- a/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
+++ b/src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
@@ -48,6 +48,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.math.BigDecimal;
+import java.nio.ByteBuffer;
 import java.sql.Array;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -103,6 +104,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
   static final String TIME_TYPE_NAME = "time";
   static final String TIMESTAMP_TYPE_NAME = "timestamp";
   static final String TIMESTAMPTZ_TYPE_NAME = "timestamptz";
+  static final String BYTEA_TYPE_NAME = "bytea";
 
   // Array.getResultSet() yields two columns per element: 1 is the index, 2 the value.
   private static final int ELEMENT_VALUE_COLUMN = 2;
@@ -135,6 +137,25 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
           UUID_TYPE_NAME
       ))
   );
+
+  /**
+   * Logical array element types that only {@link #arrayElementBinding(Schema)} can write. Their
+   * Connect values are {@code java.util.Date}/{@code Struct}, so the primitive fallback cannot bind
+   * them; without the complex types they must fail rather than advertise an unwritable column.
+   * {@code Json} is excluded deliberately: it is STRING-based and degrades losslessly to text[].
+   */
+  private static final Set<String> COMPLEX_ARRAY_ELEMENT_TYPES = Collections.unmodifiableSet(
+      new HashSet<>(Arrays.asList(
+          Decimal.LOGICAL_NAME,
+          Date.LOGICAL_NAME,
+          Time.LOGICAL_NAME,
+          Timestamp.LOGICAL_NAME
+      ))
+  );
+
+  private static final String COMPLEX_ARRAY_ELEMENT_DISABLED =
+      "Cannot write an array of %s into column %s while %s is false; set it to true to map "
+          + "complex column types.";
 
   /**
    * Create a new dialect instance with the given connector configuration.
@@ -478,6 +499,11 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       case "bool":
       case "boolean":
         return Schema.OPTIONAL_BOOLEAN_SCHEMA;
+      case BYTEA_TYPE_NAME:
+        return Schema.OPTIONAL_BYTES_SCHEMA;
+      case UUID_TYPE_NAME:
+        // Rendered as text, matching the scalar uuid mapping.
+        return Schema.OPTIONAL_STRING_SCHEMA;
       case "numeric":
       case "decimal":
         // Per-value scale carried via VariableScaleDecimal (array typmod is -1 in JDBC metadata).
@@ -584,7 +610,7 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       if (raw == null) {
         return null;
       }
-      if (raw instanceof Object[] && isMultiDimensional((Object[]) raw)) {
+      if (raw instanceof Object[] && isMultiDimensional((Object[]) raw, elementType)) {
         if (multiDimensionalWarnedColumns.add(columnId)) {
           log.warn(MULTI_DIMENSIONAL_ARRAY_MESSAGE, columnId);
         } else {
@@ -636,7 +662,11 @@ public class PostgreSqlDatabaseDialect extends GenericDatabaseDialect {
       case JSONB_TYPE_NAME:
         // Raw JSON text; toString() covers both String and PGobject.
         return element.toString();
+      case UUID_TYPE_NAME:
+        // The driver yields java.util.UUID; the scalar path emits the canonical text.
+        return element.toString();
       default:
+        // Primitives and bytea (already a byte[]) pass through.
         return element;
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -837,22 +856,47 @@ private boolean complexTypesEnabled() {
...
+  }
+
+  /**
+   * The failure for an array element the complex types would bind but the feature is off. Raised
+   * from both the DDL and the bind path so an operator sees the same actionable message either way,
+   * rather than an unwritable column or a driver-level type error.
+   */
+  private static TableAlterOrCreateException complexArrayElementDisabled(
+      Schema element, String fieldName) {
+    return new TableAlterOrCreateException(String.format(COMPLEX_ARRAY_ELEMENT_DISABLED,
```
akanimesh7: Same concern here -- 
TableAlterOrCreateException should not be used.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -709,11 +724,15 @@ private Object decodeTemporalElement(
...
+   * <p>A {@code bytea} element is legitimately a {@code byte[]}, so for that type only a nested
+   * {@code Object[]} — {@code byte[][][]}, from a {@code bytea[][]} column — is multi-dimensional.
    */
-  private static boolean isMultiDimensional(Object[] elements) {
+  private static boolean isMultiDimensional(Object[] elements, String elementType) {
     for (Object element : elements) {
       if (element != null) {
-        return element.getClass().isArray();
+        return BYTEA_TYPE_NAME.equals(elementType)
+            ? element instanceof Object[] : element.getClass().isArray();
```
akanimesh7: DIdn't get it. Is special handling for bytea needed ?

KGaneshDatta: `bytea` is the one element type whose value is itself a Java array: pgjdbc returns `byte[][]` for a `bytea[]` column, so a generic `isArray()` check finds an array in every element and nulls the whole column as multi-dimensional.

instanceof Object[] checks if `byte[]` has a primitive component so it isn't an `Object[]`, while a real `bytea[][]` arrives as `byte[][][]` whose elements are. Removing the check fails `shouldReadByteaArrayElements`.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Done updated to ConnectException.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -837,22 +855,47 @@ private boolean complexTypesEnabled() {
...
+  }
+
+  /**
+   * The failure for an array element the complex types would bind while the feature is off.
+   * Shared by the DDL and bind paths for one actionable message. Not a DDL failure: the flag is
+   * off for the whole connector, so no record can succeed and a batch split would find nothing.
+   */
+  private static ConnectException complexArrayElementDisabled(
+      Schema element, String fieldName) {
+    return new ConnectException(String.format(COMPLEX_ARRAY_ELEMENT_DISABLED,
```
akanimesh7, line 1086: Is this backward compatible ? What happens if someone is using older version of the connector and still has bytes without schema name. Does sink connector fail today ?

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
akanimesh7: Thanks — the `byte[]` vs `Object[]` distinction is right. But I don't think the special case is needed at all: `element instanceof Object[]` gives the correct answer for *every* element type, so the ternary and the `elementType` parameter can both go.

- `bytea[]` → elements are `byte[]` → false ✓
- `bytea[][]` → elements are `byte[][]` → true ✓
- `text[]` → elements are `String` → false ✓
- `text[][]` → elements are `String[]` → true ✓

It holds because bytea is the only element type whose 1-D value is an array. `org.postgresql.jdbc.ArrayDecoding` declares one decoder per type and every one produces a reference array (`Long[]`, `Integer[]`, `String[]`, `BigDecimal[]`, `Timestamp[]`, ...) with a single exception, `ArrayDecoder<byte[][]> BYTE_ARRAY_ARRAY`.

Can you try `return element instanceof Object[];` and confirm `shouldReadByteaArrayElements` and `shouldSkipMultiDimensionalByteaArrayElements` both stay green?

_🤖 flagged by Claude_

akanimesh7 (commented):
Approach looks right — deriving the gate from `arrayElementBinding(element) != null` rather than an enumerated list is what stops the DDL and bind paths drifting, and the flag gating is symmetric on both sides. Two things on the description and one on tests.

**The description contradicts the code in two places.**

- "Changes from code review" still says `complexArrayElementDisabled` raises `TableAlterOrCreateException`. It raises `ConnectException`, which is what I asked for and what you replied "Done updated to ConnectException" to. The bullet was never updated.
- The same section says an unbindable named BYTES element "is skipped with a WARN". It isn't — `getSqlType` returns `super.getSqlType(field)`, and `GenericDatabaseDialect.getSqlType` throws *"... doesn't have a mapping to the SQL database column type"*. The task fails; nothing is skipped and no WARN is logged. Behaviour is fine, the description isn't.

**The gate's element matrix is untested.** Test Strategy says "the gate refusing each affected element type on both the DDL and bind paths". The gate refuses seven shapes — `Decimal`, `VariableScaleDecimal`, `Date`, `Time`, `Timestamp`, `MAP<STRING,STRING>` and plain `BYTES`. What is actually covered:

- DDL path: plain `BYTES` only, via `shouldMapByteaArrayToSqlType`, and only `assertThrows(ConnectException.class, ...)` — the message is not asserted.
- Bind path: `Timestamp` only, via an IT.

Nothing covers `Decimal`, `VariableScaleDecimal`, `Date`, `Time` or the MAP element on either path. `isComplexArrayElement` is a pure function of the element schema plus the flag, so no database is needed — a parameterised unit test over the seven shapes x both paths covers the matrix in milliseconds, against two ITs that each deploy a connector. Keep `testTemporalArrayFailsWhenComplexTypesDisabled` for the end-to-end proof that the ArrayStoreException is gone, and push the matrix down to `PostgreSqlDatabaseDialectTest`.

Minor while you are there: `shouldComputeValueTypeCastForArrayColumns` covers `_uuid`, `_jsonb` and `_text` but not `_json`, though `CAST_TYPES` has all three.

The merge-conflict note for #1676 is useful and the `arrayElementBinding(Schema)` vs `(Schema, String)` trap is well spotted — keep it.

_🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
```diff
@@ -837,22 +855,47 @@ private boolean complexTypesEnabled() {
...
+    // STRING-based and degrades losslessly to text[].
+    return !Json.LOGICAL_NAME.equals(element.name()) && arrayElementBinding(element) != null;
+  }
+
+  /**
+   * The failure for an array element the complex types would bind while the feature is off.
+   * Shared by the DDL and bind paths for one actionable message. Not a DDL failure: the flag is
+   * off for the whole connector, so no record can succeed and a batch split would find nothing.
+   */
+  private static ConnectException complexArrayElementDisabled(
```
akanimesh7: `complexArrayElementDisabled` reads like a predicate but returns an exception. The file's own convention is `hstoreOffSearchPathError` — same shape, same `ConnectException` return. `complexArrayElementDisabledError` ?

_🤖 flagged by Claude_

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Currently it also fails too.
`case ARRAY` : has no flag check, so with the flag off an `ARRAY<Decimal>` / `ARRAY<Timestamp>` maps to DECIMAL[] and creates the column, then `bindArray` falls to `primitiveArrayFor`, which handles only INT8…STRING so every insert dies with a bare `ArrayStoreException`. 

This PR moves that to a DDL-time failure naming `sql.complex.types.enable`, so no column is created and the operator gets told what to change. 

Scalar BYTES → BYTEA is untouched and are same.

src/main/java/io/confluent/connect/jdbc/dialect/PostgreSqlDatabaseDialect.java
KGaneshDatta: Ack. I had made the suggested changes.

akanimesh7 (dismissed):

sangeet259 (approved):

PR #1686, opened 2026-08-24
fix: redact sensitive JDBC connector data reaching logs and task-status

sumit-confluent:
## What

Remediates five findings from a sensitive-log audit of the JDBC source and sink connectors (audited submodule commit `f74b4a47`, shipped as private-wrapper v0.212.0). Each change removes customer data from logs and from the connector's task-status failure reason while keeping an actionable diagnostic (exception class, SQLState, errorCode, safe identifiers).

**Report**: https://drive.google.com/file/d/1tOIm4EprPTLNFuiFMu0N5OgrGLMuP62w/view?usp=drive_link

## Findings fixed

| Finding | file:line | Fix | Test |
|---|---|---|---|
| F1 — whole offsets map logged | `JdbcSourceTask.java:278` | Log the single matched `offset`, not the entire `offsets` map (in incrementing/timestamp modes an offset is a customer column value) | `JdbcSourceTaskUpdateTest#foundOffsetLogsOnlyMatchedOffsetNotWholeMap` (log-capture; canary in a second partition) |
| F2 — raw batch SQLException on self-managed default | `JdbcSinkConfig.java` (`trim.sensitive.log` default) | Flip ConfigDef default `false`→`true` so self-managed redacts by default (cloud already pins `true`) | `JdbcSinkConfigTest#shouldEnableSensitiveLogTrimmingByDefault` |
| F3 — column value in field-mapping SQLException | `BulkTableQuerier.java:95`, `TimestampIncrementingTableQuerier.java:246`, `TimestampTableQuerier.java:154` | Redact via `LogUtil.redactSensitiveData` before the WARN log **and** the wrapped `DataException` | `TimestampIncrementingTableQuerierTest#extractRecordRedactsSqlExceptionCarryingColumnValue` (walks the cause chain) |
| F4 — same value on the task-status trace | `TableQuerierProcessor.java:188/194`, `RecordQueue.java:381` | Closed transitively by F3 (these only relay the value handed to them) | covered by F3 (the redacted exception propagates through `poll()`) |
| F5 — password in malformed connection.url | `DatabaseDialects.java:177/178` | Mask the password in the URL before the log and the thrown `ConnectException` (mirrors `sanitizedUrl()`) | `DatabaseDialectsTest#shouldMaskEmbeddedPasswordWhenUrlIsInvalid` |

## Testing

Built and tested on the repo's target JDK (Java 8). Every strengthened test was proven **red on the original code and green after the fix**. Sibling/affected suites pass (`JdbcDbWriterTest`, `JdbcSinkTaskTest`, `TableQuerierTest`, `TimestampTableQuerierTest`, `LogUtilTest`). Checkstyle clean.

## Deliberate behavior changes

- **F2** — self-managed sink deployments now redact raw driver SQLExceptions by default. `trim.sensitive.log` is an internal config, so there is no user-facing docs change; the fully-managed service is unaffected (already pinned `true`).
- **F3/F4** — the task-status failure reason (Connect REST `GET /connectors/{name}/status`) for a column type-conversion error now shows `<redacted> [class=…; SQLState=…; errorCode=…]` instead of the raw driver message. The WARN log **phrase is unchanged**, so existing log searches/dashboards still match.

## Out of scope

- F5 reuses `sanitizedUrl()`'s masking, which covers the `?password=` query-param form (not the `user:password@host` userinfo form) — consistent with the existing helper; broadening it is not part of this change.
- Log-redaction (logredactor) rules are tracked separately via the add-log-redaction workflow.

No sensitive data appears in this PR, its commits, or its tests (all canaries are synthetic).


src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
```diff
@@ -337,7 +337,9 @@ public enum TimestampPrecisionMode {
       + AUTO_EVOLVE + "``).";
 
   public static final String TRIM_SENSITIVE_LOG_ENABLED = "trim.sensitive.log";
-  private static final String TRIM_SENSITIVE_LOG_ENABLED_DEFAULT = "false";
+  // Redact by default so self-managed deployments do not log raw driver SQLExceptions (which can
+  // embed bound row VALUES) at ERROR/WARN. The fully-managed service already pins this to true.
+  private static final String TRIM_SENSITIVE_LOG_ENABLED_DEFAULT = "true";
```
KGaneshDatta: On Cloud, we had this config default set to TRUE - [REF](https://github.com/confluentinc/kafka-connect-jdbc-private/blob/0549f12e64e089ca10161c094272bc99b74cf78c/kafka-connect-jdbc-cloud/cloud-templates/common-jdbc-sink.template#L631-L632).

As, this is common code for both SM and FM, hence we wanted to keep this value as FALSE only in SM deployments.

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
```diff
@@ -92,8 +92,12 @@ public SourceRecord extractRecord() throws SQLException {
         log.warn("Error mapping fields into Connect record", e);
         throw new ConnectException(e);
       } catch (SQLException e) {
-        log.warn("SQL error mapping fields into Connect record", e);
-        throw new DataException(e);
+        // The driver SQLException message can embed the raw failing column VALUE (customer data,
+        // e.g. pgjdbc "Bad value for type ..."). Redact to class/SQLState/errorCode before it
+        // reaches this forwarded WARN log or the DataException that becomes the task-status reason.
+        SQLException redacted = LogUtil.redactSensitiveData(e);
```
KGaneshDatta, line 98: As, this is the SQLException, redacting completely would make it difficult for debugging. 
Instead can we add SQLState and errorCodes from the exception. Eventhough redacting the complete message these state and codes would be helpful for debugging.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
```diff
@@ -275,7 +275,10 @@ mode, getIncrementingColumn(tables.get(0)),
         for (Map<String, String> toCheckPartition : tablePartitionsToCheck) {
           offset = offsets.get(toCheckPartition);
           if (offset != null) {
-            log.info("Found offset {} for partition {}", offsets, toCheckPartition);
+            // Log only the single matched offset for this partition, never the whole `offsets`
+            // map: in incrementing/timestamp modes an offset holds customer column values, so
+            // dumping the full map would emit every table's latest column values on each poll.
+            log.info("Found offset {} for partition {}", offset, toCheckPartition);
```
KGaneshDatta, line 281: What has been changed here ? I see only the comments are added but the log line hasn't changed anything here.

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
```diff
@@ -243,8 +243,12 @@ public SourceRecord extractRecord() throws SQLException {
         log.warn("Error mapping fields into Connect record", e);
         throw new ConnectException(e);
       } catch (SQLException e) {
-        log.warn("SQL error mapping fields into Connect record", e);
-        throw new DataException(e);
+        // The driver SQLException message can embed the raw failing column VALUE (customer data,
+        // e.g. pgjdbc "Bad value for type ..."). Redact to class/SQLState/errorCode before it
+        // reaches this forwarded WARN log or the DataException that becomes the task-status reason.
+        SQLException redacted = LogUtil.redactSensitiveData(e);
+        log.warn("SQL error mapping fields into Connect record", redacted);
```
KGaneshDatta, line 250: Similar suggestion as above.
can we add SQLState and errorCodes from the exception. Eventhough redacting the complete message these state and codes would be helpful for debugging.

src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
```diff
@@ -151,8 +151,12 @@ private PendingRecord doExtractRecord() {
         log.warn("Error mapping fields into Connect record", e);
         throw new ConnectException(e);
       } catch (SQLException e) {
-        log.warn("SQL error mapping fields into Connect record", e);
-        throw new DataException(e);
+        // The driver SQLException message can embed the raw failing column VALUE (customer data,
+        // e.g. pgjdbc "Bad value for type ..."). Redact to class/SQLState/errorCode before it
+        // reaches this forwarded WARN log or the DataException that becomes the task-status reason.
+        SQLException redacted = LogUtil.redactSensitiveData(e);
+        log.warn("SQL error mapping fields into Connect record", redacted);
```
KGaneshDatta, line 158: Similar suggestion as above.
can we add SQLState and errorCodes from the exception. Eventhough redacting the complete message these state and codes would be helpful for debugging.

src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkConfigTest.java
```diff
@@ -69,9 +69,9 @@ public void shouldCreateConfigWithMinimalConfigs() {
   }
 
   @Test
-  public void shouldDisableSensitiveLogTrimmingByDefault() {
+  public void shouldEnableSensitiveLogTrimmingByDefault() {
     createConfig();
-    assertFalse(config.trimSensitiveLogsEnabled);
+    assertTrue(config.trimSensitiveLogsEnabled);
```
KGaneshDatta: Please update this test as we wanted to keep the default to FALSE only in Self-managed.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialects.java
```diff
@@ -174,8 +174,19 @@ static JdbcUrlInfo extractJdbcUrlInfo(final String url) {
       LOG.debug("Validated JDBC URL.");
       return new JdbcUrlDetails(matcher.group(1), matcher.group(2), url);
     }
-    LOG.error("Not a valid JDBC URL: " + url);
-    throw new ConnectException("Not a valid JDBC URL: " + url);
+    // Mask any embedded credentials before the raw URL reaches the log or the thrown message: a
+    // malformed self-managed connection.url can carry a ?password= parameter. No dialect exists
+    // yet at URL-validation time, so mirror GenericDatabaseDialect.sanitizedUrl() here.
```
KGaneshDatta, line 179: Can we make this method as conditional to TRIM_SENSITIVE_LOG_ENABLED config. So, this on CCloud would be sanitized and on SM connectors would not be sanitized as it would be helpful for CP customers to debug if any issue.

src/main/java/io/confluent/connect/jdbc/sink/JdbcSinkConfig.java
sumit-confluent: Done in 0ccda20 — changed the `trim.sensitive.log` (`TRIM_SENSITIVE_LOG_ENABLED`) default from `true` to `false` so self-managed keeps full driver SQLException detail for debugging. The fully-managed cloud template continues to pin it to `true`, so managed connectors still redact bound row values.

src/test/java/io/confluent/connect/jdbc/sink/JdbcSinkConfigTest.java
sumit-confluent: Done in 0ccda20 — renamed to `shouldDisableSensitiveLogTrimmingByDefault` and it now asserts `config.trimSensitiveLogsEnabled` is `false` by default.

src/main/java/io/confluent/connect/jdbc/source/BulkTableQuerier.java
sumit-confluent: These are already included. `LogUtil.redactSensitiveData(e)` renders the message via `buildRedactedMessage` as `<redacted> [class=<exceptionClass>; SQLState=<state>; errorCode=<code>]` and preserves `getSQLState()`/`getErrorCode()` on the returned exception. So this WARN log and the `DataException` reason both carry the SQLState and errorCode for debugging — only the free-text (value-bearing) portion of the driver message is dropped.

src/main/java/io/confluent/connect/jdbc/source/TimestampIncrementingTableQuerier.java
sumit-confluent: These are already included. `LogUtil.redactSensitiveData(e)` renders the message via `buildRedactedMessage` as `<redacted> [class=<exceptionClass>; SQLState=<state>; errorCode=<code>]` and preserves `getSQLState()`/`getErrorCode()` on the returned exception. So this WARN log and the `DataException` reason both carry the SQLState and errorCode for debugging — only the free-text (value-bearing) portion of the driver message is dropped.

src/main/java/io/confluent/connect/jdbc/source/TimestampTableQuerier.java
sumit-confluent: These are already included. `LogUtil.redactSensitiveData(e)` renders the message via `buildRedactedMessage` as `<redacted> [class=<exceptionClass>; SQLState=<state>; errorCode=<code>]` and preserves `getSQLState()`/`getErrorCode()` on the returned exception. So this WARN log and the `DataException` reason both carry the SQLState and errorCode for debugging — only the free-text (value-bearing) portion of the driver message is dropped.

src/main/java/io/confluent/connect/jdbc/source/JdbcSourceTask.java
sumit-confluent: The log line did change: the argument went from `offsets` (the whole offsets map) to `offset` (the single matched offset for this partition). In incrementing/timestamp modes an offset holds customer column values, so logging the full map dumped every table's latest values on each poll; now only the one matched offset is logged. The one-word diff sits right under the new comment, which makes it easy to miss.

src/main/java/io/confluent/connect/jdbc/dialect/DatabaseDialects.java
sumit-confluent: This masking only replaces `password=****` in the JDBC URL — it leaves host/db/other params intact, so it doesn't impair debugging, and a plaintext DB password in logs is a credential leak even on self-managed. So I'd suggest keeping it always-on, unlike the SQLException *message* redaction (which has debug value and is gated behind `trim.sensitive.log`). If you'd still prefer it gated, that would need a source-side `trim.sensitive.log` config threaded into the static `extractJdbcUrlInfo` — happy to do that if you want it.

KGaneshDatta (approved):

